Issue #45, opened 2014-11-06
Add ability to use any date format


meg2208:
You should be able to enter a custom date format as a prop of the DatePicker element. It currently forces YYYY-MM-DD format.

PR #46, opened 2014-11-06
Adds ability to use a custom date format

meg2208:
A date format string can be entered as the prop `dateFormat` to the `DatePicker` element. If you do not enter this prop, it will default to `YYYY-MM-DD`, which was previously the only valid format allowed. The requirements of one of my current projects was that I needed a different format, so I put in the issue https://github.com/Hacker0x01/react-datepicker/issues/45. Please let me know if you have any questions. Thanks!

### DIFF
--- a/example.html
+++ b/example.html
@@ -40,11 +40,13 @@
             DatePicker({
               key: 'example1',
               selected: this.state.start_date,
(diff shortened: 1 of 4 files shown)


martijnrusschen:
Closing this since #48 is already merged.

PR #65, opened 2015-01-27
Fix issues with clicking on the calendar, primarily for IE11

hahahana:
I noticed that I wasn't able to select a date or cycle through the months on IE11. If I tried to increment the month, the calendar would close, same with trying to select a date. I even checked that it wasn't just my codebase by throwing up the example file on BrowserStack.

I believe the bug comes from the blur event firing twice in IE11 (don't ask me for any literature to back this up though!). I put a breakpoint in handleBlur and it was indeed firing one more time than in Chrome, which obviously would then set focus to be false since _shouldBeFocussed was never reset to true since handleCalendarMouseDown was not being fired again, just handleBlur.

The primary purpose of both handleBlur and handleCalendarMouseDown was to detect if the blur came from clicking inside or outside the calendar. The extra blur meant I couldn't use these as a function to determine what was a legitimate intent to close. I had the best luck utilizing [react-onclickoutside](https://github.com/Pomax/react-onclickoutside), which originated from [this discussion](https://github.com/facebook/react/issues/579). It has so far proved very accurate in detecting clicks outside a component and I was able to completely remove the handleBlur and handleCalendarMouseDown.

I'm not really suggesting that this be merged in to the main library, but to offer a patch to those experiencing similar issues in IE11 and to start a larger discussion on this issue that I faced. 

----
FYI, I pushed up the example to my github project page to make it easier to view and test the bug http://hanawang.com/react-datepicker/

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -24,7 +24,8 @@
   "dependencies": {
     "react": "0.11.1",
     "moment": "2.8.2",
-    "tether": "0.6.5"
+    "tether": "0.6.5",
+    "react-onclickoutside": "0.2.1"
   },
   "ignore": [
     "**/.*",
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
     "react": "^0.11",
     "moment": "^2.8",
     "tether": "^0.6",
-    "reactify": "^0.14"
+    "reactify": "^0.14",
+    "react-onclickoutside": "0.2.1"
   },
   "scripts": {
     "test": "grunt travis --verbose"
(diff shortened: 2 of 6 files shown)


martijnrusschen:
Thanks for creating this Pull Request. I tested this in all major browsers (including IE 11 and lower) and it works perfect now.

martijnrusschen:
This fix is now released in https://github.com/Hacker0x01/react-datepicker/releases/tag/v0.4.0

PR #71, opened 2015-01-30
Make usable without globals

jbrumwell:
This PR makes it possible to require the plugin without have globals defined

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,14 +29,17 @@
     "grunt-jsxhint": "0.3.0",
     "grunt-scss-lint": "^0.3.4",
     "jest-cli": "0.1.18",
+    "react": "^0.11",
     "react-tools": "0.11.1"
   },
   "dependencies": {
-    "react": "^0.11",
     "moment": "^2.8",
     "tether": "^0.6",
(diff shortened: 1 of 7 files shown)


RSO:
Could you elaborate as to why you closed the pull request? I'd be interested in landing something that fixes this, if I can find a proper way to make it work in all cases.

jbrumwell:
I still had some work I needed to do on. Can you elaborate on the use cases you need to support?

RSO:
Well, at HackerOne we're still not up and running with modules, so for that we still need a file that compiles and sets the `DatePicker` variable on the `window`.

I would be perfectly fine with this file being generated as a separate `react-date-picker-global.js` or even as part of a separate repo, it's just that I couldn't find a clean/simple way to do this without `React` and `Tether` being compiled in that file as well.

So, I would like to have a file, that holds all datepicker-specific stuff, without having all the external dependencies compiled inside of it. Does this help?

jbrumwell:
Yeah should be fairly easy to achieve, I'll take a look this weekend if I get time

Issue #53, opened 2015-01-13
How can I initialize it with an empty value?


illing2005:
Hi, 
I like your component, but I am not able to initialize it with an empty value.
How can that be done?
Thanks!

RSO:
Hey @illing2005 !

Thanks, how would you like this to work? At this point it is not possible indeed, mostly because I'm not quite sure how the component should behave.

What do you expect to see when you open the date-picker when it's empty? I'm thinking the current month?

PR #72, opened 2015-01-31
Support for null date states

hahahana:
This PR adds the ability to initialize a datepicker with a null value. This also introduces placeholder text.

You can test this [live](http://cdn.rawgit.com/hahahana/react-datepicker/21464e529b72d17a2075ea55c26eb449c0d3d2fe/example.html).

----
This should close https://github.com/Hacker0x01/react-datepicker/issues/53

### DIFF
--- a/react-datepicker.js
+++ b/react-datepicker.js
@@ -85,7 +85,7 @@
 var Day = require('./day');
 var DateUtil = require('./util/date');
 
-var Calendar = React.createClass({displayName: "Calendar",
+var Calendar = React.createClass({displayName: 'Calendar',
   mixins: [require('react-onclickoutside')],
 
   handleClickOutside: function() {
@@ -94,7 +94,7 @@ var Calendar = React.createClass({displayName: "Calendar",
 
   getInitialState: function() {
     return {
-      date: new DateUtil(this.props.selected).clone()
+      date: !! this.props.selected ? new DateUtil(this.props.selected).clone() : new DateUtil(moment())
     };
   },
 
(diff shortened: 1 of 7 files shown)


martijnrusschen:
Would it be worth to add some tests that support this feature?

react-datepicker.js
```diff
@@ -156,29 +156,29 @@ var Calendar = React.createClass({displayName: "Calendar",
 
   render: function() {
     return (
-      React.createElement("div", {className: "datepicker"},
```
martijnrusschen: Are these changes related to your PR?

hahahana: It's probably just my build setup? I haven't really been checking the output file. Can you rerun it on your machine?

RSO: `master` has moved from React `0.11` to `0.12`, if you run `npm install` and try to rebuild this should all be back to normal!

RSO:
Sorry for the late reply, I was hoping to reply sooner, but my flight to SF took a little more preparation than expected.

I'm kind of worried about the three inline if's that power this pull request. The logic feels a little implicit. I couldn't find a better spot/way to do this, so I guess we should just do it like this!

Did you try any other solutions to achieve this goal?

hahahana:
Take a look at my second commit, I pulled in a method that I was saving in another PR, where it proved to really be necessary. I added a safeDateFormat method too, does that make it feel a bit less implicit?

What do you think? I personally can't think of anything else to do at this point.

RSO:
Yeah, this is way better!

Issue #69, opened 2015-01-30
Add ability to disable days in the future or past 


lbekkema:
The datepicker ables users to select dates in the future or past, but for some cases these dates are not applicable. For example if you'd have a dashboard, you want to force users to only select dates in the past. Or if you'd have an appointment scheduler you want to force users to select dates in the future.

hahahana:
I took a stab at this [in my repo](https://github.com/hahahana/react-datepicker/pull/3/files). It depends on my other PR [(#72)](https://github.com/Hacker0x01/react-datepicker/pull/72), so I haven't submitted it here yet.

azizijunaid:
for past date disbable

const today = new Date();
<DayPicker disabledDays={{ before: today }} />,

sabinbogati1:
Did anyone solved it??

yashmiRajamanthrie:
import moment from 'moment';
 
minDate={moment().toDate()}

try this ,

annadayadev11:
> import moment from 'moment';
> 
> minDate={moment().toDate()}
> 
> try this

Hi, it's working, how can I also disable future date? Thank you.

satish769843:
> > import moment from 'moment';
> > minDate={moment().toDate()}
> > try this
> 
> Hi, it's working, how can I also disable future date? Thank you.

use maxDate

PR #76, opened 2015-02-05
Add ability to disable dates in the future and/or past

hahahana:
Live example [here](http://cdn.rawgit.com/hahahana/react-datepicker/23f3362c1e3b7be530d2891fe4238be13a1c5469/example.html).

The example is getting a bit gnarly, what do you guys think? Maybe separate pages for different functionality with some documentation on the page too?

---
Should close https://github.com/Hacker0x01/react-datepicker/issues/69

### DIFF
--- a/src/day.js
+++ b/src/day.js
@@ -1,16 +1,23 @@
 /** @jsx React.DOM */
 
 var Day = React.createClass({
+  handleClick: function(event) {
+    if (this.props.disabled) return;
+
+    this.props.onClick(event);
+  },
+
   render: function() {
     classes = React.addons.classSet({
       'datepicker__day': true,
+      'datepicker__day--disabled': this.props.disabled,
       'datepicker__day--selected': this.props.day.sameDay(this.props.selected),
       'datepicker__day--this-month': this.props.day.sameMonth(this.props.date),
       'datepicker__day--today': this.props.day.sameDay(moment())
     });
 
     return (
-      <div className={classes} onClick={this.props.onClick}>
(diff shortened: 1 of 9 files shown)


src/day.js
```diff
@@ -4,13 +4,14 @@ var Day = React.createClass({
...
       'datepicker__day': true,
+      'datepicker__day--disabled': this.props.disabled,
       'datepicker__day--selected': this.props.day.sameDay(this.props.selected),
       'datepicker__day--this-month': this.props.day.sameMonth(this.props.date),
       'datepicker__day--today': this.props.day.sameDay(moment())
     });
 
     return (
-      <div className={classes} onClick={this.props.onClick}>
+      <div className={classes} onClick={this.props.disabled ? null : this.props.onClick}>
```
RSO: I think that I'd rather have this logic inside of it's own method:

```
handleClick: function(event) {
  if (this.props.disabled) return;

  this.props.onClick(event)
}
```

RSO:
Yeah, I agree about the example. I was hoping to trick @lbekkema or @jenskanis in designing a Github page for this component, with documentation etc.

Great job on the PR, I'll let @martijnrusschen decide if it's alright with him, but it looks good to go! Awesomesaus!

martijnrusschen:
Thanks, looks great! I agree about the example, at first that was a quick hack to show the datepicker, but with these different configurations it becomes really ugly. I'll try to make something better this weekend.

jenskanis:
@RSO we can generate a Github page with minimal effort, but we need to have sufficient permissions of the repo. Can you give us that?

PR #78, opened 2015-02-10
Remove disabled-look from days in previous/next month

lbekkema:
While working with the react-datepicker I discovered that days in the
previous and/or next months look like they're disabled, but they're
not.

We should make it more clear dates in the previous and/or next month
are selectable. In order to achieve this, I propose to remove this
`this-month` state and make the day's color default to black.

Screenshot:

screen shot 2015-02-10 at 17 27 32

### DIFF
--- a/react-datepicker.css
+++ b/react-datepicker.css
@@ -95,7 +95,7 @@
 }
 
 .datepicker__day {
-  color: #ccc;
+  color: #000;
   display: inline-block;
   width: 24px;
   line-height: 24px;
@@ -106,9 +106,6 @@
 .datepicker__day:hover {
   border-radius: 4px;
   background-color: #f0f0f0;
-}
-.datepicker__day--this-month {
-  color: #000;
 }
 .datepicker__day--today {
   font-weight: bold;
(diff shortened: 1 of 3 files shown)


martijnrusschen:
I agree, looks better!

hahahana:
I find this really confusing. I pretty much expect not current month days to look 'disabled'. It is clear that you can click on them because when you hover over a disabled date it shows that it is selectable.

For example, March is a huge month, and now it just looks like a blanket of dates:
screen shot 2015-03-16 at 4 30 24 pm

I personally find the old version much easier to read:
screen shot 2015-03-16 at 4 38 52 pm

----
See the standard in datepickers: http://www.eyecon.ro/bootstrap-datepicker/
bootstrap-datepicker

Example in the wild: 
http://www.kayak.com/
Kayak datepicker

martijnrusschen:
Agreed, the old situaon gave a better overview what's in a month and what's not. The only downside of the previous implementation was the it's hard to make a difference between disabled dates and dates that aren't in the current month. I'm not sure how we should solve that since there's 1 really big difference between those: The ability to click on a date.

hahahana:
I'm not sure what you mean, you were still able to click on the date if it wasn't in the featured month. I think that's pretty standard and expected behavior.

RSO:
@hahahana @lbekkema implemented a way of disabling dates, so you could say that the user is only allowed to select dates within a certain range. He decided to re-use the "disabled" state that we had for dates that were not in the current month, which really made things harder to understand. IMO all we have to do is split up the two different disabled states and give them their own styling:
- Disabled (can't select)
- Not in the current month (can select, but will switch months)

PR #80, opened 2015-02-12
Front page design improvements

jenskanis:
**Old**:
screen shot 2015-02-12 at 15 25 32

**New**:
screen shot 2015-02-12 at 15 25 19


martijnrusschen:
Looks good, now we should add some content :)

Issue #47, opened 2014-11-07
Unable to run with npm/browserify


meg2208:
Out of the box this is not working if I just import the package using requireJS/browserify. The external dependencies (React, moment, Tether) are assumed to be globals in the source. @RSO are you aware of this issue? I notice there is another issue https://github.com/Hacker0x01/react-datepicker/issues/42 but not sure how thoroughly tested that was.

mattdell:
I'm aslo having this issue. Would be great to be able to use it this way.

MandarinConLaBarba:
I've submitted a PR for this: https://github.com/Hacker0x01/react-datepicker/pull/81

RSO:
Now that #81 has been landed, I think that this ticket can be closed. Can you confirm?

PR #81, opened 2015-02-17
Adding in requires for cjs support

MandarinConLaBarba:


MandarinConLaBarba:
ref #47

RSO:
Hey, thanks for taking the time to submit this pull request! 

Unfortunately we're not fully running browserify yet at HackerOne, which is a shame. What this means is that we actually include the `react-datepicker.js` file in our own application JavaScript. In order for this flow to keep working for us, we need to have a compiled DatePicker file, which doesn't pack its external dependencies (`moment` and `React`).

I tried figuring out how to get this to work, but haven't found a way yet. Do you maybe know how we can achieve just this? Any solution would be fine, really, as long as we can support both browserify/global.

Ref #71

RSO:
I will spend some time on this tomorrow if you don't have time, I didn't try the `peerDependencies` stuff, so that might be the missing link!

MandarinConLaBarba:
Hmm, perhaps you can use `browserify-shim` in the projects that want `moment` and `react` as global deps? https://github.com/thlorenz/browserify-shim/wiki/browserify-shim-recipes#three-as-global

RSO:
Ah yeah, now it's starting to make sense! Awesome!

What should we do with Tether? I see that you still require it externally in the example, did you run into any trouble?

MandarinConLaBarba:
Sweet :)

Re `tether`, I think an oversight on my end. I think it should be removed from globals if it's a dep for this plugin, and instead use `require('tether')` where applicable. I can do that later if you want

lappkast:
We're getting the error:

```
ReferenceError: React is not defined
http://localhost:8080/js/app.bundle.js
Line 36348
```

With webpack and npm, and no bower and no browserify. It also seems to stem from the same problem, that React is an assumed global.

Does this PR solve that problem?

Also, it seems this PR adds react.datepicker.js as a file on its own, is that intended?

MandarinConLaBarba:
You mean `react-datepicker.js`? I thought it was already in the source tree. I will have to test this out w/ webpack, I've never used that before.

lappkast:
screen shot 2015-02-20 at 17 14 36

yes, exactly =)

I'm having trouble resolving tether now too.

RSO:
@lappkast that addition is because React/MomentJS are now packed with that file. I don't think this makes sense, since I assume that only non-browserify users will use this file. I'm testing this pull request locally to see what the best option is for packing files.

RSO:
Awesome, this should be good to go. There are some changes that I have to apply to get this working for HackerOne (or any other party relying on globals), but I'll just add that in a follow up commit. @lappkast I'll also get that the `react-datepicker.js` file back to its original size.

PR #82, opened 2015-02-17
Moved react to peerDependencies, and reactify to devDependencies

MandarinConLaBarba:
React I believe should be a `peerDependencies` item since I believe this qualifies as a plugin component. This will help reduce this chance of duplicates in the bundle. Also, I believe `reactify` is only for dev.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,13 +29,16 @@
     "grunt-jsxhint": "0.4.0",
     "grunt-scss-lint": "^0.3.4",
     "jest-cli": "0.1.18",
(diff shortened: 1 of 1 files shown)


RSO:
Thanks for the help! I'm kinda new to all the `browserify`/`bower`/`npm` stuff, so I really appreciate that you're taking the time to help out! Things are starting to make sense now ;-)

MandarinConLaBarba:
Sweet, very glad to help out.

PR #84, opened 2015-02-20
add grunt-cli to devDependencies

natlownes:
since `./node_modules/.bin` is added to your PATH when running npm
scripts, add grunt-cli to deps to cut out `npm install -g grunt` step

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,13 +24,12 @@ Installing is really simple and can be done in two different ways:
 ## Local Development
 
 - Install Bower `npm install -g bower`
-- Install Grunt CLI `npm install -g grunt-cli`
 - Install Bower Packages `bower install`
 - Install packages `npm install`
 - Start a static webserver `python -m SimpleHTTPServer`
 - And visit `localhost:8000/example.html` to see the example.
 
-To run tests, simply run `grunt jest`.
+To run tests, simply run `npm test`.
 
 ## License
 
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   "devDependencies": {
     "grunt": "0.4.5",
     "grunt-browserify": "3.2.1",
+    "grunt-cli": "0.1.13",
     "grunt-contrib-sass": "0.8.1",
     "grunt-contrib-watch": "0.6.1",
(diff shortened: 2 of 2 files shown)


natlownes:
I usually do a `npm install && npm test` when I'm first checking out a project.  Definitely a take-it-or-leave-it pull request - just something I noticed.  I won't cry a billion tears if you don't care.  Thanks!

RSO:
Thanks! This is actually my first plugin with this setup, so all help/input is appreciated!

RSO:
And yeah, now that I know about `npm test` I have to say that I like it too ;-)

RSO:
BTW, @natlownes how do you feel about grunt in this repo? I keep running into troubles upgrading packages because of their grunt counterparts not being maintained properly, and I keep reading about people dropping grunt/gulp for vanilla npm. I think I'm going to try out and drop grunt in favour of plain npm tasks.

natlownes:
Awesome, my pleasure!  And yeah, the fact that those npm scripts can be written as if `./node_modules/.bin` is in already in your path is very helpful; no more having to have to follow a list of instructions to get a project's tests running or install anything globally.

re:  grunt, gulp:  I used grunt once and thought that writing a Makefile was way easier.  I think gulp is easier to deal with; it's all mostly just something -> some number of transforms -> some destination, which is easy to understand and pretty easy to debug.  If you want, you could drop everything, define each build step as npm scripts and have an `npm run build` or whatever that runs them all in order.  

re:  further input on your project:  something I wanted to do in a project of mine was to declare this project as an npm dependency, require it in my nonsense code and use it as part of my usual build process.  I added it as a dependency to `package.json` just fine and blindly tried requiring a couple different files; the `react-datepicker.js` in the root and requiring `react-datepicker/src/datepicker` directly and both failed because of JSX.  I think the good thing to do would be to have a build step that does the JSX -> Javascript business and then for you to publish that to npm.  I was going to do that anyway, so I'll open a pull request for that if you'd like.

And hey:  thank you so much for writing tests in this project!  I think this is the only React date-thing I could find that both has some tests and doesn't require jQuery.

PR #86, opened 2015-03-01
Remove unneeded old JSX pragma

justinwoo:
The peer dependencies in package.json show "react": "^0.12" as a peer dependency, which means that the JSX pragma is no longer needed for valid consumer modules.

This also fixes some deprecation error messages with tools like babel-loader.


RSO:
Sorry for the late merge! Thanks!

PR #87, opened 2015-03-03
prevent variable leakage

nishp1:


RSO:
Sorry for the late merge, thanks for fixing this!

PR #91, opened 2015-03-16
Remove tether constraints on popover [fixes FF bug]

hahahana:
In Firefox, I was seeing behavior like this:
screen shot 2015-03-16 at 4 11 50 pm

I also noticed that on pages that scrolled, the datepicker would fix to the top of the page, due to the pin constraint. I don't think that this component needs to have scroll constraints - we really just want it to appear next to the input.

Also, has the build process changed? If so, can someone update the docs? I have no idea how to update the stuff in /example.

### DIFF
--- a/src/popover.jsx
+++ b/src/popover.jsx
@@ -38,14 +38,7 @@ var Popover = React.createClass({
       targetOffset: '10px 0',
       optimizations: {
         moveElement: false // always moves to <body> anyway!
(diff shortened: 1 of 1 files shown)


RSO:
Hey @hahahana! Thanks for the PR. I'll go through the README and update accordingly. Tested your PR in both Firefox/Chrome!

PR #100, opened 2015-04-13
First attempt at ability to clear date inputs.

ryan1234:
There might be a way to do this already without this pull request, but I couldn't figure it out. =(

I have a form with date fields and each date field is nullable. This means the value can either be a valid date (datepicker handles this great), or it can be empty, which means `null` right now.

This PR is one approach to supporting the ability to "clear" a date value if the input text is empty. 

I realize `null` might not be the best thing to emit to onChange. Open to any ideas like an empty string etc.

Hoping this PR can be a first step towards a good solution.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -27,6 +27,8 @@ var Calendar = React.createClass({
   },
 
   componentWillReceiveProps: function(nextProps) {
+    if (nextProps.selected === null) { return; }
+
     // When the selected date changed
     if (nextProps.selected !== this.props.selected) {
       this.setState({
--- a/src/date_input.jsx
+++ b/src/date_input.jsx
@@ -45,6 +45,8 @@ var DateInput = React.createClass({
 
     if (this.isValueAValidDate()) {
       this.props.setSelected(new DateUtil(date));
+    } else if (event.target.value === '') {
+      this.props.clearSelected();
     }
   },
 
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -37,6 +37,10 @@ var DatePicker = React.createClass({
     this.props.onChange(date.moment());
   },
 
+  clearSelected: function() {
+    this.props.onChange(null);
+  },  
+
   onInputClick: function() {
     this.setState({
       focus: true
@@ -71,6 +75,7 @@ var DatePicker = React.createClass({
           handleClick={this.onInputClick}
           handleEnter={this.hideCalendar}
           setSelected={this.setSelected}
+          clearSelected={this.clearSelected}
           hideCalendar={this.hideCalendar}
           placeholderText={this.props.placeholderText} />
         {this.calendar()}


RSO:
Interesting approach. I get your hesitation concerning `onChange(null)` it doesn't seem optimal. I will see if I can dig into this PR a little further early this week. Thanks for taking the time to submit this suggestion, I really appreciate it!

ryan1234:
Yeah I was thinking we could push out the function to clear to the user as an option prop, but I didn't want to clutter the "interface" to the component. Thanks for the review. =)

ryan1234:
Just checking in to see if you've had a chance to think more about this PR.

RSO:
Sorry for the wait. I didn't find the time to check this out any further, so lets just go with this solution!

QingkaiLi:
Hi RSO, when will you do release? i'm waitting for the latest code

martijnrusschen:
@QingkaiLi I'll make sure this will be released today!

QingkaiLi:
Cool, then i needn't override it in my local. just waiting your new release. Thanks

martijnrusschen:
Here it is: https://github.com/Hacker0x01/react-datepicker/releases/tag/v0.7.0

QingkaiLi:
I Checked out the new version but there is a issue: Uncaught TypeError: n.findDOMNode is not a function

ryan1234:
What version of react are you using?

QingkaiLi:
React 0.12

QingkaiLi:
can you try : this.getDOMNode()

QingkaiLi:
The issue occurs from  toggleFocus method in date_input.jsx. maybe  this.getDOMNode() works. can you fix?

martijnrusschen:
Can you try upgrading to React 0.13? getDOMNode is being deprecated in favor of findDOMNode.

martijnrusschen:
From the React 0.13 release notes:

> Added new top-level API React.findDOMNode(component), which should be used in place of component.getDOMNode(). The base class for ES6-based components will not have getDOMNode. This change will enable some more patterns moving forward.

https://facebook.github.io/react/blog/2015/03/10/react-v0.13.html

QingkaiLi:
i'm using React 0.12 and cause of too may depencies i cannot upgrade it. since react-datepicker is using React 0.12 can you just stay using the version and find other way to solve this problem?

QingkaiLi:
Or can you release  react-datepick 0.6.1 using this.refs.input/this.getDOMNode() and React 0.12 and realse react-datepicker 0.7 using React0.13?

QingkaiLi:
Because React 0.12 and 0.13 has huge differences. so can you do two release so that all of us can use your component no matter which React version we're using. Thanks

RSO:
Hi @QingkaiLi I'm sorry that we weren't clearer on dropping support for React 0.12. If you provide a PR for an older version I might consider merging it, but I will not spend time myself on supporting older versions of React. 

Because of React's alfa-stage every new version will have breaking changes, which makes it harder to support multiple versions.

SinisterGlitch:
Everything still works if you just uncomment this code.
	
  toggleFocus: function toggleFocus(focus) {
	    if (focus) {
	      //    React.findDOMNode(this.refs.input).focus();
	    } else {
	      //  React.findDOMNode(this.refs.input).blur();
	    }
  },

PR #109, opened 2015-05-18
className property added for input field.

cy6eria:


martijnrusschen:
Sorry for the delay @cy6eria. I'm really busy this week, so I'll try to get back to you this weekend or early next week! One thing I noticed is that your PR changes 20.000 lines of code, this makes things really hard to review. Is there any way to get your change smaller?

iam4x:
@martijnrusschen This is correct, I've checked it (I need it for my project too)

Look at the end of files changed, he just adds `this.props.className` if it has one.

RSO:
Hey!

Thanks for taking the time to write this! I will look in to this PR first thing tomorrow, as I have the feeling that there might be a better way to do this. I'm also going to look in to a solution to the unreadable PR's because of the big changes in the compiled files.

src/datepicker.jsx
```diff
@@ -77,6 +77,7 @@ var DatePicker = React.createClass({
           setSelected={this.setSelected}
           clearSelected={this.clearSelected}
           hideCalendar={this.hideCalendar}
+          className={this.props.className}
```
RSO: I think that `className` is not really declarative. Maybe for now we should go with `inputClassName`, unless you have another suggestion?

src/date_input.jsx
```diff
@@ -76,7 +76,7 @@ var DateInput = React.createClass({
       onKeyDown={this.handleKeyDown}
       onFocus={this.props.onFocus}
       onChange={this.handleChange}
-      className="datepicker__input"
+      className={this.props.className ? this.props.className + " datepicker__input" : "datepicker__input"}
```
RSO: How about just defaulting the `className` to `datepicker__input` through `getDefaultProps`? That way we can just do `{this.props.className]` here.

RSO: Otherwise we might want to call this `additionalClassName`.

RSO:
I'm wondering, are error classes the only use case for adding this className? I'm kinda weary of adding classes like this, as is it doesn't really seem to scale that much (e.g. I'm afraid that someone wants a different className for selected days in the future).

We could also think of adding a property that indicates that the datepicker is in an error state.

All in all, if we cant figure out a better way I'd be OK landing this once the comments are addressed, as I don't want to block you for such a small request.

martijnrusschen:
This has been fixed in b61725c82750ac4963c38c12cdc8cbf1c1ded33e, not sure if this was fixed before, but it looks like that this PR isn't needed anymore.

PR #111, opened 2015-05-20
Ran through Grunt to get updated example js

tylerlee:
I was having no luck getting the [demo](http://cdn.rawgit.com/Hacker0x01/react-datepicker/master/example/index.html)  to work --- or on the example when loaded up locally. 

(clicking into a cell and picking a new date didn't actually change the values shown in the inputs)

I believe that the JS just needed to get run through grunt again so I did that here and it seems to be working for me now. Not my normal area of expertise though.


RSO:
Thanks, I had some other changes myself, so will run `grunt build` again on master. I'm wondering if there isn't a better way of doing this. It feels like we should either:
- have more control over versions that npm installs
- not include the compiled files in source control

tylerlee:
:+1: --- I don't know a better way so I'm no help there --- but we're gonna start using react-datepicker so thanks!

PR #112, opened 2015-05-20
fix: remove `NameFormat` rule breaking `grunt build`

iam4x:
I couldn't build the application without removing this rule in the `.scss-lint.yml`

screen shot 2015-05-20 at 11 56 58


kevzettler:
+1

Issue #101, opened 2015-04-14
Support for React 0.13


ryan1234:
I'm using React 0.13 and this causes react-datepicker to show a warning about getDOMNode being deprecated. (https://facebook.github.io/react/blog/2015/03/10/react-v0.13.html#new-features).

For lines like this in the code:

https://github.com/Hacker0x01/react-datepicker/blob/1d8177aff26042c10bfd25c46d4b4c5b0c98344c/src/date_input.jsx#L33

... what would your thoughts be about abstracting out getDOMNode and putting it behind a utility function that could be swap between getDOMNode and findDOMNode? The utility function would always return a node, so the client code would never know how it got the node, only that it got one.

In the build process you could specify the React version you want and the utility function would be wired to only include the support method for the appropriate version of React. 

Something like (pseudo code):

```{js}
if (__REACT_VERSION_0.12__) {
  return getDOMNode(component);
} 

if (__REACT_VERSION_0.13__) {
  return findDOMNode(component);
}
```
That's the current approach I use with webpack and environment variables to build different versions of components.

Or maybe just live with the warnings in dev. =)

RSO:
Hey! Thanks for taking the time to write up this issue!

I like your approach, although I'm not sure if it wouldn't be a little too much for a small library like this one. 

I figure we could also just stop supporting 0.12 in the next version (I'm not sure if there are statistics as to how many people still use 0.12). Being that both react and this datepicker are still `0.` I figure we can still make breaking changes in small version bumps.

ryan1234:
Sounds good. I'll just keep my eye out for the switch when it happens. Bummer they've introduced some breaking changes, but I suppose to be expected. =)

PR #116, opened 2015-05-27
replaced findDOMNode with getDOMNode

ibjohansen:
in date_input.jsx in order to ensure React 0.12 compatibility.


RSO:
Hey there! Thanks for taking the time to create this PR. Unfortunately this code is needed to prevent deprecation warnings on React 0.13. I would be willing to support 0.12 as well, but only if we take an approach as suggested in #101. Do you think that you can adjust your PR to make the check conditional?

RSO:
Furthermore: I'm not sure how far I can go guaranteeing that we will keep supporting older versions of React. I'll promise to always consider PR's like this that don't seem to costly.

ibjohansen:
Hi!
OK, thanks, I agree that supporting older versions when we're still on 0.x is not worth the effort. Perhaps I could suggest that you update the package.jso so that it reflects this.
"peerDependencies": {"react": ">=0.12 <0.14"}

Unfortunately I'm stuck with 0.12 for my project, so I'll just use my fork there.

I also noted that popover.jsx does uses getDOMNode on line 35:
  _tetherOptions: function() {
    return {
      element: this._popoverElement,
      target: this.getDOMNode().parentElement,
      attachment: 'top left',
      targetAttachment: 'bottom left',
      targetOffset: '10px 0',
      optimizations: {
        moveElement: false // always moves to <body> anyway!
      }
    };
  },

PR #129, opened 2015-06-26
Feature/on blur

at0g:
Adds support for onBlur callback:

- Adds a new state variable (virtualFocus) to track when the focus is moved to the calendar control rather than directly to another element.
- When blurring focus from the date picker to another element (such as via tab), the calendar is closed after a short delay.
- The onBlur callback is passed the current value of the datepicker as its first argument.


martijnrusschen:
Thanks!

PR #130, opened 2015-07-01
Add window constraint, move arrow when needed

### LINKED ISSUE #29: Show the arrow on the correct side of the datepicker
Right now, we always show the arrow on the top-left corner of the date-picker. However, when there isn't enough space below the input, the date-picker is shown above the input. When this happens, the arrow should be shown on the correct side.

Tether has helper-classes for this.

ntdb:
I found that I needed to add the window constraint to the popover to keep the datepicker on-screen when there's not enough space below the input. I also implemented styles to move the arrow to the bottom of the datepicker when the popover moves above the input (should resolve #29).

Note that Tether does not currently reverse offsets when the window constraint triggers (see https://github.com/HubSpot/tether/issues/85): this means the arrow spacing will be inconsistent between the two states.


martijnrusschen:
Nice work! We could probably work a little bit on the positioning of the element when it's positioned to the top, but that's future work. Can you fix the merge conflicts so I can merge your PR?

ntdb:
@martijnrusschen Thanks! Agreed, this is essentially an minimum viable implementation for my use case. There's lots of room to leverage Tether to a greater extent.

martijnrusschen:
Thanks again, I'll look into it soon!

PR #135, opened 2015-07-09
Allow parent to pass updates to selected

### LINKED ISSUE #133: Unable to set selected state from external component
I think there is an issue with the selected state value.  Specifically, a parent component can set the selected value by passing it as a prop, but if the parent were to change its value, the DatePicker component does not receive the update.  I have a separate "Clear" button on my page that should be able to clear the input.  I believe the root of the issue is the use of getInitialState() to set selected value from the prop:

```
getInitialState: function() {
    return {
      focus: false,
      selected: this.props.selected
    };
  },
```
The documentation states this as an antipattern: https://facebook.github.io/react/tips/props-in-getInitialState-as-anti-pattern.html

I believe the solution is to have a componentWillReceiveProps that can keep the selected state updated on changes, something like:

```
componentWillReceiveProps(newProps){
  this.setState({
    selected: newProps.selected
  });
}
```

What do you think? I'm happy to submit a pull request.

sconno05:
This change will allow a parent component to control the selected value of the date picker.  Use case is:

1. Initialize component with null value
2. User selects a date
3. A separate "Clear" button clears out the field by setting the selected prop to null

Fixes #133

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -31,6 +31,12 @@ var DatePicker = React.createClass({
     };
   },
 
+  componentWillReceiveProps: function (nextProps) {
+      this.setState({
+        selected: nextProps.selected
+      });
+  },
+
   getValue: function() {
     return this.state.selected;
   },


jbrumwell:
:+1: just had this need as well

PR #140, opened 2015-07-24
Change 'lodash' from a devDependency to a regular dependency

frank-weindel:
Lodash is required in the source of 'calendar.jsx'. It should be a regular dependency instead of a 'devDependency'. The current dep configuration won't work for build environments that don't already have lodash installed.


zackify:
i was just going to make this a pull request. Please merge this guys....

PR #142, opened 2015-07-28
Replace Jest with Karma

RSO:
Decided it was time to replace Jest with Karma, as:
- Karma seems way more stable/better maintained
- Karma runs with mocha/sinon/chai which are way better maintained than the stuff jest comes with
- Karma runs in browsers, which is cool!


martijnrusschen:
lgtm

PR #144, opened 2015-07-29
Remove example in favour of gh-pages

martijnrusschen:
New demo live here now: https://hacker0x01.github.io/react-datepicker


ntdb:
This change breaks local development according to the README

martijnrusschen:
Sorry you're right. Not completely sure what the right thing is to do local dev and maintaining the gh-pages, but for now I'm relying on tests to build now functionality. Any ideas how to fix this?

ntdb:
My solution is to pull both branches and reconfigure the `gh-pages` `package.json` file to point to my local version of `master`. It's not pretty but it works.

PR #145, opened 2015-07-29
Basics of improving example

martijnrusschen:
Turns the old fashioned example page into a fancy example overview with code examples.

screen shot 2015-07-30 at 13 58 11

### DIFF
--- a/src/style.scss
+++ b/src/style.scss
@@ -15,7 +15,7 @@ html, body {
   line-height: 21px;
 
   input {
-    width: 350px;
+    width: 100%;
   }
 }
 
@@ -28,6 +28,7 @@ html, body {
 
 .example {
   display: flex;
+  margin-bottom: 20px;
 }
 
 .column {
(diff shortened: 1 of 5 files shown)


src/style.scss
```diff
@@ -0,0 +1,50 @@
...
+  border-radius: 4px;
+  border: 1px solid #e4e4e4;
+  padding: 20px;
+
+  &:first-child {
+    margin-right: 20px;
+  }
+}
+
+.example__heading {
```
lbekkema: I think we can call this `heading` as we only have this type of headings (for now).

src/topbar.scss
```diff
@@ -0,0 +1,18 @@
+.topbar__wrapper {
```
lbekkema: Let's rename this class to `topbar`.

src/topbar.scss
```diff
@@ -0,0 +1,18 @@
...
+  position: fixed;
+  z-index: 5;
+  width: 100%;
+  height: 70px;
+  background-color: #216ba5;
+}
+
+.topbar {
```
lbekkema: And this one to `topbar__wrapper`, so this looks a little bit more logical in HTML.

Issue #141, opened 2015-07-27
IE10 freeze


aferrandi:
If I run the example:

http://cdn.rawgit.com/Hacker0x01/react-datepicker/master/example/index.html

the 3rd line (the one with no date already selected) makes internet explorer 11 freeze when I try to choose a date in the datepicker.
Firefox and Chrome don't have any problem.

sconno05:
I ran into this with IE 9 and 10 as well.  It looks like it was introduced with v0.11 (I was not able to reproduce with v0.10).  I verified that this was not caused by my pull request #135.  I would appreciate a fix ASAP, I've been waiting on v0.11.

RSO:
I will check it out, however, as of yet I have no way to test in IE, I'll have to look into that as well.

RSO:
I'll have to download an IE11 vm for my mac, so that'll take some time. If somebody else could provide me with a trace/screenshot/whatever it would mean a lot to me!

sconno05:
There are no errors being console logged as far as I can see.  I am attaching a screenshot, the steps to reproduce are:

1. Open IE
2. Go to http://cdn.rawgit.com/Hacker0x01/react-datepicker/master/example/index.html
3. Click the first input field
4. Select a date
5. IE locks up, must be shut down

![image](https://cloud.githubusercontent.com/assets/2985457/8942377/4542d3c4-3532-11e5-819b-c4f8dfc07bc5.png)

RSO:
Fixed in dad57db772c492b9446e78da183f6a7984e6f1a5. @martijnrusschen will do a release. Thanks for reporting!

RSO:
Now I'll have to figure out how to run tests in sauce-labs, and write a test for this in IE!

martijnrusschen:
Version 0.11.2 has been released!

sconno05:
Thanks for the quick turn around guys.

sconno05:
I am unfortunately still seeing this issue in IE10.  IE9 seems to be working.  I was able to reproduce after upgrading locally and on the examples page.

martijnrusschen:
Debugging here #150, would be cool if we could reproduce these issues in a test.

sconno05:
It looks like this is still not working in IE10 as of 0.11.3. I unfortunately will not be able to debug until sometime next week but I would really appreciate a fix.

truckingsim:
[...]

PR #150, opened 2015-07-31
Run IE10

martijnrusschen:
Testing to see if tests are breaking in IE 10


martijnrusschen:
I added an IE10 test runner to check if this is failing because of IE10. Trying to reproduce the issues noted in #141.

martijnrusschen:
Adding this runner so we can catch this bug as soon as we're adding more tests

PR #152, opened 2015-07-31
Make it all mobile proof

lbekkema:


lbekkema:
@martijnrusschen Can you take a look at making the `style.css` file? I wasn't able to create one.

PR #159, opened 2015-08-11
Control popover location

truckingsim:
The company I work for needed the ability for the datepicker to popup.  Instead of just adding a popup property I decided to let the user define the 3 tether properties, `targetAttachment`, `attachment`, and `targetOffset` themselves this way they can make it popup wherever they want.

I'm not sure if you guys want this if you do there is one thing that needs to be handled that I wasn't sure how you would want it handled.  That is the arrow at the top.  As soon as the datepicker leaves the bottom of the target that arrow is now pointing at the wrong thing.  Any suggestions/guidance as to how you would like to handle that would be great.


truckingsim:
The last travis build failed due to sauce labs failing.  It couldn't start up a browser to do the karma tests.

Just an FYI as to why the build failed.

src/datepicker.jsx
```diff
@@ -123,6 +130,7 @@ var DatePicker = React.createClass( {
             maxDate={this.props.maxDate}
             excludeDates={this.props.excludeDates}
             weekStart={this.props.weekStart} />
+
```
martijnrusschen: This newline isn't needed.

src/datepicker.jsx
```diff
@@ -110,7 +113,11 @@ var DatePicker = React.createClass( {
   calendar: function() {
     if ( this.state.focus ) {
       return (
-        <Popover>
+        <Popover
+          attachment={this.props.popoverAttachment}
```
martijnrusschen: Would it make sense to add some defaultProps to the popover component? Not sure if you need those properties by default?

martijnrusschen:
Hi, thanks for taking the time to look into this! I added some comments and tried to answer your question. Hope you can make something out of this.

PS. Don't worry about failing travis, looks like some unreliable saucelabs stuff.

> I'm not sure if you guys want this if you do there is one thing that needs to be handled that I wasn't sure how you would want it handled. That is the arrow at the top. As soon as the datepicker leaves the bottom of the target that arrow is now pointing at the wrong thing.

This was implemented recently to position the popover at the top of the input field if there's no space below the input.

src/datepicker.jsx
dcousineau: @martijnrusschen hah, we actually had some back and forth on this today and IIRC were relying on an unset `this.props.popoverAttachment` being `undefined` (and thus allowing the defaultProps), however @truckingsim did all the work so I'll let him chime in when he wakes back up (I may have some wires crossed here).

martijnrusschen: Np, Lets wait for the final verdict and I'm happy to merge it.

truckingsim: @martijnrusschen as @dcousineau said we had some back and forth on this.  I really wanted to leave the default values for popover in the popover component, so I did.  React will use the default value if you pass undefined as the prop value.  So `this.props.popoverAttachment` in `datepicker.jsx` has no default, and if not passed will return undefined.  In `popover.jsx` I moved the values for the three tether properties to props defaults.  So now `this.props.attachment` and the other two properties will use those if people use the datepicker without those props ([see this line in the diff](https://github.com/Hacker0x01/react-datepicker/pull/159/files#diff-d1829fe907ab3e0fb05f2d87d36342fdR12) for the defaults).

So yeah, basically everything should work as it did before but with the added option of defining the location of the popover.

martijnrusschen:
Thanks, merged the PR and will do a release soon!

PR #162, opened 2015-08-13
Bugfix ie infinite loop

dcousineau:
This PR does 2 specific things IRT to fixing #141 

It makes date_input.jsx properly state-less, reducing complexity.

It also upgrades datepicker.jsx to prevent repeated calls clearing the state if it has already been cleared. IE would trigger an onChange every time we set its value to null, causing the clearSelection to bubble up, set state, set the value, and repeat ad nauseam. This also upgrades datepicker.jsx to follow best practices of using setState callbacks to prevent subtle timing bugs.

----

### Notes for the future.

This bug was caused by a loop of competing states (and IE being overzealous about change events). Current best practice in React would be to eliminate the usage of state to track value and instead mimic a managed component (e.g. `<input value={} onChange={} />`). This PR is just a bugfix only without changing too much, but it could be done easily with 1 exception: a BC break by eliminating passing the value in the `onBlur` callback.


dcousineau:
Looks like Travis is still barfing on browser launching

PR #164, opened 2015-08-14
Reduce Size via lodash require changes

bceskavich:
**tl;dr** - Moving from full lodash requires to method-specific requires reduces react-datepicker's size in production by 84% from 466kb to 75kb.

All tests should pass, and I'm using this fix in production currently

--

While working with react-datepicker I noticed it was contributing to a good portion of my final bundle size, over 400kb in total and the second largest dependency after React proper. I was able to trace this issue back to the practice of requiring the entire lodash library (i.e. `var _ = require("lodash")` ), which in turn leads to the entirety of the (very large) lodash library to be included in one's final bundle.

The fix is pretty simple, just requiring the specific methods needed from lodash rather than the entire library. It doesn't seem like this project needs any of the benefits of a full lodash require (i.e. chaining) and lodash itself points to this approach as a better practice.

I used [webpack-bundle-size-analyzer](https://github.com/robertknight/webpack-bundle-size-analyzer) to track down the size changes.

**As Is Today**
react-datepicker is the second biggest dependency, 86% of which is from lodash
```shell
$ analyze-bundle-size bundle-stats.json 
react: 569.4 kB (28.0%)
  <self>: 569.4 kB (100%)
react-datepicker: 466.84 kB (23.0%)
  lodash: 401.81 kB (86.1%)
    <self>: 401.81 kB (100%)
  tether: 50.46 kB (10.8%)
    <self>: 50.46 kB (100%)
  react-onclickoutside: 3.62 kB (0.775%)
    <self>: 3.62 kB (100%)
  <self>: 10.95 kB (2.34%)
```

**Method-Specific Requires**
Reduced in size by 84%
```shell
$ analyze-bundle-size bundle-stats.json 
react: 569.4 kB (34.7%)
  <self>: 569.4 kB (100%)

...

react-datepicker: 75.03 kB (4.58%)
  tether: 50.46 kB (67.3%)
    <self>: 50.46 kB (100%)
  react-onclickoutside: 3.62 kB (4.83%)
    <self>: 3.62 kB (100%)
  <self>: 20.94 kB (27.9%)

```


martijnrusschen:
Awesome, this might solve some problems I've been experiencing with Lodash too.

PR #174, opened 2015-08-21
PropTypes Added to Calendar and Datepicker Component and Readme Updated

jwdotjs:
Add proptypes to Calendar and Datepicker component so it’s a bit
clearer about what is accepted by them. Also update readme to specify
that weekStart is an acceptable configuration option for the Datepicker


martijnrusschen:
Great improvements, thanks!

PR #176, opened 2015-08-22
Added readonly

liamhubers:
When using the datepicker on mobile the keyboard will pop up when selecting a date, using readonly will prevent this from happening.


martijnrusschen:
Thanks!

PR #183, opened 2015-09-13
remove note about server rendering

vladikoff:
Seems like this was removed in https://github.com/Hacker0x01/react-datepicker/pull/144/files


martijnrusschen:
This was moved to Github pages, removing for now since we're working on new documentation.

PR #185, opened 2015-09-18
Modifications to allow tether constraints to be passed in via props

SpaceMacGyver:
Due to bizarre layout issues in the application in which the **DatePicker** component is being consumed, it was necessary for me to modify constraints.  To better facilitate this for future users, I've modified constraints so that it can be passed in via `props` from the parent component (as `props.tetherConstraints` when creating the **DatePicker**, and internally as `props.constraints` when passing from **DatePicker** to **Popover**).

Example usage:

```jsx
var constraints = [{ to: "window" }],
    selected = moment(),
    onChange = function () { };
<DatePicker tetherConstraints={constraints} selected={selected} onChange={onChange} />
```


martijnrusschen:
Thanks, sorry for the delay!

montogeek:
@SpaceMacGyver How do you select a DOM element for the `to` option?

Issue #179, opened 2015-08-24
Server-Side Rendering doesn't work


sherodtaylor:
``` bash
webapp_1     | /code/node_modules/react-datepicker/node_modules/tether/dist/js/tether.js:406
webapp_1     |   var el = document.createElement('div');
webapp_1     |            ^
webapp_1     | ReferenceError: document is not defined
webapp_1     |     at /code/node_modules/react-datepicker/node_modules/tether/dist/js/tether.js:406:12
webapp_1     |     at /code/node_modules/react-datepicker/node_modules/tether/dist/js/tether.js:415:3
webapp_1     |     at i (/code/node_modules/react-datepicker/node_modules/tether/dist/js/tether.js:7:22)
webapp_1     |     at Object.<anonymous> (/code/node_modules/react-datepicker/node_modules/tether/dist/js/tether.js:11:2)
webapp_1     |     at Module._compile (module.js:460:26)
webapp_1     |     at Object.require.extensions.(anonymous function) [as .js] (/code/node_modules/node-jsx/index.js:26:12)
webapp_1     |     at Module.load (module.js:355:32)
webapp_1     |     at Function.Module._load (module.js:310:12)
webapp_1     |     at Module.require (module.js:365:17)
webapp_1     |     at require (module.js:384:17)
```

tcwilson:
I'm having this same issue.  Also see #137

I've tried to find a workaround using webpack, but have not been able to.  Currently there is code that attempts to delay the tether require, if `window` and `document` are undefined.  https://github.com/Hacker0x01/react-datepicker/blob/master/src/popover.jsx#L70

```javascript
    } else if ( window && document ) {
      var Tether = require( "tether" );
```

Unfortunately, it seems webpack looks at all `require`s, bundles them up, and evaluates the code at once.  I don't know enough about webpack to go further than this right now.

sujaykathrotia:
This issue is resolved now. Please check the merged PR where this issue was referenced.

PR #190, opened 2015-09-25
Move require

martijnrusschen:
Makes more sense IMO


RSO:
We can't do it like this. The require was placed inside of the if because otherwise serverside rendering would fail (at the moment of writing that code Tether did not support server side rendering).

martijnrusschen:
Hmm, okay. Looks like its still failing for people: #179

RSO:
Yeah we should probably set up automated testing if we want to keep supporting SSR

PR #211, opened 2015-10-27
fixes for react 0.14

PanJ:
Note: This will break previous version of react


martijnrusschen:
Thanks, this is great!

PR #79, opened 2015-02-12
Show range of selected days in calendar

lbekkema:
This diff adds the ability to show the range of days selected in the calendar when a start and end date is given. 

Example:

screen shot 2015-02-12 at 14 04 29

I used the [`moment-range`](https://github.com/gf3/moment-range) library, so it's easier to determine if a day is within the range. I wasn't able to achieve this with `moment-js`'s `isBetween` method. 

Let me know what you guys think!

PR #212, opened 2015-10-29
Allow for date range selection

xzyfer:
This patch is a fork of #79 with the suggested modifications and
merge conflicts resolved. The credit should go to the original
author. This is simply something I wanted and the original author
appears to have gone MIA.

### DIFF
--- a/src/day.jsx
+++ b/src/day.jsx
@@ -22,12 +22,14 @@ var Day = React.createClass( {
     if ( this.props.day.sameDay( this.props.selected ) )
       classes.push( "datepicker__day--selected" );
 
+    if ( this.props.day.inRange )
+      classes.push( "datepicker__day--in-range" );
+
     if ( this.props.day.sameDay( moment() ) )
       classes.push( "datepicker__day--today" );
 
-    if ( this.isWeekend() ) {
+    if ( this.isWeekend() )
       classes.push( "datepicker__day--weekend" );
-    }
 
     return (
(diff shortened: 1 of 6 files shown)


xzyfer:
Refactored and added test coverage for the `inRange` detection.

RSO:
:+1: Awesome stuff, thanks!

src/day.jsx
```diff
@@ -22,12 +22,14 @@ var Day = React.createClass( {
     if ( this.props.day.sameDay( this.props.selected ) )
       classes.push( "datepicker__day--selected" );
 
+    if ( this.props.day.inRange )
```
rafeememon: I believe this should be `this.props.inRange`; `day.inRange` is a method that will always be defined/truthy

xzyfer: You could be right. I mostly just rebased the existing PR. Without the
example page I don't have an environment to easily run this component so
I'm relying on the tests.

IMHO bringing back the example page would really help with contributions.

rafeememon: Hey @xzyfer, no prob -- I too was finding myself stumbling to find the right way to bootstrap this.  I'm working on a PR that will begin a test suite for the Day component.

martijnrusschen:
Having some issues with creating an example for this awesome thing. Mind looking into it what's going wrong? https://github.com/Hacker0x01/react-datepicker/pull/219

xzyfer:
Sorry, timezones. It's late here I'll happily take a look in the morning.

PR #214, opened 2015-10-29
startDate and endDate props should not be required

xzyfer:
This PR fixes errors being thrown when either `startDate` or `endDate`
props are not supplied.

Fixes https://github.com/Hacker0x01/react-datepicker/commit/9715c69f4ad68866de35c350a6b0c42c5bd6cc0e#commitcomment-14071280

/cc @Reggino

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -16,6 +16,8 @@ var Calendar = React.createClass( {
     hideCalendar: React.PropTypes.func.isRequired,
     minDate: React.PropTypes.object,
     maxDate: React.PropTypes.object,
+    startDate: React.PropTypes.object,
+    endDate: React.PropTypes.object,
     excludeDates: React.PropTypes.array,
     weekStart: React.PropTypes.string.isRequired
   },
--- a/src/util/date.js
+++ b/src/util/date.js
@@ -19,9 +19,10 @@ DateUtil.prototype.sameMonth = function( other ) {
 };
 
 DateUtil.prototype.inRange = function( startDate, endDate ) {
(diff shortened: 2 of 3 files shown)


src/util/date.js
```diff
@@ -19,9 +19,10 @@ DateUtil.prototype.sameMonth = function( other ) {
 };
 
 DateUtil.prototype.inRange = function( startDate, endDate ) {
-  var startDate = startDate._date.startOf( "day" ).subtract( 1, "seconds" );
-  var endDate = endDate._date.startOf( "day" ).add( 1, "seconds" );
-  return this._date.isBetween( startDate, endDate );
+  if ( !startDate || !endDate ) return false;
```
martijnrusschen: I think `return unless startDate || endDate` might look cleaner but thanks for fixing this issue!

Reggino:
Thanks for this!

src/calendar.jsx
```diff
@@ -16,6 +16,8 @@ var Calendar = React.createClass( {
     hideCalendar: React.PropTypes.func.isRequired,
     minDate: React.PropTypes.object,
     maxDate: React.PropTypes.object,
+    startDate: React.PropTypes.object,
+    endDate: React.PropTypes.object,
```
np: Are these properties, really?

Judging from the lack of apparent effect of setting startDate or endDate on the DatePicker component and a quick look at the code, these two lines could be removed.

rafeememon: Yep, they are -- setting them in conjunction results in a range being selected

Issue #194, opened 2015-09-28
CSS Not Loading - Redux Hot Example


yomolify:
Trying to integrate this beautiful datepicker into this very popular React-Redux boilerplate https://github.com/erikras/react-redux-universal-hot-example. The input box appears and it is possible to select dates but the styling does not load at all. I have tried the canusedom approach from https://github.com/Hacker0x01/react-datepicker/issues/137 but the most it does is not give the "document is not defined" error. Screenshot is attached to show the kind of component that gets rendered. 
<img width="225" alt="screen shot 2015-09-28 at 12 03 42 am" src="https://cloud.githubusercontent.com/assets/3376555/10129563/6b0074be-6574-11e5-8338-9f8856a47476.png">
Clicking on any of the numbers does select the right date, but the nicely styled date selector component is not rendered. Is there a way to make this render as it is supposed to?

eversadia:
+1

jasongonzales23:
Seeing the same issue. Using React 0.14.0 and 0.15.1 of the datepicker.

jasongonzales23:
Also, this appears to have nothing to do with HMR. I am using redux, but it shouldn't matter since I am just trying to init a basic version of the component with hardcoded values.

Switching to Zippy UI for now until you get this resolved.

jasongonzales23:
Oh holy wow. Here's the fix. Add this:

`require('react-datepicker/dist/react-datepicker.css')`

Why is this not in the documentation!!!

martijnrusschen:
Oh that might be an oversight from our side! Sorry about that.. Would be
awesome if someone could contribute to the documentation with their
experiences.

jasongonzales23:
Happy to add to the docs, will squeeze it in here in a sec, but now noticing another bug. The datepicker stays display: none right now. I can manipulate in the dev console and then it works until page refresh.

Will create a new issue for this.

martijnrusschen:
Thanks!

ctrlaltdylan:
Even with importing the CSS the datepicker has the same issue for me :disappointed: Playing around with it

yasserzubair:
any solution for this yets?

stphnlngdncoding:
If anyone is using CSS Modules, you have to import ```react-datepicker/src/stylesheets/datepicker-cssmodules.scss```

yoavgecht:
I'm using CSS Modules
I import react-datepicker/src/stylesheets/datepicker-cssmodules.scss
[...]

PR #215, opened 2015-10-29
explain CSS inclusion

jasongonzales23:


jasongonzales23:
This addresses https://github.com/Hacker0x01/react-datepicker/issues/194

PR #223, opened 2015-11-06
Datepicker reopens after clicking on the input when the datepicker is already open

liamhubers:
We're having some problems with people clicking on the input to close to datepicker ( mainly on mobile ) however right now the datepicker will close and reopen directly after.


liamhubers:
Not quite sure why I'm getting conflicts..

martijnrusschen:
Did you try to rebase with the latest master? Looks like you're still running against an old master.

liamhubers:
Fixed it :)

src/datepicker.jsx
```diff
@@ -110,10 +109,18 @@ var DatePicker = React.createClass( {
...
+    if (!this.state.virtualFocus) {
+      return this.setState({
+        focus: true,
+        virtualFocus: true
+      });
+    }
+    this.setState({ virtualFocus: false });
+  },
+
+  onClearClick: function( event ) {
```
rafeememon: Looks like this wasn't merged properly

liamhubers: Merged correctly, seeing anything weird I'm not seeing?

rafeememon: This adds a duplicated definition of `onClearClick`, which is defined below this diff

liamhubers: Nice catch, thanks!

PR #224, opened 2015-11-06
Allow id attribute to be set

theSmaw:
I would like to be able to provide an id attribute to the text input created by react-datepicker, so that I can associate the input with a label.

With this PR, a DatePicker component created with the following:

```
<DatePicker id="date-picker" />
```

Creates the following input element:

```
<input type="text" id="date-picker" class="datepicker__input" data-reactid="blah">
```


theSmaw:
Fixes https://github.com/Hacker0x01/react-datepicker/issues/225

PR #228, opened 2015-11-10
Use moment defaults and preserve global settings.

### LINKED ISSUE #178: Use moment.weekdaysMin() for weekdays
Considering that `moment` is used for formatting of date-related stuff in datepicker (e.g. `month+year` in calendar header), why is it not used to obtain weekdays?

I am currently setting it as

    const weekdays = moment.weekdaysMin()
    <Datepicker weekdays={weekdays} />

I assume this would be sensible default - also taking into account `locale` property of `<Datepicker />`?

Current implementation also overrides moment locale setting with `weekdays` as set in `Datepicker` (https://github.com/Hacker0x01/react-datepicker/blob/master/src/calendar.jsx#L59), which can be rather surprising for the user of the component.

mancontr:
This fixes #178 by making the weekday, start of week, and locale default to the moment ones, and making props override them only for the moment instance we keep as state, instead of globally.


martijnrusschen:
Could you please rebase your diff to make sure the changes are only affecting new code?

martijnrusschen:
There's no activity on this PR anymore. This issue will be addressed in #277

PR #230, opened 2015-11-10
Add disabled to proptypes

ryanwilsonperkin:
Let me know if there's a place in the docs this should also be updated.


ryanwilsonperkin:
Fixes #230

martijnrusschen:
Thanks!

PR #231, opened 2015-11-11
Change default dateFormat from YYYY-MM-DD to YYYY/MM/DD

### LINKED ISSUE #154: Consider changing default date format
Currently the default date format is `YYYY-MM-DD`, while a more frequently used format is `YYYY/MM/DD`. Let's consider making that the default.

lbekkema:
Fixes #154


martijnrusschen:
Not sure if we should do this.

rafeememon:
Agreed -- this is a breaking change

martijnrusschen:
Yep, don't think it's worth it at this moment.

Issue #160, opened 2015-08-12
calendar displaying incorrect dates


benrlodge:
Great component thanks, just noticed an issue though. As seen below, July 25, 2015 should be a Saturday, not a Sunday. Sometimes I'll see it correct, but not always. FWIW the date that is saved is correct, just what's displaying on the calendar. 

![screen shot 2015-08-11 at 9 25 56 pm](https://cloud.githubusercontent.com/assets/3275453/9214762/c179c554-406f-11e5-89bf-e2eec749b1c3.png)

![screen shot 2015-08-11 at 9 32 17 pm](https://cloud.githubusercontent.com/assets/3275453/9214814/7fe443b6-4070-11e5-8044-af7359ca0510.png)

blackjk3:
We are also experiencing the same behavior.  @benrlodge were you thinking about submitting a PR? Otherwise, we could look into it at well.

benrlodge:
@blackjk3 this was just for a personal project so I haven't revisited it, likely won't in the future.

GalacticWeb:
+1 !!!

gl-kunalgupta:
+1

martijnrusschen:
Did someone track down the cause of this issue? Would love to fix this.

ge3kusa:
We use this in a few spots and one instance renders the dates correctly. The other wouldn't until I forced the start of the week to be Sunday by passing weekStart="0" as a param. It cleared up after that. I'm not calling that a solution but it might lead to one in the future. At the very least it got us past the issue.

dnl:
I used this as a hacky workaround.

```
  // moment.weekdaysMin is [sun, mon, tue...]
  // DatePicker.weekdays needs [sat, sun, mon...] if weekStart=1
  //                           [sun, mon, tue...] if weekStart=0
  // if this isn't used it will label the first column 'tuesday' for 'starts on monday' locale.
  // DatePicker appears to be doing the offset twice.
  weekDays () {
    var offset = moment.localeData().firstDayOfWeek();
    var weekdays = moment.weekdaysMin();
    if (offset > 0) {
      var start = weekdays.slice(7-offset);
      var end = weekdays.slice(0,-offset);
      return end.concat(start);
    } else {
      return weekdays;
    }
  }
```

rafeememon:
Probably fixed by #277

PR #257, opened 2015-12-13
Updated proposal: Use moment defaults and preserve global settings.

### LINKED ISSUE #178: Use moment.weekdaysMin() for weekdays

martijnrusschen:
Updated version of #228
Should fix #178

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -54,8 +54,10 @@ var Calendar = React.createClass({
 
   localizeDateUtil(dateUtil) {
     var thisMoment = dateUtil.moment();
-    if (this.props.locale) thisMoment.locale(this.props.locale);
-    if (this.props.weekStart) thisMoment._locale._week.dow = this.props.weekStart;
+
+    if (this.props.locale) { thisMoment.locale(this.props.locale); }
+    if (this.props.weekStart) { thisMoment._locale._week.dow = this.props.weekStart; }
+
     return dateUtil;
   },
 
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -9,7 +9,7 @@ var Calendar = React.createClass({
   mixins: [require("react-onclickoutside")],
 
   propTypes: {
-    weekdays: React.PropTypes.array.isRequired,
+    weekdays: React.PropTypes.array,
     locale: React.PropTypes.string,
     moment: React.PropTypes.func.isRequired,
     dateFormat: React.PropTypes.string.isRequired,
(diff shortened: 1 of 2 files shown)


martijnrusschen:
We need better tests.. Build passes, but examples are failing.

rafeememon:
I have high hopes that this might also fix #160

martijnrusschen:
Care to look into this? I ported the changes from #228, but it doesn't work for me.

rafeememon:
I'll be traveling over the next week or so, but I can take a crack at it after that!

src/calendar.jsx
```diff
@@ -41,31 +41,24 @@ var Calendar = React.createClass({
...
+    var thisMoment = dateUtil.moment();
 
-    this.props.moment.locale(this.props.locale, {
-      week: {
-        dow: this.props.weekStart
-      },
-      weekdaysMin: weekdays
-    });
+    if (this.props.locale) { thisMoment.locale(this.props.locale); }
+    if (this.props.weekStart) { thisMoment._locale._week.dow = this.props.weekStart; }
```
rafeememon: Unfortunately this line is problematic.  Modifying `_locale` touches the global locale, which we want to avoid.  I found this out by trying to change `_locale._weekdaysMin` in one picker and seeing that it affected all pickers.

rafeememon:
Found some time to dig into this on the train, and came away feeling that this will be extremely fragile.  Supporting all of `locale`, `weekdays`, and `weekStart` is a fight against moment.js when we try not to modify global state.  I think there are three options here:

1) Fix this up and accept that we're still modifying global state, even though we're trying to avoid that.

2) Wait for a feature like https://github.com/moment/moment/pull/2774, which would effectively let us subclass the global locale with the modifications that we want.

3) Remove the `locale`, `weekdays`, and `weekStart` options entirely, and defer to the global moment.js handling of these.  I figure users will be localizing their moment configurations anyway, but this removes the ability to have picker-specific locale configuration.  Then again, picker-specific locales seem a bit contrived to me.

martijnrusschen:
Replaced by #277

PR #258, opened 2015-12-13
Show valid date range in initial view

### LINKED ISSUE #209: valid date range is not in the initial view
1, don't pass selected
2, pass min date one year later
3, pass max date after min date

Then: If I don't know the valid date range, I can't select a date.

rafeememon:
Fixes #209

If minDate is after the current date, the focus is set to minDate, and if maxDate is before the current date, the focus is set to maxDate.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -37,7 +37,10 @@ module.exports = function(config) {
         new webpack.DefinePlugin({
           "process.env.NODE_ENV": JSON.stringify("test")
         })
-      ]
+      ],
+      resolve: {
+        extensions: ["", ".jsx", ".js"]
+      }
     },
 
     webpackServer: {
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -10,7 +10,7 @@ var Calendar = React.createClass({
 
   propTypes: {
     weekdays: React.PropTypes.array.isRequired,
-    locale: React.PropTypes.string,
+    locale: React.PropTypes.string.isRequired,
     moment: React.PropTypes.func.isRequired,
     dateFormat: React.PropTypes.string.isRequired,
     onSelect: React.PropTypes.func.isRequired,
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -5,6 +5,17 @@ import YearDropdown from "./year_dropdown";
 import Day from "./day";
 import React from "react";
 
+function getDateInView({ moment, minDate, maxDate }) {
+  var current = new DateUtil(moment());
+  if (minDate && new DateUtil(minDate).isAfter(current)) {
+    return minDate;
+  } else if (maxDate && new DateUtil(maxDate).isBefore(current)) {
+    return maxDate;
+  } else {
+    return current.moment();
+  }
+}
+
 var Calendar = React.createClass({
   mixins: [require("react-onclickoutside")],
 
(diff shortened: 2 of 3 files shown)


karma.conf.js
```diff
@@ -37,7 +37,10 @@ module.exports = function(config) {
         new webpack.DefinePlugin({
           "process.env.NODE_ENV": JSON.stringify("test")
         })
-      ]
+      ],
+      resolve: {
+        extensions: ["", ".jsx", ".js"]
```
martijnrusschen: Not sure what this actually does, but do you need to add `""` here? Or does `[".jsx", ".js"]` work too?

src/calendar.jsx
```diff
@@ -30,8 +30,18 @@ var Calendar = React.createClass({
   },
 
   getInitialState() {
+    function getDateInView(props) {
```
martijnrusschen: Does it make sense to have this function in `getInitialState`, what about moving this out of it, and just call it from the `getInitialState`?

test/calendar_test.js
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+import TestUtils from "react-addons-test-utils";
+import moment from "moment";
+import DateUtil from "../src/util/date";
+import Calendar from "../src/calendar";
+
+describe("Calendar", function() {
```
martijnrusschen: +1 for adding tests!

src/calendar.jsx
```diff
@@ -10,7 +10,7 @@ var Calendar = React.createClass({
 
   propTypes: {
     weekdays: React.PropTypes.array.isRequired,
-    locale: React.PropTypes.string,
+    locale: React.PropTypes.string.isRequired,
```
martijnrusschen: Is this needed?

karma.conf.js
rafeememon: Adding `""` makes it so files required with their extension (e.g. `import "./calendar.jsx"`) can be resolved.  It's a default in webpack and I figured it would be good to keep around.

src/calendar.jsx
rafeememon: Not explicitly, but I ran into it when constructing `Calendar` components in my test and thought it would be good to update this.  It's in a separate commit that I can remove if you'd like.

src/calendar.jsx
rafeememon: Haha, yeah, I was more or less just following the (bad) style from another `getInitialState`.  Will fix.

src/calendar.jsx
martijnrusschen: No problem, as long as it doesn't break the current implementations :)

karma.conf.js
martijnrusschen: Adding @RSO here since he did the initial Karma setup.

src/calendar.jsx
martijnrusschen: :+1:

src/calendar.jsx
```diff
@@ -5,12 +5,23 @@ import YearDropdown from "./year_dropdown";
 import Day from "./day";
 import React from "react";
 
+function getDateInView(props) {
```
RSO: This would probably look nice with destructuring:

```
function getDateInView({ moment, minDate, maxDate)
```

RSO:
Apart from that small stylistic suggestion this looks good to me!

src/calendar.jsx
rafeememon: Good call -- in fact `moment` should probably be left as is because it's the actual `moment` library being passed in as a prop

martijnrusschen:
Thanks!

rafeememon:
Awesome, thanks for the CR!

PR #261, opened 2015-12-14
Properly hide calendar on blur

### LINKED ISSUE #156: Problem with tabbing navigation
If you open the demo and press Tab to move focus from one DatePicker to another, the calendar of the original DatePicker stays open. I think that by default it should hide the same way it would when you click outside of the calendar.

rafeememon:
Fixes #156 

It seemed fishy that `hideCalendar` was only called when an `onBlur` function is passed in as a prop.  I could have simply refactored the `if` statement, but I decided to follow the style of `onFocus` by having a default no-op `onBlur`.

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -32,7 +32,8 @@ var DatePicker = React.createClass({
       moment: moment,
       onChange() {},
       disabled: false,
-      onFocus() {}
+      onFocus() {},
+      onBlur() {}
     };
   },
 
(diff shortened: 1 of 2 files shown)


martijnrusschen:
Great work! Looks like you're on a hunt to track down all the issues :)

martijnrusschen:
Wondering if there's a way to test this.. I have a feeling that the hideCalendar stuff is really fragile and would love to have some tests.

martijnrusschen:
But that shouldn't block this PR of course. Just looking for a way to improve our coverage on this.

rafeememon:
Done!  There's definitely a lot of room for tests to be written on the top-level component

PR #265, opened 2015-12-15
Add option to hide year dropdown menu

rafeememon:
In many of my use cases, the user will be selecting dates near the present and will rarely need to navigate several years.  Having the year dropdown menu on a separate line takes up valuable screen real estate and IMO somewhat spoils the sleek look of the date picker.  This PR adds a `showYearDropdown` option to toggle its display (default true).

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -169,6 +169,18 @@ var Calendar = React.createClass({
     });
   },
 
+  renderCurrentMonth() {
+    var classes = ["datepicker__current-month"];
+    if (this.props.showYearDropdown) {
+      classes.push("datepicker__current-month--hasYearDropdown");
+    }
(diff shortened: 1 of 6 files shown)


src/calendar.jsx
```diff
@@ -175,13 +200,8 @@ var Calendar = React.createClass({
           <a className="datepicker__navigation datepicker__navigation--previous"
               onClick={this.decreaseMonth}>
           </a>
-          <h2 className="datepicker__current-month">
```
rafeememon: There seemed to be a dependency on the example page's `h2` margins for this to look right, which I ran into when I removed the dropdown

martijnrusschen:
Valid point. Thanks for adding tests and an example!

PR #266, opened 2015-12-16
Fix positioning of the popover when the popover is shown on top

### LINKED ISSUE #193: Incorrect popover position
if is no scrolling then the popover position little different. If has scrolling then everything works as it should.

martijnrusschen:
Fixes #193


martijnrusschen:
@rafeememon This should fix the issue described in #193

rafeememon:
LGTM :+1:

PR #268, opened 2015-12-17
Don't modify dates when calling inRange

### LINKED ISSUE #267: Date Range doesn't work in Demo
I want to select a date range, from ... to .... Like in hotel/tickets booking sites: 

687474703a2f2f692e696d6775722e636f6d2f4c62414d6633442e706e67

I check Demo http://hacker0x01.github.io/react-datepicker/#example-8 but it seems it doesn't work, does it? I'd like to see the mechanic. Or it only doesn't work in Demo?

rafeememon:
Fixes #267 

The `inRange` method is modifying the input dates, which cascades into other failures.


martijnrusschen:
LGTM!

PR #269, opened 2015-12-17
Make date range selection example more realistic

### LINKED ISSUE #267: Date Range doesn't work in Demo

rafeememon:
Fixes #267 some more

This changes the `handleChange` method for the date range example to be a more plausible use case

### DIFF
--- a/docs/src/examples/date_range.jsx
+++ b/docs/src/examples/date_range.jsx
@@ -12,10 +12,25 @@ var DateRange = React.createClass({
     };
   },
 
-  handleChange: function(date) {
-    this.setState({
-      startDate: date
-    });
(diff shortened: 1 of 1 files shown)


martijnrusschen:
We might want to consider showing the example with 2 Datepickers, I guess that's the most used situation. What's your take on that?

rafeememon:
Good idea, pushing a fix now

Issue #272, opened 2015-12-18
Clicking too fast when changing months or years on the date picker causes it to close


benmliang:
I think it's because user clicked on non-Calendar area when the UI tries to change, and then turn off the Calendar. Is there any quick fix?

martijnrusschen:
Should be fixed with #273

benmliang:
Thanks a lot :+1:

PR #273, opened 2015-12-19
Fix bug which closes calendar if multiple clicks are made in quick succession.

lucygilbert:
I found this bug while using the date picker and fixed it in my fork. I thought you might want to include the fix.

It's been logged as bug [#272](https://github.com/Hacker0x01/react-datepicker/issues/272).


src/datepicker.jsx
```diff
@@ -42,7 +43,7 @@ var DatePicker = React.createClass({
   getInitialState() {
     return {
       focus: false,
-      virtualFocus: false,
+      triggerUpdate: false,
```
rafeememon: It looks like `triggerUpdate` might be unused, does it have a purpose that I'm missing?

src/datepicker.jsx
```diff
@@ -63,15 +64,12 @@ var DatePicker = React.createClass({
 
   handleFocus() {
     this.props.onFocus();
-    this.setState({
```
rafeememon: This removes the ability to open the datepicker by tabbing into the input =(  Is there a way to make this work with your fix?

test/datepicker_test.js
```diff
@@ -10,7 +10,7 @@ describe("DatePicker", () => {
       <DatePicker />
     );
     var dateInput = datePicker.refs.input;
-    TestUtils.Simulate.focus(ReactDOM.findDOMNode(dateInput));
```
rafeememon: I think we want this to stay as `focus` so that we can test that tabbing into the input opens the picker

src/datepicker.jsx
lucygilbert: triggerUpdate is just used to force a re-render, because the state has to have changed for the component to update.

It could probably be refactored into a forceUpdate, but I noticed that the code doesn't currently use a forceUpdate anywhere, and there are people who are totally against using forceUpdate.

If a forceUpdate is okay, then that could be used instead.

src/datepicker.jsx
lucygilbert: I hadn't considered that, I'll work the setState back into the code.

test/datepicker_test.js
lucygilbert: I'll fix this along with the handleFocus setState.

lucygilbert:
All the comments have now been addressed.

src/datepicker.jsx
rafeememon: Ah okay, that makes sense.  I think `forceUpdate` may be a little more semantic in this case, but I'd defer to @martijnrusschen for the final call here.

rafeememon:
Hey @lucygilbert, thanks for the quick updates here!  I'm having a problem with your latest change set where the picker quickly appears and disappears when I click to focus (using Chrome).  Could you take a look at that?

lucygilbert:
I'll have a look into it. I can't seem to recreate it (Chrome/OSX), but it has appeared before, with early versions of my code. I think it's a problem surrounding setting of focus in the onInputClick.

lucygilbert:
After testing in a few different browser/configurations, I think the problem is just that 100 milliseconds was an over-optimistic wait time for the `setTimeout`.

rafeememon:
Awesome, it's working on my end now.  Thanks!

Re: forceUpdate; I'm having a little trouble seeing why the calls are necessary -- I naively tried commenting them out and I can't seem to break the picker.  Can you provide a test case or a workflow that breaks if the calls are removed?

lucygilbert:
Hmm, I'll take a look, it might be the case that it was simply previous versions of the code were too brittle and so needed it. It might work without it now.

lucygilbert:
Okay. The forceUpdate in handleBlur can be removed. The one in onInputClick is necessary.

I'm still working out exactly why this happens but, if you open the date picker then make at least 2 clicks inside it (not selecting a date, just things like changing the month or year), then the calendar will reopen itself if you click in the date input field to close it.

lucygilbert:
I've worked it out. Without the forceUpdate in onInputClick, React allows the date input field to lose focus after the second click inside the date picker (because after the first click it updates because the state properties are different, but after the second click the state properties being set are always the same, so they never trigger an update). 
So when you click in the input field to close the date picker, it also triggers handleFocus, reopening the date picker.

rafeememon:
That makes sense.  Thanks for tracking that down, and thanks for the work you put in here!  LGTM :+1:

martijnrusschen:
Interesting and thanks for your contribution! I'll check this tomorrow.

martijnrusschen:
Thanks!

PR #274, opened 2015-12-19
Remove Object.assign for node v0.12 compatibility

rafeememon:
Most of my dev'ing happens on a box running node v0.12.  This is the only line that needs to be changed to make running the docs page compatible.


martijnrusschen:
LGTM

PR #282, opened 2015-12-23
Refactor and improve test coverage of Day component

rafeememon:
The bulk of this refactor involved moving day-specific logic from the `Calendar` component to the `Day` component.  This makes it a breeze to test how a day will be rendered under specific conditions.


src/util/date.js
```diff
@@ -18,13 +18,6 @@ DateUtil.prototype.sameMonth = function(other) {
   return this._date.isSame(other._date, "month");
 };
 
-DateUtil.prototype.inRange = function(startDate, endDate) {
```
rafeememon: Ultimately I think `DateUtil` objects should be phased out in favor of just passing around `moment` objects.  I've often confused myself when trying to figure out if one or the other is required in parts of the code.  For complex operations I think they should be pulled out into static utility methods that act on `moment` objects.

martijnrusschen: Yes, as far as I know, DateUtil was introduced as a way to avoid using moment objects. But since we're still heavily using moment, I would prefer using moment across the app instead of heaving this in-between state we're in now. So yeah, I guess I agree with refactoring this, and getting rid of the DateUtil object.

martijnrusschen:
Nice, @RSO can you take a look?

rafeememon:
@martijnrusschen, @RSO, I've also finished a refactor of the calendar display components that removes all uses of `DateUtil`.  It builds on this PR but I'll wait to push that for the sake of keeping the diff small.

Happy holidays!

src/util/date.js
RSO: Yes, `DateUtil` should go. Looking forward to your follow-up PR!

RSO:
@rafeememon thanks for the PRs! Sorry for taking so long to respond. I will test this diff and merge it. I also heard that you plan on shooting in some issues this weekend, I will reserve a day next week to work on the datepicker in the hope that we can work to an v1.0.0 release somewhere at the end of this month.

PR #284, opened 2015-12-28
Add on clear callback

aspirisen:
Sometimes it is useful to have different callbacks for on change and on clear events. For example in on change handler we can't distinguish incorrect data input from clear button click because in both cases the handler will receive null. Of course there are maybe another use cases.


RSO:
Thanks! Change looks good. If you fix the lint errors this should be good to go!

aspirisen:
Ok, when i will fix it should  i create a new pull request, or what? Because I haven't had experience in it.

RSO:
You can update the files in your branch and git-push the changes

aspirisen:
Done. Thanks

RSO:
Thank you very much for taking the time to contribute!

Issue #250, opened 2015-12-09
Be able to change the date by typing


xabikos:
Once a selection has be done it can only change by clicking the component and select another date. A nice addition which is related to #249 is to allow changing the selected from the keyboard.

martijnrusschen:
Fixed by #285

PR #285, opened 2016-01-03
Make datepicker to be editable by user input

aspirisen:
This improvement is connected with #250 issue. I haven't found better solution to make this enhancment and this one has one disadvantage that it is impossible to reset date to null when the date input has focus. 

P.S. I named the datepicker prop as 'isTypeable' because 'isEditable' may confuse user like it is imposible to change date.


martijnrusschen:
Nice work, and thanks for adding test + an example!

PR #292, opened 2016-01-08
WIP: Clear timeouts on unmount

### LINKED ISSUE #271: Blur timeout bug
Hi!

if parent component get destroyed, after datepicker lost focus, there is bug with timeouts (Warning: setState(...): Can only update a mounted or mounting component).

You need to clear timeouts in datepicker.jsx

RSO:
This needs tests.

Fixes #271

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -60,19 +60,26 @@ var DatePicker = React.createClass({
     return !(isEqual(nextProps, this.props) && isEqual(nextState, this.state));
   },
 
+  componentWillUnmount() {
+    window.clearTimeout(this._focusTimeout);
+    window.clearTimeout(this._blurTimeout);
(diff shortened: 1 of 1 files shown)


rafeememon:
Took a crack at writing tests, but it's going to be annoying... I started with the approach recommended on [stack overflow](http://stackoverflow.com/questions/23973942/unmount-destroy-component-in-jsdom-test), but it turns out an error isn't thrown when `setState` is called in an unmounted component -- only a warning message is printed =(

gaastonsr:
I think this is not the right way to do it. The way they recommend here https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html is a much better way in my opinion.

I'll do a PR.

rafeememon:
Not required after #319

Issue #290, opened 2016-01-08
Remove DateUtil abstraction


RSO:
When we initially introduced the `DateUtil` we only used `moment` for a couple of minor tasks (looping over all days in the week for example). At that point I decided that the abstraction would help us get rid of `moment` as a dependency in the future.

Looking at the codebase now it is save to say that we have grown more dependent on moment than was initially expected, and therefor it is also time to remove the `DateUtil` abstraction.

Ref #282 (sparked the initial discussion).

RSO:
(@rafeememon I'm assigning you since you mentioned you had a branch ready for this. Let me know if you can't work on this and I will happily take over)

rafeememon:
@RSO, please take a look when you get a chance, thanks!

PR #293, opened 2016-01-08
Calendar refactor, remove DateUtil

rafeememon:
This PR refactors the components comprising the calendar into things that are more semantic and testable, and also removes the `DateUtil` abstraction. (#290)


RSO:
Awesome stuff! This looks pretty much ready to go. Few concerns: 
- aren't we losing test coverage?
- it seems easy to miss a `.clone` call on moment. Something like moment/moment#1754 would be awesome, but I'm not sure how we could enforce this now. Thoughts?

rafeememon:
Hey @RSO, thanks for taking a look.

* I suppose we do lose test coverage, but it is mostly just from removing the moment wrapper methods and their tests.  Now that this uses moment directly we don't need to re-test moment.  More tests can (and should) be added to test the things that used these.
* Agreed that it's easy to miss a `.clone` call, and unfortunately I don't have a good answer other than being hypervigilant in PRs =(  Maybe @martijnrusschen has thoughts here

martijnrusschen:
> Agreed that it's easy to miss a .clone call, and unfortunately I don't have a good answer other than being hypervigilant in PRs =( Maybe @martijnrusschen has thoughts here

Yep agree. Not sure if it's possible to write a linter for this, but that could help in this case. Reading up on the discussion in moment/moment#1754, it looks like https://github.com/WhoopInc/frozen-moment could help here. Did you consider this? (or am I missing the point here? :) )

RSO:
No, I wouldn't go for a moment fork, that would defy the purpose of having moment (and we would need to re-introduce DateUtil ;-)). Okay, @rafeememon this PR looks good to me. I'll leave the merging to you. Agree that we need to be looking out for possible `.clone` mistakes.

martijnrusschen:
Okay I'm convinced :)

rafeememon:
A linting rule is an interesting idea.  I started putting a rule together here: https://github.com/rafeememon/jscs-moment-immutability  I'll put the finishing touches on it and integrate it here tomorrow.

rafeememon:
Got the rule working.  I'll make a separate PR to add it in.

PR #300, opened 2016-01-11
WIP: Avoid setState if component is unmounted

gaastonsr:
Right now I'm on Windows and I'm not able to make run the tests.

But when I get a chance I'll run this on Mac to test it.

What do you think?

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -56,6 +56,14 @@ var DatePicker = React.createClass({
     });
   },
 
+  componentDidMount() {
+    this._isMounted = true;
+  },
+
+  componentWillUnmount() {
+    this._isMounted = false;
+  },
+
   shouldComponentUpdate(nextProps, nextState) {
     return !(isEqual(nextProps, this.props) && isEqual(nextState, this.state));
   },
@@ -82,7 +90,7 @@ var DatePicker = React.createClass({
 
   hideCalendar() {
     setTimeout(() => {
-      this.setState({
+      this._isMounted && this.setState({
         focus: false
       });
     }, 0);
--- a/test/datepicker_test.js
+++ b/test/datepicker_test.js
@@ -74,4 +74,18 @@ describe("DatePicker", () => {
       done();
     }, 300);
   });
+
+  it("should not modify state if component is unmounted", done => {
+    var datePicker = <DatePicker />;
+    var container = document.createElement('div');
+    ReactDOM.render(datePicker, container);
+
+    // check is mounted
+    console.log('this._isMounted', datePicker._isMounted);
+
+    ReactDOM.unmountComponentAtNode(container);
+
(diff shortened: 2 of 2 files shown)


RSO:
From https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html:

> An easy migration strategy for anyone upgrading their code to avoid isMounted() is to track the mounted status yourself. Just set a _isMounted property to true in componentDidMount and set it to false in componentWillUnmount, and use this variable to check your component's status.
>
> An optimal solution would be to find places where setState() might be called after a component has unmounted, and fix them. Such situations most commonly occur due to callbacks, when a component is waiting for some data and gets unmounted before the data arrives. Ideally, any callbacks should be canceled in componentWillUnmount, prior to unmounting.

So basically your approach is what is suggested in the first paragraph and mine (#292) is the one suggested in the second paragraph, right? I would really prefer just not to have callbacks dangling when the component isn't mounted anymore. The code in your PR works for now, but might be error prone when more then one action is being done inside the callback.

gaastonsr:
If we analyze where the problem relies I think the only problematic point is when we lose focus from the input because it is unmounted and it tries to hide the calendar by calling to setState.

Scenario:
```
1. Input has focus (state.focus = true)
2. We unmount the element
3. Input loses focus because it was unmounted (state.focus still true), sets timeout
4. Timeout callback runs and tries to hide the calendar (state.focus = false)
5. Warning: you can't call setState from an element that is unmounted
```

handleFocus can't be problematic because you can't focus an element that is unmounted, handleBur and handleSelect are a proxy to hideCalendar so if we fix hideCalendar it would fix handleBlur.

rafeememon:
IMO, this is still an anti-pattern because this effectively implements `isMounted` and only avoids using React's implementation.  Agreed with @RSO that we should cancel the callbacks to fix the places where `setState` might be called, as the blog post suggests.

RSO:
> If we analyze where the problem relies I think the only problematic point is when we lose focus from the input because it is unmounted and it tries to hide the calendar by calling to setState.
> 
> Scenario:
> ```
1. Input has focus (state.focus = true)
2. We unmount the element
3. Input loses focus because it was unmounted (state.focus still true), sets timeout
4. Timeout callback runs and tries to hide the calendar (state.focus = false)
5. Warning: you can't call setState from an element that is unmounted
```
> handleFocus can't be problematic because you can't focus an element that is unmounted, handleBur and handleSelect are a proxy to hideCalendar so if we fix hideCalendar it would fix handleBlur.

Interesting trail of thought. I agree that the other cases are pretty hypothetical. Would you like it more if I update my PR to only track/clear the blur timeout?

gaastonsr:
Sorry for the late reply, let me check the PR.

gaastonsr:
@RSO I think your way works too :)

PR #306, opened 2016-01-16
Add JSCS moment immutability rule

rafeememon:
This will hopefully catch cases where we accidentally mutate moment objects.

Re: #293

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -80,5 +80,8 @@
   "disallowSpaceAfterPrefixUnaryOperators": true,
   "disallowSpaceBeforePostfixUnaryOperators": true,
   "disallowSpaceBeforeBinaryOperators": [ ",", ":" ],
-  "disallowMultipleLineBreaks": true
+  "disallowMultipleLineBreaks": true,
+
+  "additionalRules": ["moment-immutability"],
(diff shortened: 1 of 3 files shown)


rafeememon:
```
Running "jscs:files" (jscs) task
momentImmutability: Clone required for moment.js operations at src/day.jsx :
    46 |    if (!startDate || !endDate) return false;
    47 |
    48 |    const before = startOfDay(startDate).subtract(1, "seconds");
-------------------------------------------------^
    49 |    const after = startOfDay(endDate).add(1, "seconds");
    50 |    return startOfDay(day).isBetween(before, after);
momentImmutability: Clone required for moment.js operations at src/day.jsx :
    47 |
    48 |    const before = startOfDay(startDate).subtract(1, "seconds");
    49 |    const after = startOfDay(endDate).add(1, "seconds");
----------------------------------------------^
    50 |    return startOfDay(day).isBetween(before, after);
    51 |  },
>> 2 code style errors found!
```

It worked!  It turns out this case is actually okay because `startOfDay` returns a cloned object, but it's probably better to make it more explicit anyway.

martijnrusschen:
Nice work!

PR #310, opened 2016-01-19
Cleanups and refactors

rafeememon:
This is a random collection of refactors.  Sorry that it's not more structured.  I'll leave comments inline in the diff.


docs/src/examples/clear_input.jsx
```diff
@@ -24,6 +31,7 @@ export default React.createClass({
       <div className="column">
         <DatePicker
           selected={this.state.startDate}
+          onChange={this.handleChange}
```
rafeememon: This was depending on the internal state for the date display, not the selected prop.  If just depending on the state and not the prop, there's no way to use the date picker in a useful manner.

docs/src/examples/custom_class_name.jsx
```diff
@@ -24,14 +24,14 @@ export default React.createClass({
...
-              {"className='datepicker__input red-border' />"}
+              {"className='red-border' />"}
         </code>
       </pre>
       <div className="column">
         <DatePicker
           selected={this.state.startDate}
           onChange={this.handleChange}
-          className="datepicker__input red-border" />
+          className="red-border" />
```
rafeememon: It seems more sane to not require the user to pass in `datepicker__input` when giving a custom class name.

src/datepicker.jsx
```diff
@@ -45,25 +46,14 @@ var DatePicker = React.createClass({
...
-    this.setState({
-      selected: nextProps.selected
-    });
-  },
-
   shouldComponentUpdate(nextProps, nextState) {
     return !(isEqual(nextProps, this.props) && isEqual(nextState, this.state));
   },
 
-  getValue() {
```
rafeememon: This method confused me.  I guess this is how the consumer is supposed to get at the date state if not using the prop, but it's a very anti-React way of thinking IMO.

docs/src/examples/clear_input.jsx
martijnrusschen: Nice one!

src/date_input.jsx
```diff
@@ -72,15 +66,9 @@ var DateInput = React.createClass({
   },
 
   handleKeyDown(event) {
-    switch (event.key) {
-    case "Enter":
-      event.preventDefault();
-      this.props.handleEnter();
```
martijnrusschen: This isn't used anymore?

rafeememon: Pressing enter and escape both ended up eventually calling `hideCalendar`, so this simplifies it a bit

martijnrusschen: Ah!

martijnrusschen:
Looks like one of the tests is still failing.

src/datepicker.jsx
```diff
@@ -133,16 +116,8 @@ var DatePicker = React.createClass({
...
-    // Due to issues with IE onchange events sometimes this gets noisy, so skip if we've already cleared
-    if (this.state.selected === null) return;
-
-    this.setState({
-      selected: null
-    }, () => {
-      this.props.onClear();
-      this.props.onChange(null);
-    });
+    this.props.onClear();
```
rafeememon: `onClear` seems like API bloat in that it's just `onChange` with `null`  What do you think?

rafeememon:
Aha, forgot I had a fix for the other test in my git stash =P

src/datepicker.jsx
```diff
@@ -133,16 +116,8 @@ var DatePicker = React.createClass({
 
   onClearClick(event) {
     event.preventDefault();
-
-    // Due to issues with IE onchange events sometimes this gets noisy, so skip if we've already cleared
```
rafeememon: I think this isn't relevant anymore since this is now just a click event handler

rafeememon:
@martijnrusschen, the codecov CI results always seem to get stuck in pending, do you know what might be going on?

martijnrusschen:
I pushed a change for the codecov hook in travis.yml. Not sure if it fixes the issue, but this looks to be the recommended way to do this.

src/datepicker.jsx
rafeememon: Relevant discussion on #284.  Looks like the issue is that `onChange` is called when an invalid date is typed into the text input.  I think the right solution is to only call `setSelected` when the date is valid, rather than clearing out the selected date.  That way `onChange` will function as @aspirisen expected.

rafeememon: @RSO do you have any thoughts here?

RSO: Yeah, I remember having some doubts when landing that PR, but didn't put too much thought in it. Are you suggesting to get rid of `onClear` and rely on `onChange` to detect clearing the input?

The only thing I'm wondering is if people actually would want to distinguish the clear action from just manually clearing the input. I can't think of a reason why people would want this though.

src/datepicker.jsx
RSO: Agreed.

src/datepicker.jsx
rafeememon: Yeah, I think `onClear` should be handled with `onChange(null)`.  Personally I don't think consumers will need to know the difference between the two clearing actions you mentioned.

RSO: SGTM.

rafeememon: Cool -- I'll make fix to `setSelected` that I mentioned in another PR

martijnrusschen:
Okay, looks like I made a mistake in the codecov config. I fixed that, so it should be able to post its status to GH now. This diff looks good to go.

rafeememon:
Hooray, codecov worked now!

PR #312, opened 2016-01-20
Clear timeouts on unmount, do-over

### LINKED ISSUE #271: Blur timeout bug

rafeememon:
Fixes #271, replaces #292

Inspired by the following to get the tests to be meaningful:

* https://gist.github.com/jsdf/6fc35890e4ed4a219072
* http://stackoverflow.com/questions/23973942/unmount-destroy-component-in-jsdom-test

Not super happy with how this turned out, but let me know what you think.

### DIFF
--- a/test/datepicker_test.js
+++ b/test/datepicker_test.js
@@ -72,4 +72,35 @@ describe("DatePicker", () => {
       done();
     }, 300);
   });
+
+  describe("unmounting", () => {
+    var container;
+
+    beforeEach(() => {
+      container = document.createElement("div");
+    });
+
+    afterEach(() => {
+      container.remove();
+    });
+
+    it("should unmount successfully after a focus", done => {
+      var datePicker = ReactDOM.render(<DatePicker />, container);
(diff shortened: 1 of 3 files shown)


test/datepicker_test.js
```diff
@@ -74,4 +74,35 @@ describe("DatePicker", () => {
...
+
+    beforeEach(() => {
+      container = document.createElement('div');
+    });
+
+    afterEach(() => {
+      container.remove();
+    });
+
+    it("should unmount successfully after a focus", done => {
```
rafeememon: These are super fragile.  When they work everything is fine, but when they fail it somehow cascades into other test failures completely related to it.  I suspect it has something to do with how the component is mounted/unmounted outside of the system of React TestUtils.

rafeememon:
Not required anymore after #319

PR #313, opened 2016-01-21
Refactor date input selection by keyboard

rafeememon:
Relates to #250, #284, #285; discussion started on #310 

This started as a refactor of how date changes propagate out of the date input, namely how the date was cleared when an invalid input was entered.  The goal was to make `onClear` be the equivalent of `onChange(null)`.  As the dominoes fell, it turned out it no longer made sense for `isTypeable` to exist either.  From what I can tell this was introduced to let the user continue to type after their text would be wiped out from the date clearing on an invalid input... yeah, confusing.  I think this succeeds at simplifying both the implementation and the API, and I hope my test cases illustrate my points here.

The new date input behavior on change is:

* Valid date entered -> `setSelected(moment)`
* Empty string entered -> `setSelected(null)`
* Invalid date entered -> **do nothing**


src/date_input.jsx
```diff
@@ -55,7 +45,11 @@ var DateInput = React.createClass({
   handleChange(event) {
     var value = event.target.value;
     var date = moment(value, this.props.dateFormat, true);
-    this.props.setSelected(date.isValid() ? date : null);
+    if (date.isValid()) {
```
rafeememon: This is the meat and potatoes of this change.

rafeememon:
@aspirisen, heads up that these changes are going in

PR #316, opened 2016-01-21
Today button

Follio:
Add today button with some translations and the example.

When moment implements https://github.com/moment/moment/issues/2415, the translations can be taken from there.

### DIFF
--- a/docs/src/hero_example.jsx
+++ b/docs/src/hero_example.jsx
@@ -21,6 +21,7 @@ export default React.createClass({
     return <DatePicker
         selected={this.state.startDate}
         onChange={this.handleChange}
+        showTodayButton
         isTypeable={true} />;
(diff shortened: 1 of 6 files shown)


docs/src/hero_example.jsx
```diff
@@ -21,6 +21,7 @@ export default React.createClass({
     return <DatePicker
         selected={this.state.startDate}
         onChange={this.handleChange}
+        showTodayButton
```
martijnrusschen: I'd like to make this more explicit by saying `showTodayButton={true}`

martijnrusschen:
Nice work. Can you make sure the CI passes, and would love to see some tests added if you have time for that.

martijnrusschen:
Finishing in #326

Issue #309, opened 2016-01-19
Datepicker flashes between open and closed when input is clicked in IE


rcoy-v:
Version 0.18 introduced a bug in Internet Explorer. When the datepicker is opened and the input is clicked, the calendar quickly flashes between opened and closed. Rolling back to version 0.17, this issue does not exist.

rafeememon:
Possibly related to #273

brettjanssen:
It has the same problem in Chrome using a Windows Precision Touchpad (and in fact when you select the date it doesn't populate the field).  It works fine with a mouse though!  Weird.

raptoria:
+1   I have the same problem with IE10, my error message is:
Line: 110121
Error: Object doesn't support property or method 'setState'

xabikos:
I have the same behavior with IE11. Currently I am using version 0.17 but this has some performance issues as on onBlur is called multiple times something that makes the component slow.

rafeememon:
Should be fixed by #319

Issue #302, opened 2016-01-13
The second datepicker doesn't close


boyarskiy:
When open the first datepicker, change the month, then open the second one, and the first doesn't close.

Screenshot:
http://dl3.joxi.net/drive/0004/1007/275439/160113/5826535989.jpg

rafeememon:
Thanks for the report.  Looks like this applies to any two pickers on the example page.

rafeememon:
This is a limitation of `react-onclickoutside`, namely the `ignore-react-onclickoutside` class.  When clicking on another picker, the click doesn't trigger the first picker to close because of the class.

rafeememon:
Should be fixed by #319

Issue #283, opened 2015-12-23
open calander via external component?


aaronmgdr:
Id like to have a little calendar icon next to the select box. When it is clicked id like the calender selection to open. If this was a custom component id just included the icon in it and add an onclick, or take an optional prop / children. 

Is there a good way to do this with react-datepicker non hackily?

martijnrusschen:
Nice suggestion, I don't think we have any non-hacky ways to do this at this moment. How would you like to implement this. Any suggestions?

rafeememon:
I've been thinking about how this could work by having open/closed state managed by a prop rather than state.  Then the consumer has complete control over when it is open.  I'll experiment around with this.

fremn:
two ways I think could work are using optional children prop
```
<DatePicker  selected={this.props.startDate} onChange={this.onChangeStart} >
<i className="cal-icon"/>
</DatePicker>
```

or using an optional Icon prop

```
<DatePicker  selected={this.props.startDate} onChange={this.onChangeStart} icon={<i className="cal-icon"/>} >
```

internally something like 
```
icon() {
   if (this.props.icon) {
      <span onClick={this.onInputClick}> {this.props.icon} </span>
   }
}
render() {
   return (
      <div className="datepicker__input-container">
        <DateInput
          ref="input"
          id={this.props.id}
          name={this.props.name}
          date={this.state.selected}
          dateFormat={this.props.dateFormat}
          focus={this.state.focus}
          onFocus={this.handleFocus}
          onBlur={this.handleBlur}
          handleClick={this.onInputClick}
          handleEnter={this.hideCalendar}
          setSelected={this.setSelected}
          invalidateSelected={this.invalidateSelected}
          placeholderText={this.props.placeholderText}
          disabled={this.props.disabled}
          className={this.props.className}
          title={this.props.title}
          readOnly={this.props.readOnly}
          required={this.props.required}
          tabIndex={this.props.tabIndex} />
        {clearButton}
        {this.props.disabled ? null : this.calendar()}
        {this.icon()} 
      </div>
    );
}
```

martijnrusschen:
Yeah, opening/close via prop sounds a good idea.

fremn:
yep props sounds great to me.

gaastonsr:
Why not add an open method to the component?

Also, is somebody working on this? I can make a PR.

martijnrusschen:
Would be great if you could kick this off!

RSO:
[...]

Issue #161, opened 2015-08-13
Calendar does not hide when pressing Enter in IE


sconno05:
Steps to reproduce:

1 Go to example page: https://hacker0x01.github.io/react-datepicker/
2. In the first example (Default), enter a new value into the input field and press enter
3. The calendar does not close, but it will close in Chrome/Firefox

Tested in IE9, IE10, IE11 - all have the same behavior

tomhirschfeld:
+1 here, having the issue in IE.

benmliang:
+1, on IE 11, The calendar doesn't close, flash back and forth.

gweinert:
+1 here. On IE 11, 10, 9. I also see the calendar flash back and forth.

fabricior:
+1

rcoy-v:
I added a new [issue](#309) for the recent flashing issue in IE, as this thread is for a separate bug.

martijnrusschen:
Should be fixed with #319.

rafeememon:
Yup, just verified

PR #319, opened 2016-01-25
Refactor calendar open state

rafeememon:
Re: #273; addresses #106, #161, #302, #309; supersedes #315; sets up #283 

The `setTimeout`s around showing and hiding the calendar were becoming a bit unwieldy and difficult to reason about/test.  Instead of using `focus` state, I instead started thinking in terms of `open` state, and refactoring with this mindset cleaned up the logic magnificently.  Many bug fixes easily fell into place.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -24,7 +24,7 @@ var Calendar = React.createClass({
     moment: React.PropTypes.func.isRequired,
     dateFormat: React.PropTypes.string.isRequired,
     onSelect: React.PropTypes.func.isRequired,
-    handleClick: React.PropTypes.func.isRequired,
+    onClickOutside: React.PropTypes.func.isRequired,
     minDate: React.PropTypes.object,
     maxDate: React.PropTypes.object,
     startDate: React.PropTypes.object,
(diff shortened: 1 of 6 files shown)


src/calendar.jsx
```diff
@@ -37,7 +37,7 @@ var Calendar = React.createClass({
   },
 
   handleClickOutside(event) {
-    this.props.handleClick(event);
+    this.props.onClickOutside(event);
```
martijnrusschen: Not 100% related to this diff, but does it make sense to wrap a function in this function?

martijnrusschen: Not sure where this is being used but feels cleaner to just call the root function instead of wrapping the function in a wrapping local function.

rafeememon: For this case in particular the `onclickoutside` API requires that we have a `handleClickOutside` method.

For other cases, I seem to recall that React gives a warning message if you try to use a prop directly as a callback passed into a component.  Not sure if I'm making that up.

martijnrusschen: Fair enough.

PR #326, opened 2016-02-01
Today button

martijnrusschen:
Add today button and the example.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -24,7 +24,7 @@ var Calendar = React.createClass({
     moment: React.PropTypes.func.isRequired,
     dateFormat: React.PropTypes.string.isRequired,
     onSelect: React.PropTypes.func.isRequired,
-    handleClick: React.PropTypes.func.isRequired,
+    onClickOutside: React.PropTypes.func.isRequired,
     minDate: React.PropTypes.object,
     maxDate: React.PropTypes.object,
     startDate: React.PropTypes.object,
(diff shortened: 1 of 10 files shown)


src/calendar.jsx
```diff
@@ -2,6 +2,15 @@ import YearDropdown from "./year_dropdown";
 import Month from "./month";
 import React from "react";
 
+var todayTr = {
```
rafeememon: This is kind of yucky.  Maybe we can allow the today button text to be passed in as a prop, or can we get it out of localeData?

src/datepicker.jsx
```diff
@@ -36,7 +37,9 @@ var DatePicker = React.createClass({
       onChange() {},
       disabled: false,
       onFocus() {},
-      onBlur() {}
+      onBlur() {},
+      onClear() {},
```
rafeememon: Looks like onClear snuck back in here

src/datepicker.jsx
```diff
@@ -36,7 +37,9 @@ var DatePicker = React.createClass({
...
+      showTodayButton: false
```
rafeememon: Probably don't need this default, falsy would be implied by not passing in the prop

martijnrusschen: Correct.

src/datepicker.jsx
martijnrusschen: Fixed

src/calendar.jsx
martijnrusschen: Moved this to a prop.

martijnrusschen:
Added some tests for the today button. Still need to add tests to test the actual onClick

martijnrusschen:
Looks like I screwed up a merge conflict..

martijnrusschen:
@rafeememon Can you help out here? Not sure what the should be removed.

PR #327, opened 2016-02-01
Revert invalid date input when blurring

### LINKED ISSUE #106: Invalid changes to the input box are not reverted when the calendar is hidden
If you try this:

1. Click on the text box, this should bring up the calendar
2. Clic again inside it, this will hide the calendar
3. Type some nonsense in the textbox
4. Click outside

This will leave the textbox with garbage inside it, while the state still contains the last valid date.

Correct behaviour would be either reverting the garbage, or disabling hiding the calendar altogether while the text box is focused (by adding className `ignore-react-onclickoutside` to the input element)

rafeememon:
Fixes #106

When blurring the date input, either sets the value to the formatted date, or clears it if no date is selected.  Has the effect of wiping out invalid dates.

### DIFF
--- a/src/date_input.jsx
+++ b/src/date_input.jsx
@@ -61,6 +61,15 @@ var DateInput = React.createClass({
     }
   },
 
+  handleBlur(event) {
+    this.setState({
+      maybeDate: this.safeDateFormat(this.props.date)
+    });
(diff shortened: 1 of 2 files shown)


martijnrusschen:
lgtm

PR #328, opened 2016-02-02
Use react-tether for popovers

rafeememon:
This deletes a large portion of untested code implementing the popover functionality and replaces it with [`react-tether`](https://github.com/souporserious/react-tether), a small wrapper around `tether`.


martijnrusschen:
Sucks that the package has no tests too ;) But yeah, code looks cleaner!

rafeememon:
Yeah =( It might be good if we can contribute back!

Issue #303, opened 2016-01-13
Installing via Bower doesn't work.


lucygilbert:
After a bit of testing, it doesn't look like it has worked since 0.15.2. In 0.16.0, the dist folder was removed from the package, but the Gruntfile isn't running during the `bower install` so a dist folder is never being made.

RSO:
Thanks for the report. Hopefully I can check this out later today!

martijnrusschen:
Not sure on how to fix this, we fixed this issue in for NPM by adding an .npmignore that overwrites the .gitignore file, but doesn't look like something like that is available for Bower.

RSO:
Hmm, I'm also unsure as to how to fix this. I really don't want to add back those dist files (they really pollute PRs). We could maybe add something to the release script that adds them, create a tag, and then removes them again? Not sure how that would work.

I would love to know how many people still use Bower, and would like us to support it.

rafeememon:
It looks like React libraries generally have poor support for bower -- in fact Facebook doesn't publish a bower library for React (there's a shim repository, [yuck](https://github.com/reactjs/react-bower)), and popular React libraries don't publish to bower (redux, react-router).  I suspect not many people are using the datepicker bower library since we just discovered this now.

RSO:
Yeah, I think that the `react-datepicker-bower` repository might be the way to go. @martijnrusschen what do you think, would we be able to manage that as well, or should we leave that to someone else in the community?

martijnrusschen:
@rso do you think it would be doable to create a script that includes this in the release process? Way to go here looks like:
- Fork current repo into new bower specific repo
- Add dist folder back in the bower repo
- On a release of the main repo, also release bower repo.

I think this would be doable if we can automate this form the main repo.

rafeememon:
Stumbled upon this today: the approach taken by [Elemental UI](https://github.com/elementalui/elemental) may work, specifically their approach to using a bower publishing script.

rafeememon:
We're removing support for bower for the time being.  Please re-open this if you'd like to see us bring it back!

PR #330, opened 2016-02-02
Remove bower support

rafeememon:
Addresses #303

The bower package has been broken for quite some time, and it looks like we won't be supporting it here.  Maybe we'll create a bower distribution repo in the future if it's worth it.


martijnrusschen:
Can you check if there are any bower references left? I guess the gitignore will have some references to bower, for example.

rafeememon:
Haha yeah, realized that as soon as I pushed.  Updating the docs

PR #331, opened 2016-02-02
Add simple bash script to automate release steps

martijnrusschen:
Didn't test this yet, but this should be simple enough to try it out the next time. This can be used as:
`./release.sh patch`
`./release.sh minor`
`./release.sh major`


rafeememon:
Might be better to have this in a `/scripts` directory

Issue #329, opened 2016-02-02
Re-rendering datepicker causes setState error 


raptoria:
I have a page that re-mounts the same component when a new date is selected (with a new key, so the old DOM is tossed out).  The problem I have is 
the hideCalendar function gets called after the old datepicker component has already been removed.

the error:
setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the DatePicker component.

the source:
setTimeout(function(){e.setState({focus:!1})},0)}

Are there any plans to check if the component is mounted first before calling setState inside this function? Or is there a better way to solve this problem?

Thanks!

rafeememon:
This was fixed with #319 (all async operations where this can happen have been removed), but I don't think it's been released yet.  @martijnrusschen should we cut a release?

martijnrusschen:
Yep sounds good. I'll try to get a release out today.

raptoria:
awesome. can't wait!

martijnrusschen:
Just ran a release and updating the component to 0.20.0 should solve the problem. Can you confirm this works?

raptoria:
getting this error upgrading to the latest:

clacey@clacey-VirtualBox:~/react-flask$ npm install react-datepicker --save
npm WARN package.json react-flask@0.0.0 No repository field.
npm WARN peerDependencies The peer dependency react@^0.14.7 included from react-addons-shallow-compare will no
npm WARN peerDependencies longer be automatically installed to fulfill the peerDependency 
npm WARN peerDependencies in npm 3+. Your application will need to depend on it explicitly.
react-datepicker@0.20.0 node_modules/react-datepicker
├── react-onclickoutside@4.5.0
├── classnames@2.2.3
└── react-tether@0.1.2 (tether@1.1.1, react@0.14.7, react-addons-shallow-compare@0.14.7)
clacey@clacey-Virtu

running the code:

ERROR in ./~/react-datepicker/dist/react-datepicker.min.js
Module not found: Error: Cannot resolve module 'tether' in /home/clacey/react-flask/node_modules/react-datepicker/dist
 @ ./~/react-datepicker/dist/react-datepicker.min.js 1:170-187
webpack: bundle is now VALID.

martijnrusschen:
[...]

PR #335, opened 2016-02-05
don't expose tether anymore

martijnrusschen:
Ref #329

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -53,14 +53,6 @@ module.exports = {
       }
     },
     {
-      "tether": {
-        root: "Tether",
-        commonjs2: "tether",
-        commonjs: "tether",
-        amd: "tether"
-      }
-    },
-    {
       "lodash": {
         root: "_",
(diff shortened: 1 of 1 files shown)


rafeememon:
@raptoria, is there a chance you could try this out before we merge it so we can make sure it solves your problem?  I built the branch and pasted the contents of react-datepicker.min.js here: http://hastebin.com/falidujulo.js  If you replace the contents of react-datepicker.min.js with that you should be able to try it out.  Let us know!

raptoria:
It solves the setState error, and works wonderfully in IE10 as well.

rafeememon:
Ah yeah, I think `datepicker__container` got changed to `datepicker__tether-element`.  Here's the new stylesheet: http://hastebin.com/cuheroyema.css  Let us know if that works.

raptoria:
are you guys going to publish the latest version to NPM? thanx!

martijnrusschen:
Just released v0.21

Issue #338, opened 2016-02-08
How to set Today


nsisodiya:
I want to show a different Today. Because my client can visit to any other country he want to see date as his actual today and not the today in the country he is present.
Any idea or hack how to do it ?

nsisodiya:
I checked the code.
Instead of this
```js
"datepicker__day--today": this.isSameDay(moment()),
```

I suggest for having 

```js
"datepicker__day--today": this.isSameDay(this.props.today),
```

and you can set default value of today is `moment()` and user will be able to overwrite it.

nsisodiya:
Generated Pull Request - https://github.com/Hacker0x01/react-datepicker/pull/339

rafeememon:
Is it possible to use moment localizations so that the dates are always in the timezone that you specify?  I'm weary of adding this to the API of the datepicker and breaking the simple contract of today being "today."

nsisodiya:
@rafeememon - instead of timezone, specifying today is much better approach just like `selected` prop. In my pull request, I have added this props as optional prop.

Also, there is no breaking because advance user want this behaviour. normal user will not pass `today` props. There is not a single date picker exist which provide this functionality. (Setting as custom today)

rafeememon:
Understood that this is not a breaking change, but in general I would prefer to not add to the API, as we need to support it.  Can you explain why this approach is better than using localization?

nsisodiya:
Because, If you want to pass timezone then you have to add moment-timezone library.
http://momentjs.com/timezone/ It is 25KB on gzipped.
I do not know, just for one simple task, why will we add another library specially this is a client side code.

`Can you explain why this approach is better than using localization`
I exactly do not understand this. `localization` has nothing to do with it. 
I made a pull request for API so that user can set a different today other than today as seen by the browser.

rafeememon:
I think part of the job of software engineering is finding the right set of features that:

* makes common use cases easy
* makes advanced use cases **possible**

I think a big advantage of using the moment.js library is that we are able to leverage its plugins, like moment-timezone, making it unnecessary to duplicate its effects.  As we move towards a 1.0 release of the datepicker we're focusing on cleaning up the API, and I'm not convinced that the `today` prop belongs for this use case because you can use moment-timezone.  Keep in mind that once a prop is added, we need to support it long-term.

For your use case I suggest that you do one of the following:

1. Use moment-timezone
[...]

PR #339, opened 2016-02-08
Taking Today as props. Help in some extreme case when you wanted to …

nsisodiya:
…show different today than your current timeZone.

### DIFF
--- a/src/day.jsx
+++ b/src/day.jsx
@@ -16,16 +16,16 @@ var Day = React.createClass({
     includeDates: React.PropTypes.array,
     filterDate: React.PropTypes.func,
     selected: React.PropTypes.object,
-	  today: React.PropTypes.object,
+    today: React.PropTypes.object,
     startDate: React.PropTypes.object,
     endDate: React.PropTypes.object
   },
 
-	getDefaultProps: function() {
-		return {
-			today: moment()
-		};
-	},
+  getDefaultProps: function() {
+    return {
+      today: moment()
+    };
+  },
 
   handleClick(event) {
     if (!this.isDisabled() && this.props.onClick) {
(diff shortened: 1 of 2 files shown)


src/day.jsx
```diff
@@ -16,10 +16,17 @@ var Day = React.createClass({
     includeDates: React.PropTypes.array,
     filterDate: React.PropTypes.func,
     selected: React.PropTypes.object,
+	  today: React.PropTypes.object,
```
rafeememon: How is this prop set?  Does it need to be passed down from the top level?

src/day.jsx
```diff
@@ -16,10 +16,17 @@ var Day = React.createClass({
...
     startDate: React.PropTypes.object,
     endDate: React.PropTypes.object
   },
 
+	getDefaultProps: function() {
```
rafeememon: Please fix up the indentation =)

rafeememon:
Please also add test cases for the new prop.  Thanks!

src/day.jsx
martijnrusschen: Strange, linter should catch this too.

src/day.jsx
nsisodiya: Its optional. you can pass or you need not to pass. that is why I have added `getDefaultProps`

src/day.jsx
```diff
@@ -16,10 +16,17 @@ var Day = React.createClass({
     includeDates: React.PropTypes.array,
     filterDate: React.PropTypes.func,
     selected: React.PropTypes.object,
+    today: React.PropTypes.object,
```
nsisodiya: This is Optional.

nsisodiya:
Ok, I am done with Test Cases !!

rafeememon:
Hold off on merging until discusson on #338 is resolved

src/day.jsx
rafeememon: Continuing from before -- it's optional, but how does the consumer use it?  It's not being passed down from the datepicker component.

PR #341, opened 2016-02-09
Upgrade moment

martijnrusschen:


rafeememon:
:+1: 

It would be cool to get rid of `lodash` as a hard dependency (probably still need it in devDependencies).  I think the only thing that would need to be figured out is one use of `isEqual`

martijnrusschen:
@rafeememon Yeah would be nice to get rid of it. Somehow lodash failed to upgrade so took that one out for now.

rafeememon:
@martijnrusschen yeah that was odd.  Taking a complete shot in the dark, but maybe it has to do with the `cache` settings in `.travis.yml`?  It's probably safe to remove that block regardless

martijnrusschen:
Had the same issues locally. Looks like lodash moved some things in this update, since we're only including some parts of the lodash lib this may impact us more than expected.

rafeememon:
Cool, I'll take a look at it when I get a chance

martijnrusschen:
Extracted to https://github.com/Hacker0x01/react-datepicker/pull/342

PR #346, opened 2016-02-10
Date helper refactor

### LINKED ISSUE #259: TextInput does not validate minDate
When `minDate` is set, a user can still manually enter a date in the past using the text input instead of the date chooser.

rafeememon:
Time to finally fix #259.  This extracts some common date logic into a utility file.  Most importantly, it extracts the check to see if a date is enabled, which can then be used easily by the date input.


src/date_utils.js
```diff
@@ -0,0 +1,18 @@
...
+  }
+}
+
+export function isDayDisabled(day, { minDate, maxDate, excludeDates, includeDates, filterDate } = {}) {
+  return (minDate && day.isBefore(minDate, "day")) ||
+    (maxDate && day.isAfter(maxDate, "day")) ||
+    (excludeDates && excludeDates.some(excludeDate => isSameDay(day, excludeDate))) ||
+    (includeDates && !includeDates.some(includeDate => isSameDay(day, includeDate))) ||
+    (filterDate && !filterDate(day.clone())) ||
+    false;
```
rafeememon: Sometimes I hate Javascript.  At least the tests caught this with ease.

PR #359, opened 2016-02-15
Upgrade react-tether to 0.3.3

rafeememon:
Fixes #352, fixes #354, fixes #357

PR #360, opened 2016-02-15
Lock version of react-tether

rafeememon:
I'm having issues on #359 and I'm not quite sure what's going on yet.  In the meantime we should lock down the version of react-tether and make a release.

@martijnrusschen, can you merge this and make a patch release?


rafeememon:
New version of react-tether released, probably don't need this

PR #366, opened 2016-02-16
Use localized date format for DateInput

rafeememon:
As part of our localization overhaul, I think we should use the default date format for the selected locale in `DateInput`.  The `L` format expands to the locale-specific format.  Examples:

* `en` : `MM/DD/YYYY`
* `en-gb` : `DD/MM/YYYY`
* `ko` : `YYYY.MM.DD`

### DIFF
--- a/src/date_input.jsx
+++ b/src/date_input.jsx
@@ -9,6 +9,7 @@ var DateInput = React.createClass({
   propTypes: {
     date: React.PropTypes.object,
     locale: React.PropTypes.string,
+    dateFormat: React.PropTypes.string,
     minDate: React.PropTypes.object,
     maxDate: React.PropTypes.object,
     excludeDates: React.PropTypes.array,
@@ -19,7 +20,7 @@ var DateInput = React.createClass({
 
   getDefaultProps() {
     return {
-      dateFormat: "YYYY-MM-DD"
+      dateFormat: "L"
     };
   },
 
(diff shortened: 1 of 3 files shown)


src/date_input.jsx
```diff
@@ -19,7 +20,7 @@ var DateInput = React.createClass({
 
   getDefaultProps() {
     return {
-      dateFormat: "YYYY-MM-DD"
+      dateFormat: "L"
```
martijnrusschen: Where does this format come from? Any documentation on what's available?

rafeememon: http://momentjs.com/docs/#/displaying/ under "Localized formats"

martijnrusschen: Should we link that from the readme or maybe the docs? (docs has an option to add summaries per proptype

rafeememon: Yeah, adding it to the proptype summary sounds like a good idea

martijnrusschen: See https://github.com/reactjs/react-docgen/blob/master/example/components/Component.js on how to add that

martijnrusschen: @rafeememon Are you adding a summary to this proptype?

rafeememon: Yep, working towards some deadlines on my real work though, haha

aouaki:
Hey guys,
Any news on this? I can help if you need, there doesn't to be much work left before this is merged...
Thanks

src/date_input.jsx
rafeememon: Looking at this again -- the default format actually belongs to `DateInput`, which is currently internal API.  Re: docs I think we should handle this as we overhaul docs but I don't think it should block this merge

rafeememon:
Hey @aouaki, merging this now

PR #369, opened 2016-02-23
doc: add className props

PBRT:
Add the 'className' params in the readMe related to props. 

Super small add but useful since I searched this props for a while ;)

Awesome comp!

### DIFF
--- a/docs/datepicker.md
+++ b/docs/datepicker.md
@@ -99,3 +99,6 @@ defaultValue: `[
 
 type: `string`
 
+### `className`
+
+type: `string`


martijnrusschen:
Hi @PBRT 

Thanks for contributing to our component. I'm using a automated script to generate those docs. I created a new PR that uses the scripts and renames the prop (I found the prop name really confusing). See #370

martijnrusschen:
The docs are based on the proptypes in the component. If there's no proptype defined, this change would be gone the next time I generate the docs.

PR #370, opened 2016-02-23
Rename className prop and add to docs

martijnrusschen:
Continuing work from #369. Added proptype and used generator to add the prop to the docs. Also renamed the prop to a more descriptive name. (**Breaking change**)

### DIFF
--- a/docs/datepicker.md
+++ b/docs/datepicker.md
@@ -28,6 +28,11 @@ type: `func`
 type: `string`
 
 
+### `inputClass`
+
+type: `string`
+
+
 ### `locale`
 
 type: `string`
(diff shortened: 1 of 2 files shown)


martijnrusschen:
There are more missing propTypes we should add. I'm working on a new set of linters that should flag these.

rafeememon:
Hey @martijnrusschen, what's the reasoning for changing the name of the prop?  I think `className` is the generally accepted prop name for this type of thing.

martijnrusschen:
yeah className would be preferable. But what does className refer too in this case? This could refer to multiple things here (eg. inputfield, calendar, etc.) Or am I thinking to complicated here?

rafeememon:
Fair point -- personally my first thought is that it refers to the DateInput field since it's the first thing visible and is the only thing that's consistently visible.  Currently it looks like we don't have a way of passing a class name to the calendar, but I could imagine that prop being called `calendarClassName` or similar.

rafeememon:
Hey @martijnrusschen, not sure this should have been merged -- I feel pretty strongly that `className` is the right prop name here, and we should use `calendarClassName` for the calendar class.  Let me know what you think.

Issue #353, opened 2016-02-12
Datepicker wrap in bootstrap group input has wrong offset


james4388:
I want to add input group with calendar icon for datepicker, this is what I've done:

```
<div className="input-group">
          <span className="input-group-addon" id="basic-addon1"><i className="glyphicon glyphicon-calendar"></i></span>
          <DatePicker className="form-control"
              selected={this.state.selectedDate}
              onChange={this.handleDateChange.bind(this)}
              {...field}
         />
</div>
```

It's worked, amazing

<img width="521" alt="screen shot 2016-02-12 at 1 46 19 pm" src="https://cloud.githubusercontent.com/assets/174132/13021262/167302e0-d18f-11e5-805f-9f0a3fc59950.png">

But when click on it, the calendar got wrong offset coordinate and z-index do not ritght.

<img width="269" alt="screen shot 2016-02-12 at 1 47 55 pm" src="https://cloud.githubusercontent.com/assets/174132/13021288/40e59e20-d18f-11e5-80b5-619678e27b4e.png">

What should I do to improve

rafeememon:
You can adjust the dropdown placement with the `popoverTargetOffset` prop, described here: https://github.com/Hacker0x01/react-datepicker/blob/master/docs/datepicker.md

james4388:
Oh nice, thanks

HadiAkmal:
@rafeememon 

There's no `popoverTargetOffset` prop in the link you provided. I'm having this problem:

![image](https://user-images.githubusercontent.com/14165645/28086445-9949068e-66b1-11e7-9d9c-b73d81a238c4.png)

Top calendar is having overlay problem with bottom date input field. How do I get the calendar to the front of the input field?

artemv:
For those who came here to fix the input group addon layout issues with DatePicker - there's a relevant https://github.com/Hacker0x01/react-datepicker/issues/484 issue, and solution based on https://github.com/Hacker0x01/react-datepicker/issues/484#issuecomment-264408923 worked for me to solve this issue as well.

Issue #343, opened 2016-02-09
Add container's parameter for <Popover/> component


arthur-creek:
There are some problems when I use `<DatePiker/>` with <Overlay rootClose /> from https://github.com/react-bootstrap/react-bootstrap
Input of `<DatePicker/>` is located inside `<Overlay/>`, but `<Popover/>` with `<Calendar/>` is outside.
And `<Overlay/>` hides on select date action.
It will be beautiful to have ability to change parent element for `<Popover/>`

`<DatePiker tetherConstraints={[...]}/>`

rafeememon:
Unfortunately it's not possible to change the parent element for the popover because of the tethering library we use.  The issue you're seeing is probably a z-index problem, so try increasing the z-index of `.datepicker__tether-element`.

arthur-creek:
No. `z-index` param resolves  the visibility problem only. It was fixed with CSS.

rafeememon:
What is the problem you are seeing?

vsai:
@rafeememon In https://github.com/souporserious/react-tether/commit/c806cb1fd21438038980e382f4b996ef6d07c8ac, react-tether added support to specify which parent element the tether-element (in this case the calendar) will be added to.

rafeememon:
@vsai, there were some issues with that prop the last time I tried to use it, but it may be fixed now (from a couple bugs I fixed over there =P)

This may be something we implement, but I'm still interested in hearing the exact problems aside from the z-index issue so that we can have the right reasoning for adding this to the API.

vsai:
@rafeememon The z-index doesn't seem to be working for me. I tried setting the z-index of the calendar div to 1, 100, 1000, etc., but wasn't able to get it to work.
The context: I have the DatePicker on a react-bootstrap modal, and I want the DatePicker element be within the modal. The input box works fine, but the actual Calendar is hidden below the div.

I would imagine adding the calendar div to the same parent as the input box would fix it. But that would involved using the tether-component change.

rafeememon:
@vsai, could you try adding the new lines from #378 to your stylesheet, and if you're still having the issue, would you mind putting together an example app showing the problem?  I'm a little perplexed that adding the z-index didn't work for you.

rafeememon:
Added the `renderCalendarTo` prop with #380

Issue #332, opened 2016-02-03
Datepicker 0.20 doesn't work in react-bootstrap's modal


vojty:
0.19 was fine
There is no error in console...

rafeememon:
Probably a z-index issue; try finding the `datepicker__tether-element` at the end of the DOM when you click the picker, and increasing its z-index until it's visible.

vojty:
yeah that's, thank you!

rafeememon:
Woot!

PR #378, opened 2016-03-01
Add z-index to tether-element

### LINKED ISSUE #375: Parameterize z-index for tether-element
Hi, I'm very new to React so sorry in advance if I missed something.

My first time I tried to use this component was inside a takeover which uses an increased z-index, and so the `.datepicker__tether-element` appeared behind the takeover markup. It would be great to be able to pass z-index as a parameter so that style overrides are not required.

Like I said I'm new to React, but would be happy to try to do the work and send a pull request if you can point me in the right direction :-)

martijnrusschen:
Fixes #375

### DIFF
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -306,5 +306,5 @@
 }
 
 .datepicker__tether-element {
-  z-index: 2;
+  z-index: 2147483647;
 }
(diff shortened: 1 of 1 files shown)


rafeememon:
:+1:

rafeememon:
Also relates to #332, #343, #353

PR #379, opened 2016-03-01
Add peerDependency of webpack-dev-middleware

martijnrusschen:
I guess this is a side effect of me upgrading to NPM 3. Did anyone else encounter the same issue? Didn’t had this issue with NPM 2.x


rafeememon:
Seems okay to me -- NPMv3 no longer installs peer dependencies automatically so they need to be added manually like this

PR #380, opened 2016-03-02
Parameterize the calendar's parent node

karleberts:
Addresses #343, my specific issue with appending the calendar to the body has to do with triggering a mouseout of the datepicker container.

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -40,6 +40,7 @@ var DatePicker = React.createClass({
       popoverAttachment: "top left",
       popoverTargetAttachment: "bottom left",
       popoverTargetOffset: "10px 0",
+	  renderCalendarTo: null,
       tetherConstraints: [
         {
           to: "window",
@@ -137,6 +138,7 @@ var DatePicker = React.createClass({
         attachment={this.props.popoverAttachment}
         targetAttachment={this.props.popoverTargetAttachment}
         targetOffset={this.props.popoverTargetOffset}
+		renderElementTo={this.props.renderCalendarTo}
         constraints={this.props.tetherConstraints}>
         <div className="datepicker__input-container">
           <DateInput
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -40,7 +40,6 @@ var DatePicker = React.createClass({
       popoverAttachment: "top left",
       popoverTargetAttachment: "bottom left",
       popoverTargetOffset: "10px 0",
-	  renderCalendarTo: null,
       tetherConstraints: [
         {
           to: "window",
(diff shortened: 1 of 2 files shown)


rafeememon:
Hey @karleberts, thanks for helping out here.  I'm wondering if there's another way to approach this without adding to the API.  You mentioned that you want to detect mouseout events on the `datepicker__input-container`; I think it might be reasonable to just make that the default element appended to.  What do you think?

cc @martijnrusschen

karleberts:
@rafeememon That would work for my case specifically, but I imagine there are times where it might be necessary to put the input in a container with `overflow: hidden`, which would prevent the calendar from displaying properly.

I think rendering to an external node is best, I'd just like to be able to decide which node ;)

rafeememon:
@karleberts, yeah that all makes sense, thanks.  In fact, not making the default as `datepicker__input-container` leaves open the possibility of getting rid of the container element altogether so that the `<input>` is the top level element.  I've been wanting to make that refactor for a while...

src/datepicker.jsx
```diff
@@ -40,6 +40,7 @@ var DatePicker = React.createClass({
       popoverAttachment: "top left",
       popoverTargetAttachment: "bottom left",
       popoverTargetOffset: "10px 0",
+	  renderCalendarTo: null,
```
rafeememon: nit: it's safe to get rid of this line, when it's not specified it's undefined which is good enough

src/datepicker.jsx
```diff
@@ -137,6 +138,7 @@ var DatePicker = React.createClass({
         attachment={this.props.popoverAttachment}
         targetAttachment={this.props.popoverTargetAttachment}
         targetOffset={this.props.popoverTargetOffset}
+		renderElementTo={this.props.renderCalendarTo}
```
rafeememon: nit: please fix indenting =)

karleberts:
Cool cool. Fixed the indent/prop

rafeememon:
:+1:

src/datepicker.jsx
```diff
@@ -137,6 +137,7 @@ var DatePicker = React.createClass({
         attachment={this.props.popoverAttachment}
         targetAttachment={this.props.popoverTargetAttachment}
         targetOffset={this.props.popoverTargetOffset}
+        renderElementTo={this.props.renderCalendarTo}
```
martijnrusschen: Do we need to add a proptype for this?

rafeememon: Yeah think so, want to take care of that?

martijnrusschen: Trying to finalize the ESlint conversion. I'll add it as part of cleaning up all lint errors.

PR #381, opened 2016-03-02
Add documentation for more properties

arrel:
Added dateFormat, className, placeholder properties
Added property descriptions
Reformatted to use JSDoc notation

### DIFF
--- a/docs/datepicker.md
+++ b/docs/datepicker.md
@@ -6,96 +6,77 @@ General datepicker component.
 Props
 -----
 
-### `dateFormatCalendar`
-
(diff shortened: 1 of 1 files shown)


rafeememon:
Hey @arrel, thanks for helping out here.  The API docs are generated with react-docgen so unfortunately these would be wiped out the next time they're generated.  Could you look into adding this info into the props in `datepicker.jsx`?

arrel:
Aha! that's why the formatting was so terrible. I'll open another one tomorrow.

PR #383, opened 2016-03-03
Set up ESLint with Standard and React config

martijnrusschen:
Did a first pass on setting up the rules. I'll enforce and fix the issues once we've settled on the rules. I used Standard config, combined with a more specific React config.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,65 @@
+{
+  "env": {
+    "browser": true,
+    "es6": true,
+  },
+
+  "plugins": [
+    "react"
+  ],
+
+  "extends": "standard",
+
+  "ecmaFeatures": {
+    "arrowFunctions": true,
+    "blockBindings": true,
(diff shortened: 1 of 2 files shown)


martijnrusschen:
@rafeememon Can you give your opinion about this?

.eslintrc
```diff
@@ -0,0 +1,65 @@
...
+    "objectLiteralDuplicateProperties": true,
+    "objectLiteralShorthandMethods": true,
+    "objectLiteralShorthandProperties": true,
+    "spread": true,
+    "superInFunctions": true,
+    "templateStrings": true,
+  },
+
+  "parser": null,
+  "rules": {
```
martijnrusschen: Tried to use the `"extends": ["eslint:recommended", "plugin:react/recommended"]` config but didn't get it working. This rules below are copied over from our own app and can be modified if we want something different.

PR #386, opened 2016-03-04
Make proptypes required 

martijnrusschen:


martijnrusschen:
@rafeememon Can you do a sanity check on the added propTypes. It's Friday evening, could have missed something :)

martijnrusschen:
Wondering if we should set the isRequired flag on some of the new propTypes. Thoughts?

docs-site/src/code_example_component.jsx
```diff
@@ -3,6 +3,12 @@ import React from 'react'
 export default React.createClass({
   displayName: 'CodeExampleComponent',
 
+  propTypes: {
+    id: React.PropTypes.string,
+    title: React.PropTypes.string,
+    children: React.PropTypes.element
```
rafeememon: In general I think dev usually go for `PropTypes.any` for the `children` prop, but since this is only code for the demo site it's not a huge deal

src/calendar.jsx
```diff
@@ -19,7 +19,9 @@ var Calendar = React.createClass({
     excludeDates: React.PropTypes.array,
     includeDates: React.PropTypes.array,
     filterDate: React.PropTypes.func,
-    showYearDropdown: React.PropTypes.bool
+    showYearDropdown: React.PropTypes.bool,
+    selected: React.PropTypes.object,
+    todayButton: React.PropTypes.bool
```
rafeememon: I think this should be `string` instead

src/datepicker.jsx
```diff
@@ -29,7 +29,20 @@ var DatePicker = React.createClass({
...
+    endDate: React.PropTypes.object,
+    excludeDates: React.PropTypes.array,
+    includeDates: React.PropTypes.array,
+    name: React.PropTypes.string,
+    isClearable: React.PropTypes.bool,
+    placeholderText: React.PropTypes.string,
+    title: React.PropTypes.string,
+    readOnly: React.PropTypes.bool,
+    required: React.PropTypes.bool,
+    renderCalendarTo: React.PropTypes.string
```
rafeememon: I don't think this should be `string`, but I'm not sure what it should be instead.  Probably `any` but it's worth looking at what can be passed into the tether API for this

rafeememon:
Re: making things required; I think we should try to keep as few things required as possible.  This makes it much easier to begin using the component and then use more advanced features when applicable, but not being forced to use them.

docs-site/src/code_example_component.jsx
martijnrusschen: Works at this moment. We can relax the settings at any time :)

martijnrusschen:
Fixed the comments. Will merge this when Travis is happy.

src/calendar.jsx
```diff
@@ -19,7 +19,9 @@ var Calendar = React.createClass({
     excludeDates: React.PropTypes.array,
     includeDates: React.PropTypes.array,
     filterDate: React.PropTypes.func,
-    showYearDropdown: React.PropTypes.bool
+    showYearDropdown: React.PropTypes.string,
```
rafeememon: Think this one should've stayed as a bool

Issue #377, opened 2016-03-01
Calendar styling


pinturic:
It would be nice to have the chance to add a className (or any other way to style) each calendar popup on its own: for example I have a daterange in my project and I want the triangle to be left aligned for the "fromDate" control and right aligned for the "toDate" one

ekrokowski:
I have the same case. 2 input fields horizontally in one row:

Left input / calendar for the startDate with 
```
popoverAttachment="top left"
popoverTargetAttachment="bottom left"
```

And a right input / calendar for the endDate with 
```
popoverAttachment="top right"
popoverTargetAttachment="bottom right"
``` 

The triangle is on both situations aligned on the left side. There is just missing CSS.

Simple fix would be to add the following CSS:
```
.react-datepicker__tether-element-attached-right .react-datepicker__triangle{
  left: auto;
  right: 50px;
}
```
`react-datepicker` only sets `left: 50px;` (even on right-aligned).

ian-moore:
@ekrokowski yes I have the same problem. Would be nice if there was a way to reposition the triangle

![image](https://cloud.githubusercontent.com/assets/1709416/22212347/fc09e0cc-e155-11e6-889a-5d3fe1f1832c.png)

12tp12:
hey, any update?

PR #387, opened 2016-03-06
New calendarClassName prop to override datepicker class

dmnd:
By default a class of `datepicker` is used in the markup. This is
annoying because I already have something else using that class in my
app.

This PR introduces a `calendarClassName` prop which allows the caller to
override the `datepicker` class.

Test plan:

`npm test`

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -21,10 +21,17 @@ var Calendar = React.createClass({
     filterDate: React.PropTypes.func,
     showYearDropdown: React.PropTypes.string,
     selected: React.PropTypes.object,
-    todayButton: React.PropTypes.string
+    todayButton: React.PropTypes.string,
+    className: React.PropTypes.string
   },
 
   mixins: [require('react-onclickoutside')],
+
+  getDefaultProps () {
+    return {
+      className: 'datepicker'
+    }
+  },
 
   getInitialState () {
     return {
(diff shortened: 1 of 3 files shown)


src/calendar.jsx
```diff
@@ -132,7 +139,7 @@ var Calendar = React.createClass({
 
   render () {
     return (
-      <div className="datepicker">
+      <div className={this.props.className}>
```
rafeememon: This seems problematic because the provided stylesheet assumes a `.datepicker` class, and it's not obvious how to make the existing styles work without copying/pasting them to a new class.  I'd prefer the `className` prop to be strictly additive to `.datepicker`, which is the way I've seen other react common components implemented.

rafeememon:
Re #370, #377

src/calendar.jsx
dmnd: The readme says "You will also need to require the css file from this package (or _provide your own_)", so I assumed this would be ok. I'm using my own stylesheet because of the clash with `.datepicker`.

If you'd prefer not to make this this class overridable, would you consider changing the class to something more specific?

rafeememon: Changing the class seems reasonable.  It looks like most of the styles are named as `datepicker__component` although we're admittedly inconsistent with this.  The class should probably be `datepicker__calendar` instead.

rafeememon: See #388

rafeememon:
Addressing in #388 instead

PR #388, opened 2016-03-07
Change calendar class to be more specific

rafeememon:
Re #387

Change the class name of the calendar component from `datepicker` to `datepicker__calendar` to help avoid potential naming conflicts.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -132,7 +132,7 @@ var Calendar = React.createClass({
 
   render () {
     return (
-      <div className="datepicker">
+      <div className="datepicker__calendar">
         <div className="datepicker__triangle"></div>
         <div className="datepicker__header">
           <a className="datepicker__navigation datepicker__navigation--previous"
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -1,7 +1,7 @@
 @import "variables";
 @import "mixins";
 
-.datepicker {
+.datepicker__calendar {
   font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
   font-size: 11px;
   background-color: #fff;


dmnd:
This works for me :+1:

martijnrusschen:
Not sure about this. If you look at the way BEM CSS works, you should have a root class that's used like we do now. Getting rid of that makes the CSS less BEM'ish. Can't we change the classes to something more specific for this component instead?

martijnrusschen:
Something like `.react-datepicker`?

rafeememon:
Isn't the block `datepicker` and element `calendar`, like most of the components are styled now?  I think this change at least makes it more consistent with that style.

rafeememon:
Thinking about this a little more, the change does remove the `.datepicker` top-level element, but technically the calendar doesn't nest in the element -- it's appended to `<body>` with the tethering library.  Also, I don't think we should limit ourselves to having a wrapper `div` with the `datepicker` class as this sets us up to have the top level element be the `input` instead.

martijnrusschen:
I think the problem is 2-sided:
1.  I think we should try to wrap all our classes with a specific prefix for our component. This avoids any potential override issues. For example, `datepicker` we have here, but also `close-icon` are easily overwrite-able where this shouldn't happen by default.
2. We should try to maintain a "block" class as the base class for this component. We can debate what the block is. The current `datepicker` class is probably the the best candidate, but I agree that it's not 100% correct. (for example, the input isn't an element of the datepicker block.)

rafeememon:
1. Definitely agree that all classes should be prefixed with `datepicker`.  The `close-icon` class is sadness that we should fix.  =P
2. I see what you mean, but take a look at my comment right before yours.  I think it would be great to have it possible for the `input` to be the top-level element to integrate with forms as seamlessly as possible.  Some styles in common libraries (bootstrap, elemental) look for inputs to be immediate children of certain elements, but forcing the datepicker to be wrapped in a div makes it impossible to work with this without nasty custom styling.

martijnrusschen:
What about:
- react-datepicker
  - react-datepicker__input
    - possible child classes
  - react-datepicker__calendar
    - children 

Trying to figure out the best structure on solving this and moving forward.. Still have the wrapper div this way, lets try to iterate and find the best solution.

martijnrusschen:
Or we introduce to 2 level classes:
- react-datepicker-input
- react-datepicker-calendar

rafeememon:
Two levels is probably the way to go (since the calendar will not nested), but we probably won't need a `react-datepicker-input` class if it's just an `<input>` element.  In fact I think we should *not* provide styling for the `<input>` element so that it can integrate with forms as easily as possible.

dmnd:
> I think we should not provide styling for the `<input>` element so that it can integrate with forms as easily as possible

:+1: That sounds great to me.

rafeememon:
Continuing on #391, #396

PR #391, opened 2016-03-08
CSS refactor

martijnrusschen:
Proposal as discussed in #388

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -89,7 +89,7 @@ var Calendar = React.createClass({
     return [0, 1, 2, 3, 4, 5, 6].map(offset => {
       const day = startOfWeek.clone().add(offset, 'days')
       return (
-        <div key={offset} className="datepicker__day">
+        <div key={offset} className="react-datepicker__day">
           {day.localeData().weekdaysMin(day)}
         </div>
       )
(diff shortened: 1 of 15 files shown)


martijnrusschen:
This is probably a breaking change since people are relying on our classes.

dmnd:
I'd also be happy with this :+1:

martijnrusschen:
Do we like this new format? I'll rebase and merge this if we like it. (lets keep the input field styling discussion out of this PR)

rafeememon:
I think we should go for it. And yep, addressing the field styling separately in #396

rafeememon:
LGTM :+1:

pinturic:
I still have a question about this: I have 4 calendars in the page and I want them to have different styles. Since the "react-datepicker" className is not editable (ora at least configurable to be addictive)  there is no chance for the 4 calendars to have different styles. Am I correct ?

rafeememon:
Same as #377

Seems reasonable to add the `calendarClassName` prop. PRs to this effect are welcome

pinturic:
There was one #387 as far as I know.

rafeememon:
Yep and see the discussion there for why it wasn't merged

pinturic:
to tell the truth it is not clear what and how to overcome the situation. In my opinion it should be enough if the className is "additive" so that nothing actually present breaks and the other added class is enough from outside to set the style.

rafeememon:
Yep, that's what a PR implementing this should do

pinturic:
okay perfect

PR #392, opened 2016-03-09
use locale for text input date

salmanhoque:
The props locale wasn't being used when typing into the date input field and not using the calendar
and instead would clear the input, even if it was exactly what would be output by the clicking on a date in the calendar.

This passes the locale to moment.

### DIFF
--- a/src/date_input.jsx
+++ b/src/date_input.jsx
@@ -53,7 +53,7 @@ var DateInput = React.createClass({
 
   handleChange (event) {
     var value = event.target.value
-    var date = moment(value, this.props.dateFormat, true)
+    var date = moment(value, this.props.dateFormat, this.props.locale || moment.locale(), true)
     if (date.isValid() && !isDayDisabled(date, this.props)) {
       this.props.setSelected(date)
     } else if (value === '') {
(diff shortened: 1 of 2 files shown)


src/date_input.jsx
```diff
@@ -92,6 +92,7 @@ var DateInput = React.createClass({
     if (this.props.onBlur) {
       this.props.onBlur(event)
     }
+    this.props.handleDone()
```
rafeememon: Don't think we want this line, this causes the picker to close when clicking on the calendar

src/date_input.jsx
```diff
@@ -53,7 +53,7 @@ var DateInput = React.createClass({
 
   handleChange (event) {
     var value = event.target.value
-    var date = moment(value, this.props.dateFormat, true)
+    var date = moment(value, this.props.dateFormat, this.props.locale || moment.locale(), true)
```
rafeememon: Can you give test cases for this?

rafeememon:
After you've made the fixes can you squash your commits into one commit?  Thanks!

salmanhoque:
Hi, thanks for your quick feedback on this. I have updated and added test case for that. If this looks alright to you, could you able to merge it?

rafeememon:
Thanks for the fix!

PR #394, opened 2016-03-10
Clean up DateInput API

rafeememon:
The API of the `DateInput` component is a bit of a mess, and I think there are a lot of easy fixes that can be made.  Luckily it is still internal API and we can work on getting it into a good state before v1.0.  The main things this PR accomplishes are:

* **Decouple from the `DatePicker` component** -- There were a couple things making this tightly integrated with the `DatePicker`, namely things around the `open` prop.  Thinking about how an external consumer would use the `DateInput` set the direction of the cleanup.
* **Pass-thru props** -- Over time there were many requests to add various props to the `<input>`, e.g. `readOnly`, `disabled`, `tabIndex`.  If we start treating the `DateInput` as a traditional `<input>` element, we can pass through *all* props and knock out all of these cases and more, including `onClick`, `onKeyPress`, ...


src/date_input.jsx
```diff
@@ -98,31 +73,15 @@ var DateInput = React.createClass({
...
-
   render () {
     return <input
-        ref="input"
-        type="text"
-        id={this.props.id}
-        name={this.props.name}
+        ref='input'
+        type='text'
+        {...this.props}
```
rafeememon: This is the magic line that lets us pass arbitrary props to the `<input>` element

src/date_input.jsx
```diff
@@ -52,12 +43,22 @@ var DateInput = React.createClass({
...
 
   handleChange (event) {
-    var value = event.target.value
-    var date = moment(value, this.props.dateFormat, this.props.locale || moment.locale(), true)
-    if (date.isValid() && !isDayDisabled(date, this.props)) {
-      this.props.setSelected(date)
-    } else if (value === '') {
-      this.props.setSelected(null)
+    if (this.props.onChange) {
+      this.props.onChange(event)
```
rafeememon: Re #397

PR #397, opened 2016-03-11
Add isValidOrEmpty Event

peterheesterman:
Had the use case to show if a date field was invalid as a user was typing into the text box.  

Also, re-ordered a couple of attribute lists -- my bad.


rafeememon:
Hey @peterheesterman, thanks for contributing.  Unfortunately I don't think this is an API that we can support.  It seems like what you really want to get at is the actual text typed in by the user, which can probably be gotten at by following the chains of refs down to the input element.  I think a better approach here would be to expose the onChange event in the DateInput component, which we can probably work into the refactor of that component.

PR #400, opened 2016-03-12
respect date format change on rerender

gnawjaren:
Hello,
I changed `componentWillReceiveProps` to compare formatted date instead of raw date - and added arguments locale and dateFormat to `safeDateFormat` because props are not set at this time.

Also added test cases for rerendering on format change...

Regards

### DIFF
--- a/src/date_input.jsx
+++ b/src/date_input.jsx
@@ -30,14 +30,16 @@ var DateInput = React.createClass({
 
   getInitialState () {
     return {
-      maybeDate: this.safeDateFormat(this.props.date)
+      maybeDate: this.safeDateFormat(this.props)
     }
   },
 
   componentWillReceiveProps (newProps) {
-    if (!isSameDay(newProps.date, this.props.date)) {
+    if (!isSameDay(newProps.date, this.props.date) ||
+          newProps.locale !== this.props.locale ||
+          newProps.dateFormat !== this.props.dateFormat) {
       this.setState({
-        maybeDate: this.safeDateFormat(newProps.date)
+        maybeDate: this.safeDateFormat(newProps)
       })
     }
   },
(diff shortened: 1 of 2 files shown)


src/date_input.jsx
```diff
@@ -35,9 +36,10 @@ var DateInput = React.createClass({
   },
 
   componentWillReceiveProps (newProps) {
-    if (!isSameDay(newProps.date, this.props.date)) {
+    var newMaybeDate = this.safeDateFormat(newProps.date, newProps.locale, newProps.dateFormat)
```
rafeememon: Can we minimize to change in logic by simply checking if `dateFormat` or `locale` has changed?

gnawjaren: That is what I was up to Initially - but moment global locale is out of the scope then.

rafeememon: It's out of scope anyway, we don't receive notifications when the global locale changes

gnawjaren: Say you have top level locale state, which is in sync with moment (I know violates single source of truth) it will get picked up - without passing props. But since that is not a good approach anyway it does maybe not matter. I will change that. (and remove the test for global change)

rafeememon: Right -- my point is that if the global locale changes, we have no way of knowing until we we-render anyway

rafeememon:
Aha, now I understand what you meant.  Yeah, don't think that case is super important as it's nice to keep render state entirely based on props

src/date_input.jsx
```diff
@@ -1,3 +1,4 @@
+
```
rafeememon: not please remove this

src/date_input.jsx
```diff
@@ -35,9 +36,11 @@ var DateInput = React.createClass({
   },
 
   componentWillReceiveProps (newProps) {
-    if (!isSameDay(newProps.date, this.props.date)) {
+    if (!isSameDay(newProps.date, this.props.date) ||
+          newProps.locale !== this.props.locale ||
+          newProps.dateFormat !== this.props.dateFormat) {
       this.setState({
-        maybeDate: this.safeDateFormat(newProps.date)
+        maybeDate: this.safeDateFormat(newProps.date, newProps.locale, newProps.dateFormat)
```
rafeememon: Can we keep the calling API consistent for safeDateFormat, maybe by passing the entire props object at each caller?

rafeememon:
After you're done making the fixes could you squash into one commit?  Thanks!

martijnrusschen:
This looks good to go

rafeememon:
Thanks @gnawjaren!

PR #404, opened 2016-03-16
Reduced-bloat packaging for npm & webpack

dmnd:
Reduced-bloat packaging for npm & webpack

`react-datepicker`'s npm module has a main file
`./dist/react-datepicker.min.js`. This file is built by webpack, so it
inlines both `classnames` and `react-tether` dependencies.  

The `classnames` and `react-tether` dependencies in package.json are
hence unused by consumers of `react-datepicker`. Right now they'd be
more appropriate in devDependencies. But this commit proposes another
solution.

I'm using `react-datepicker` with npm, and importing via webpack. My app
also uses `classnames`, and could presumably use `react-tether` too. But
since these `react-datepicker` dependencies are inlined, they cannot be
shared with the host application, despite being present in package.json.

This commit proposes that the main file for distribution via npm is
`./lib/datepicker.js`, compiled by Babel, but not webpack. This allows
users of `react-datepicker` to share the same version of `classnames`
and `react-tether` with `react-datepicker`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
   "style": "dist/react-datepicker.min.css",
   "files": [
     "*.md",
-    "dist"
+    "dist",
+    "lib"
   ],
   "keywords": [
     "react",
@@ -42,6 +43,7 @@
     "express": "^4.13.3",
     "extract-text-webpack-plugin": "^1.0.1",
     "grunt": "0.4.5",
+    "grunt-babel": "^6.0.0",
     "grunt-contrib-sass": "0.9.2",
     "grunt-contrib-watch": "0.6.1",
     "grunt-eslint": "^17.3.2",
--- a/src/year_dropdown.jsx
+++ b/src/year_dropdown.jsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import YearDropdownOptions from './year_dropdown_options.jsx'
+import YearDropdownOptions from './year_dropdown_options'
 
 var YearDropdown = React.createClass({
   displayName: 'YearDropdown',
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ node_modules
 npm-debug.log
 .idea
 dist
+lib
 tmp
 coverage
 
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -88,6 +88,7 @@ module.exports = function (grunt) {
       }
     },
 
+    // standalone build for ./dist
     webpack: {
       unmin: mergeWebpackConfig({
         output: {
@@ -107,17 +108,30 @@ module.exports = function (grunt) {
         ]
       }),
       docs: require('./webpack.docs.config')
+    },
+
+    // source build for ./lib
+    babel: {
+      lib: {
+        files: [{
+          expand: true,
+          cwd: 'src/',
+          src: ['**/*.js', '**/*.jsx'],
+          dest: 'lib/'
+        }]
+      }
     }
   })
 
   grunt.loadNpmTasks('grunt-contrib-sass')
   grunt.loadNpmTasks('grunt-scss-lint')
   grunt.loadNpmTasks('grunt-contrib-watch')
+  grunt.loadNpmTasks('grunt-babel')
   grunt.loadNpmTasks('grunt-webpack')
   grunt.loadNpmTasks('grunt-karma')
   grunt.loadNpmTasks('grunt-eslint')
 
   grunt.registerTask('default', ['watch', 'scsslint'])
   grunt.registerTask('travis', ['eslint', 'karma', 'scsslint'])
-  grunt.registerTask('build', ['scsslint', 'webpack', 'sass'])
+  grunt.registerTask('build', ['scsslint', 'babel', 'webpack', 'sass'])
 }


src/year_dropdown.jsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import YearDropdownOptions from './year_dropdown_options.jsx'
+import YearDropdownOptions from './year_dropdown_options'
```
dmnd: This was necessary because babel converts jsx -> js. You might want to lint for this kind of thing in the future; I don't think there's a good reason to be specific about the extension when importing?

package.json
```diff
@@ -5,11 +5,12 @@
   "version": "0.24.0",
   "license": "MIT",
   "homepage": "https://github.com/Hacker0x01/react-datepicker",
-  "main": "dist/react-datepicker.min.js",
+  "main": "lib/datepicker.js",
```
dmnd: This could _possibly_ be a breaking change for some users that aren't using module bundlers. I'm not really sure how to test that since I don't know the use cases very well.

But I think it's a good change. If it still makes sense to publish a standalone build, that can be imported via `require('react-datepicker/dist/react-datepicker.min.js')`.

package.json
```diff
@@ -88,6 +90,7 @@
   },
   "scripts": {
     "build": "NODE_ENV=production grunt build",
+    "build:lib": "NODE_ENV=production babel src --out-dir lib",
```
dmnd: Apologies for not making this part of the grunt script. If you insist, I can figure that out, but this works as is.

src/year_dropdown.jsx
rafeememon: Yep -- I think this is an artifact of the webpack config months ago not including `.jsx` files by default

package.json
rafeememon: This is a breaking change, but we can break this soon.  I think ideally in v1.0 we will stop publishing the `dist` build altogether and only publish the cross-compiled `lib` directory.  When this merges you will be able to include the non-bloated files with `require('react-datepicker/lib/datepicker')` but for now this will need to stay pointing at dist.

package.json
rafeememon: I think we will want this as part of the grunt script so that it runs as part of the prepublish hook.  That way the compilation is completely automated and we don't need to have the `lib` directory committed here

rafeememon:
Hey @dmnd, thanks for putting this together -- this is a change I've been wanting to make for a while.  Summarizing my main feedback:
* I think we should git-ignore the `lib` directory and have it built as part of the prepublish hook
* Let's wait to make the API break (i.e. including `lib` instead of `dist`) until v1.0

Once those are addressed I think this should be good to go.

package.json
rafeememon: Tracking this in #405

src/year_dropdown.jsx
rafeememon: Actually would you mind fixing this in a separate commit?

dmnd: Done; see 46e6fd9e01dd11640d1a6d86a5262e9374dbdf40

dmnd:
> I think we should git-ignore the `lib` directory and have it built as part of the prepublish hook

I can understand that, but it does have one small downside:

When making a PR to a third party library I'm using, I change my application's package.json to depend on my PR instead of the published npm module. This lets me fix bugs I hit and not have to wait for other people to merge my changes, which I think makes me more likely to send fixes upstream.

As far as I understand, this workflow isn't possible if there are files in the npm module that are not part of the git repository. So while it's nice to not have built files checked into git, it does have a small advantage for consumers of the library. Unless there's some other easy way to depend on a PR that I'm unaware of.

That said, I'm happy to change this PR if you don't agree this benefit is worthwhile for your library.

package.json
dmnd: Cool, reverted this line in a9690a1

rafeememon:
We made the decision a while ago to not include built scripts in the repo =) tl;dr they're a maintenance burden and pollute PRs and diffs (more on #303)

For the case you mentioned, npm runs the prepublish hook after install, which I think is for this exact use case.  Hopefully that would work for your case of wanting to use fixes immediately and not wait for a merge and release.

dmnd:
Ok, removed `./lib` from the repo and switched to grunt. Let me know if there's anything else blocking a merge.

rafeememon:
Awesome -- LGTM, but I'd like @martijnrusschen to take a look before we merge

martijnrusschen:
Since, we're excluding /lib in .git now, does this actually change something on how we serve files in our package? Looks like the initial description of the PR is different from the actual implementation now.

rafeememon:
Right now this only adds to the files that we publish in the package, allowing access to the "raw" JS files without the bundled dependencies, and letting the consumer's module bundler take care of the dependencies.  For v1.0 I think we shouldn't publish our current dist build and only publish `lib` since module bundlers are so standard now.

martijnrusschen:
SGTM

rafeememon:
Thanks for the work here @dmnd!

dmnd:
No worries. Any chance I could get a release? After some investigation, depending on git repos that need a build step is pretty broken. See https://github.com/npm/npm/issues/3055

Issue #398, opened 2016-03-11
Internet Explorer 9 support


vguzev:
Hi!

We're currently looking for replacement for our DatePicker component and your project looks great! 
Unfortunately we still have requirement of supporting Internet Explorer 9, because a lot of customers are still using it in banking sphere. Do you support IE9? 
Demo page https://hacker0x01.github.io/react-datepicker/ is not working in IE9. It gives error:
SCRIPT1028: Expected identifier, string or number 
bundle.js, line 91 character 80

P.S. You can get Virtual Machine with IE9 for free from Microsoft's site: https://dev.windows.com/en-us/microsoft-edge/tools/vms/windows/

rafeememon:
After some googling it looks like that error is from how we compile the demo site (appears to be the same  issue described [here](http://jamesknelson.com/using-es6-in-the-browser-with-babel-6-and-webpack/#comment-4187)).  Re: IE9; the best way to find out is to give it a try, and we'd be happy to accept any PRs to make it compatible.

Pizdetss:
Just FYI

The datepicker also has a dependency called 'react-onclickoutside' which doesn't support IE9 and never will. So unless they change this dependency your datepicker will not function correclty in IE9.

rafeememon:
Thanks for digging into that @Pizdetss.  The reasoning for this is detailed on [this issue](https://github.com/Pomax/react-onclickoutside/issues/60) (perfectly valid IMO).  In another issue, [this comment](https://github.com/Pomax/react-onclickoutside/pull/14#issuecomment-103211295) describes a possible workaround, but I haven't tried it out.

vguzev:
@Pizdetss @rafeememon We've already created a clone of react-onclickoutside with IE9 support: https://www.npmjs.com/package/react-onclickoutside-ie9

vguzev:
Please, see PR for this issue: https://github.com/Hacker0x01/react-datepicker/pull/406

rafeememon:
@vguzev, unfortunately I'm skeptical of including dependencies on forks of repositories because of the risk of the forks being abandoned, and I'd like to try to stay with the main plugin.  I'm also now skeptical of supporting IE9 for the reasons mentioned in the main plugin.  Can you try the workaround I mentioned in https://github.com/Hacker0x01/react-datepicker/issues/398#issuecomment-196706326 ?

rafeememon:
Also, from the `react-onclickoutside` readme:

```
[...]

PR #406, opened 2016-03-16
IE9 support

vguzev:
Fix for issue https://github.com/Hacker0x01/react-datepicker/issues/398


rafeememon:
Use the `classlist-polyfill` for this instead, as mentioned in #398

PR #407, opened 2016-03-16
Fix CSS refactor regressions

rafeememon:
There were a couple regressions caused by the CSS refactor in #391 -- only mismatched class names.

cc @martijnrusschen

### DIFF
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -155,7 +155,7 @@
   }
 }
 
-.react-datepicker-input {
+.react-datepicker__input {
   position: relative;
   font-size: 13px;
   border-radius: $border-radius;
@@ -169,13 +169,9 @@
     border-color: $border-color;
     box-shadow: inset 0 2px 2px #e9e9e9, 0 0 10px 0 rgba(73, 107, 125, .3);
   }
-
-  &:not(:valid) ~ .close-icon {
-    display: none;
-  }
 }
 
-.react-datepicker-input__container {
+.react-datepicker__input-container {
   position: relative;
 }
 


rafeememon:
It looks like these were a step towards the BEM classname model we discussed.  This PR fixes the breaks; let's talk about getting BEM working soon though.

src/stylesheets/datepicker.scss
```diff
@@ -155,7 +155,7 @@
   }
 }
 
-.react-datepicker-input {
+.react-datepicker__input {
```
martijnrusschen: Not sure what we discussed here but I thought we were going with a top level `react-datepicker-input` class first. In that case, we should fix the jsx instead.

src/stylesheets/datepicker.scss
```diff
@@ -170,12 +170,12 @@
     box-shadow: inset 0 2px 2px #e9e9e9, 0 0 10px 0 rgba(73, 107, 125, .3);
   }
 
-  &:not(:valid) ~ .close-icon {
+  &:not(:valid) ~ .react-datepicker__close-icon {
```
martijnrusschen: My bad, probably missed those.

src/stylesheets/datepicker.scss
```diff
@@ -170,12 +170,12 @@
...
     display: none;
   }
 }
 
-.react-datepicker-input__container {
+.react-datepicker__input-container {
```
martijnrusschen: Same as first comment.

src/stylesheets/datepicker.scss
rafeememon: No prob =)

src/stylesheets/datepicker.scss
rafeememon: Yeah -- like I mentioned in my comment let's handle all of those in a separate PR

src/stylesheets/datepicker.scss
```diff
@@ -169,13 +169,9 @@
     border-color: $border-color;
     box-shadow: inset 0 2px 2px #e9e9e9, 0 0 10px 0 rgba(73, 107, 125, .3);
   }
-
```
martijnrusschen: This isn't used anymore?

rafeememon: The display of the close icon is controlled in JS instead by not adding it to the DOM if nothing is selected

martijnrusschen: :+1:

martijnrusschen:
LGTM, didn't test it but I assume you tested the new CSS.

Issue #396, opened 2016-03-10
Proposal to remove DateInput styles


rafeememon:
Continuing discussion from #388, related to work on #394 

The [styles applied to the DateInput](https://github.com/Hacker0x01/react-datepicker/blob/v0.24.0/src/stylesheets/datepicker.scss#L162) make the `<input>` element look nice enough for most situations.  However, I think it's a good idea to move in the direction of exposing the `<input>` element directly, allowing consumers to treat the datepicker component as a traditional form element rather than something special.  A case I've run into is that it's currently difficult to integrate with a few styling frameworks (e.g. [elemental](http://elemental-ui.com/)) because:

* The top-level element exposed by the datepicker is a `div`, not an `input`
* The `input` has CSS styles that need to be overridden to fit in with the framework

I think a good first step here is to remove the `.datepicker__input` styles, and perhaps move them to the examples stylesheet so that it still demos well.  The new solution would be to pass the required `className` prop to the datepicker to fit in with a styling framework or custom styles.  What do you think?

cc @martijnrusschen, @RSO

rafeememon:
Removed in #412

PR #412, opened 2016-03-17
Remove date input styles

rafeememon:
Addresses #396

Removes the default styling applied to the `<input>` in `DateInput`, allowing for greater styling flexibility.  Moves the default styles to the CSS of the demo site so that the examples still look nice.

### DIFF
--- a/docs-site/src/style.scss
+++ b/docs-site/src/style.scss
@@ -93,3 +93,18 @@ strong {
     }
   }
 }
+
+input {
+  font-size: 13px;
+  border-radius: 4px;
+  box-shadow: inset 0 2px 2px #e9e9e9;
+  border: 1px solid #aeaeae;
+  line-height: 16px;
+  padding: 6px 10px 5px;
+
+  &:focus {
+    outline: none;
+    border-color: #aeaeae;
+    box-shadow: inset 0 2px 2px #e9e9e9, 0 0 10px 0 rgba(73, 107, 125, .3);
+  }
+}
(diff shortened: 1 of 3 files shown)


src/date_input.jsx
```diff
@@ -93,8 +91,7 @@ var DateInput = React.createClass({
         {...this.props}
         value={this.state.maybeDate}
         onBlur={this.handleBlur}
-        onChange={this.handleChange}
-        className={classnames('react-datepicker__input', this.props.className)} />
```
rafeememon: `className` can still be specified by passing it in as a prop

docs-site/src/style.scss
```diff
@@ -93,3 +93,21 @@ strong {
     }
   }
 }
+
+.react-datepicker__input-container {
```
martijnrusschen: Where are we using this now in the examples? Shouldn't we pass this into one of the examples?

rafeememon: Moving the styles to the docs-site stylesheet will apply the old styles to all of the examples on the demo site -- is that what you mean?

martijnrusschen: No I thought with removing the className from the from the input field wouldn't render the input with the correct class anymore. We should apply this somewhere in an example right?

rafeememon: It turns out we won't need a default class anymore; from what I could tell it was only to apply the style, and I don't think it should be there if we want to move towards treating it as a plain `<input>`.  We have a "Custom class name" example that looks like it broke...looking into it

rafeememon: The problem was from the specificity of CSS selectors; refactored the demo page styles to make more sense.  No implementation problems =)

martijnrusschen:
LHTM

PR #417, opened 2016-03-19
Style days outside month

### LINKED ISSUE #296: Add CSS class for days not in current month
The calendar shows days adjacent to the current month (end of previous month, beginning of next month), but they look the same as days in the current month.  We should add a class to days not in the current month (maybe `datepicker__day--outside-month`) and style them slightly differently so that they can be distinguished.

martijnrusschen:
Trying to solve the issue where the days outside a month have the same styling as days in the same month.

I’m not totally convinced by this solution yet, but maybe this could help as some inspiration.

1) Changed hover styling for normal days to the blue (selected) state the make this more explicit. Personally I like this change
2) Added styling for days outside the month by giving it a grey background.

Examples:
screen shot 2016-03-19 at 15 30 34
screen shot 2016-03-19 at 15 30 26

Fixes #296

### DIFF
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -126,8 +126,9 @@
   cursor: pointer;
 
   &:hover {
+    background-color: darken($selected-color, 5%);
     border-radius: $border-radius;
-    background-color: $background-color;
+    color: #fff;
   }
 
   &--today {
@@ -151,6 +152,17 @@
 
     &:hover {
       background-color: transparent;
+      color: $muted-color;
+    }
+  }
+
+  &--outside-month {
+    border-radius: $border-radius;
+    background-color: $background-color;
+
+    &:hover {
+      background-color: darken($selected-color, 5%);
+      color: #fff;
     }
   }
 }


rafeememon:
Not sure I like the way this looks.  I'm tempted to think that we should completely defer to the consumers to style this, which is now possible since we give the `--outside-month` class.  This situation is similar to what we have for `--weeked` as well.

martijnrusschen:
> I think it draws more attention to those dates rather than deemphasizing them, which is what I think we should aim for.

Exactly, that's the reason I'm still doubting about this.

I'm not sure if we shouldn't add some basic styling here. This feels as a pretty basic feature and something that should be included in the datepicker by default.

rafeememon:
> I'm not sure if we shouldn't add some basic styling here. This feels as a pretty basic feature and something that should be included in the datepicker by default.

Agreed, but I don't think we should feel pressured to pick a style just for the sake of picking when it's not clear what the right thing to do is.  In the end the easiest thing to do is to do nothing and defer to consumers =P

martijnrusschen:
Of course :) Trying to give this some attention by creating this PR. Hoping this could gives others some inspiration on how to solve this.

iammai:
We could make the inside-the-month days a grey color and the outside-the-month days a white background color. alternatively, we could exclude the out-side-the-month days altogether for clarity like in https://jqueryui.com/datepicker/

jochenberger:
What about changing the `color` to a lighter shade instead of changing the `backgroundColor`?

jochenberger:
http://codepen.io/jochenberger/pen/RoxqyZ

martijnrusschen:
I think the proposal in http://codepen.io/jochenberger/pen/RoxqyZ looks too much like disabled days, and that could confuse people when picking a date. What do you think?

martijnrusschen:
Closing this out because of inactivity.

PR #421, opened 2016-03-20
Add grunt-cli to devDependencies

dmnd:
`npm run build` depends on `grunt`. Before this commit, that required
a global version of `grunt-cli`. Now the project installs a local
version.


rafeememon:
SGTM

PR #423, opened 2016-03-21
Fix onBlur handling

alurim:
Thanks for writing this component, I'm very happy with it, but encountered the following problem. I had solved it by adding this code to my own onBlur handler but figure that this should probably exist in the library so others can benefit.

Problem: When onBlur is fired we pass the event up rather than then the current date. This is problematic because:
a) we're expecting the date (even in the onBlur example in the docs-site)
b) if gibberish is entered into the input, we don't sanitize it

Solution: When the input is blurred we should try to format the current value in the input and if it's valid pass that as the value in onBlur or otherwise the currently selected date

Test Plan:
- Before the onBlur example wasn't even working! Now it is.
- Try blurring with a valid date, make sure that the onBlur example prints that date out
- Try blurring with gibberish in it, make sure that the onBlue example prints out the date that was there last and the input is changed to reflect it

### DIFF
--- a/src/date_input.jsx
+++ b/src/date_input.jsx
@@ -76,7 +76,8 @@ var DateInput = React.createClass({
       maybeDate: this.safeDateFormat(this.props)
     })
     if (this.props.onBlur) {
-      this.props.onBlur(event)
+      const blurredValue = moment(event.target.value, this.props.dateFormat)
+      this.props.onBlur(blurredValue.isValid() ? blurredValue : this.props.date)
     }
   },
 


rafeememon:
Hey @alurim, thanks for submitting this, but unfortunately I don't think we can include it.  We're trying to move in a direction where the `DateInput` component behaves as closely to a standard `<input>` as possible, and adding non-standard event propagation is a step away from that.  A better solution here is to instead listen for `onChange` events, and fix the docs site =)

rafeememon:
Closing, but feel free to open an issue if you want to discuss further

alurim:
@rafeememon thanks for the quick feedback, I'll keep the onBlur that I added in my code that does basically what I put up for review as I'm using `redux-form` (which expects `onBlur` to either return the current value or an event which has a target element with the current value). Totally understand your reasoning though!

rafeememon:
> an event which has a target element with the current value

If this is the protocol that `redux-form` is expecting, that must be the standard behavior of `onBlur` for `<input>`s, right?  (I know it's the protocol for at least `onChange`.)  I browsed through our code and the blur event from the `DateInput` is eventually passed to the `onBlur` prop of `DatePicker`.  I'm not familiar with `redux-form` but I'm surprised that didn't work.

DriesVandermeulen:
@rafeememon The current demo page does not describe this behaviour thought. 

From [demo](https://hacker0x01.github.io/react-datepicker/#example-13)
```javascript
handleOnBlur: function (date) {
    if (date === null) {
        console.log('selected date: %s', date);
    }
    else {
        console.log('selected date: %s', date.format('DD/MM/YYYY'));
    }
};
```
It also bugs on the demo page

`Uncaught TypeError: date.format is not a function`

rafeememon:
Agreed; a fix for the demo page is welcome :smile:

dpwrussell:
@alurim Thanks for this, I was having the exact same problem with `redux-form`.

PR #428, opened 2016-03-24
Add inline-block to react-datepicker__input-container

dmnd:
`input` elements are inline-block but `div` is a block element.

react-datepicker wraps its input with a div. As far as I can tell, the
purpose of the wrapper is setting up a positioned container for the
absolutely positioned `react-datepicker__close-icon`. But since the
`input` isn't a block element, the close-icon doesn't line up with the
`input`. Making the wrapper `div` have `display: inline-block` fixes 
that.

Before:

image

After:

image

### DIFF
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -157,6 +157,7 @@
 
 .react-datepicker__input-container {
   position: relative;
+  display: inline-block;
 }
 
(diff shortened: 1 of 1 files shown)


martijnrusschen:
Thanks!

matthewmueller:
is there anyway to disable this? sometimes it's nice to have it be a block element.

PR #431, opened 2016-03-30
Use includeDates to derive implied minDate or maxDate for getDateInVi…

dlwalsh:
Use case: I had a list of dates starting in April, but my calendar still defaulted to a March display. I could fix this by explicitly setting minDate. But this seemed redundant, since the minimum date can be derived from the includeDates.

Solution: I've rewritten getDateInView() to derive an implied minimum date from includeDates. That is, the earliest current or future date. For symmetry, I've also derived an implied maximum date where no implied minimum exists.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -3,6 +3,7 @@ import YearDropdown from './year_dropdown'
 import Month from './month'
 import React from 'react'
 import { isSameDay, allDaysDisabledBefore, allDaysDisabledAfter } from './date_utils'
+import minBy from 'lodash/minBy'
 
 var Calendar = React.createClass({
   displayName: 'Calendar',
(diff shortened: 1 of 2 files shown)


dlwalsh:
Crap! My own unit test broke. It worked fine for me locally.

I'll take a look at this later and resubmit.

src/calendar.jsx
```diff
@@ -45,10 +46,19 @@ var Calendar = React.createClass({
   },
 
   getDateInView () {
-    const { selected, minDate, maxDate } = this.props
+    const { selected, includeDates, minDate, maxDate } = this.props
     const current = moment()
     if (selected) {
       return selected
+    } else if (includeDates) {
```
rafeememon: A day is enabled if it is both in includeDates _and_ between minDate and maxDate, so this needs to take that into account as well

src/calendar.jsx
```diff
@@ -3,6 +3,7 @@ import YearDropdown from './year_dropdown'
 import Month from './month'
 import React from 'react'
 import { isSameDay, allDaysDisabledBefore, allDaysDisabledAfter } from './date_utils'
+import minBy from 'lodash/minBy'
```
rafeememon: We're trying to keep lodash out of consumer dependencies, if possible

src/calendar.jsx
dlwalsh: Good point!

src/calendar.jsx
dlwalsh: OK. I'll refactor it out.

PR #432, opened 2016-03-30
Include dates enhancement

dlwalsh:
This is a resubmission of [#431](https://github.com/Hacker0x01/react-datepicker/pull/431).

> Use case: I had a list of dates starting in April, but my calendar still defaulted to a March display. I could fix this by explicitly setting minDate. But this seemed redundant, since the minimum date can be derived from the includeDates.

> Solution: I've rewritten getDateInView() to derive an implied minimum date from includeDates. That is, the earliest current or future date. For symmetry, I've also derived an implied maximum date where no implied minimum exists.

I've refactored getDateInView() to remove the lodash dependency, and take into account that minDate and maxDate can be used alongside includeDates.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -3,6 +3,7 @@ import YearDropdown from './year_dropdown'
 import Month from './month'
 import React from 'react'
 import { isSameDay, allDaysDisabledBefore, allDaysDisabledAfter } from './date_utils'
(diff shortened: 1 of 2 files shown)


dlwalsh:
Pull request withdrawn.

My logic isn't quite right. And the unit tests I wrote for it aren't good enough.

I will resubmit tomorrow.

PR #435, opened 2016-03-30
Include dates enhancement

dlwalsh:
This is my third attempt at this change. Aborted attempts [here](https://github.com/Hacker0x01/react-datepicker/pull/431) and [here](https://github.com/Hacker0x01/react-datepicker/pull/432). To quote the original intent of this change:

> Use case: I had a list of dates starting in April, but my calendar still defaulted to a March display. I could fix this by explicitly setting minDate. But this seemed redundant, since the minimum date can be derived from the includeDates.

> Solution: I've rewritten getDateInView() to derive an implied minimum date from includeDates. That is, the earliest current or future date. For symmetry, I've also derived an implied maximum date where no implied minimum exists.

I've also taken into account that minDate and maxDate can be used alongside includeDates. Some additional unit tests have been added to validate the new logic.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -3,6 +3,7 @@ import YearDropdown from './year_dropdown'
 import Month from './month'
 import React from 'react'
 import { isSameDay, allDaysDisabledBefore, allDaysDisabledAfter } from './date_utils'
+import minBy from 'lodash/minBy'
 
 var Calendar = React.createClass({
   displayName: 'Calendar',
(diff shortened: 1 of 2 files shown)


dlwalsh:
[#435](https://github.com/Hacker0x01/react-datepicker/pull/436) is indeed a more elegant solution. Thanks!

rafeememon:
No prob!  Closing this PR in favor of #436

PR #437, opened 2016-04-03
Manually control clickoutside class

rafeememon:
If the default clickoutside class were to change it would be an API break, so theoretically we are already protected against that.  But manually specifying the class avoids having to match the "magic" string in the clickoutside library.


rafeememon:
Failing because of #438

PR #439, opened 2016-04-03
Transpile to .js

rafeememon:
Currently the file extensions from `src` are being preserved, which means that the transpiled files have the `.jsx` extension.  This makes it semi-painful to consume `lib` because the module bundler must be configured to recognize `.jsx`.

In the long run I think we should probably just rename `.jsx` files in `src` to `.js`

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -117,7 +117,8 @@ module.exports = function (grunt) {
           expand: true,
           cwd: 'src/',
           src: ['**/*.js', '**/*.jsx'],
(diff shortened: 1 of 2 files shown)


rafeememon:
@dmnd, does this sound good to you?

dmnd:
Absolutely 👍

Thanks for asking!

rafeememon:
Awesome!

martijnrusschen:
Nice work!

Issue #440, opened 2016-04-06
Add CSS class for days name


montogeek:
Hi!

Thanks for this component :)

It would be nice to add a class for the day names: Su Mo Tu We Th Fr Sa.
Something like `__days`, I am happy submitting a PR for this, let me know if it works for you.

I am suggesting it because I need to style them :)

rafeememon:
Yeah, I don't think it makes sense for the day names to have the same class as the dates (`react-datepicker__day`).  I think the day names should probably be separated into a separate class, probably `react-datepicker__day-name`.

montogeek:
Hi @rafeememon I said `__days` but `__day-name` is good.
I will create a PR later today!

Thank you :)

rafeememon:
Yeah please call it `react-datepicker__day-name`, I think it's much more descriptive of what's going on here

montogeek:
I will! :D

PR #441, opened 2016-04-06
Add CSS class to day name

montogeek:
Implements https://github.com/Hacker0x01/react-datepicker/issues/440

Added a CSS class for days name.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -159,7 +159,7 @@ var Calendar = React.createClass({
           {this.renderCurrentMonth()}
           {this.renderYearDropdown()}
           {this.renderNextMonthButton()}
-          <div>
+          <div className="react-datepicker__day-name">
             {this.header()}
           </div>
         </div>
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -91,7 +91,7 @@ var Calendar = React.createClass({
     return [0, 1, 2, 3, 4, 5, 6].map(offset => {
       const day = startOfWeek.clone().add(offset, 'days')
       return (
-        <div key={offset} className="react-datepicker__day">
+        <div key={offset} className="react-datepicker__day react-datepicker__day-name">
           {day.localeData().weekdaysMin(day)}
         </div>
       )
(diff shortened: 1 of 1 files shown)


src/calendar.jsx
```diff
@@ -159,7 +159,7 @@ var Calendar = React.createClass({
           {this.renderCurrentMonth()}
           {this.renderYearDropdown()}
           {this.renderNextMonthButton()}
-          <div>
+          <div className="react-datepicker__day-name">
```
rafeememon: I was envisioning this fix as changing the CSS class of the day headers, as in https://github.com/Hacker0x01/react-datepicker/blob/v0.25.0/src/calendar.jsx#L92

I think changing that class from `react-datepicker_-day` to `react-datepicker__day-name` makes much more sense

montogeek: We can't change it since it will lose `react-datepicker__day` styles, we have two alternatives:
Add `react-datepicker__day-name` class or add it to its container.

rafeememon: Right, I think it should be refactored to make it look right with the `__day-name` class instead of `__day`.  The things they will have in common are defined by variables that can be reused.

montogeek: There are several ways:
1. Change the class to `__day-name` and in the SCSS extend `__day`.
2. Add the class `__day-name` without change the SCSS.

Since this project is using BEM, I would vote for the second option, is another day, the modifier is that is the name day.

Let me know what works best for you :)

rafeememon: Technically `-day` isn't a modifier, and I don't think it should be because the day names and dates don't have much in common other than width and height.  Go with #2

src/calendar.jsx
```diff
@@ -91,7 +91,7 @@ var Calendar = React.createClass({
     return [0, 1, 2, 3, 4, 5, 6].map(offset => {
       const day = startOfWeek.clone().add(offset, 'days')
       return (
-        <div key={offset} className="react-datepicker__day">
+        <div key={offset} className="react-datepicker__day react-datepicker__day-name">
```
rafeememon: Please remove the `__day` class entirely

rafeememon:
See #443

PR #443, opened 2016-04-08
Day name class

### LINKED ISSUE #440: Add CSS class for days name

rafeememon:
Fixes #440, continues #441

Separates weekday headers into a new CSS class

### DIFF
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -104,13 +104,6 @@
   }
 }
 
-.react-datepicker__week-day {
-  color: $muted-color;
-  display: inline-block;
-  width: $item-size + (2 * $day-margin);
-  line-height: $item-size;
-}
-
 .react-datepicker__month {
   margin: $datepicker__margin;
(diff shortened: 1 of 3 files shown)


src/stylesheets/datepicker.scss
```diff
@@ -104,25 +104,22 @@
...
-  width: $item-size + (2 * $day-margin);
-  line-height: $item-size;
-}
-
 .react-datepicker__month {
   margin: $datepicker__margin;
   text-align: center;
 }
 
+.react-datepicker__day-name,
```
rafeememon: I think this should be the way we do it -- day names and days are fundamentally different and shouldn't inherit from the CSS tree

montogeek:
Thank you :)

Issue #444, opened 2016-04-08
React v15.0


tkawachi:
React v15.0 was released.
http://facebook.github.io/react/blog/2016/04/07/react-v15.html

martijnrusschen:
Will upgrade this weekend if no-one beats me.

rafeememon:
We need https://github.com/souporserious/react-tether to upgrade before we can do this

martijnrusschen:
Opened a PR for react-tether: https://github.com/souporserious/react-tether/pull/15

joelzimmer:
Hey guys, when will this be merged with a new version?

halloffame:
Looks like a new version of react-tether was released with the dependency change
v0.5.1
https://www.npmjs.com/package/react-tether

pocketjoso:
Would be great to get this bumped!

rafeememon:
Looks like [redbox-react](https://github.com/KeywordBrain/redbox-react) also needs an updated dependency.  Also could ditch this entirely; the console is good enough for error messages IMO

joelzimmer:
We'd really appreciate a version bump since it would help us upgrade one of our projects to use React 15.

halloffame:
I'm all for ditching dependencies if they are not really necessary.

martijnrusschen:
Lets try to keep redbox-react in the stack is possible. It's a great tool when developing the examples site.

@halloffame It's just a devDependency. Doesn't hurt the component itself.

I'll try to make the dependency compatible with React 15.

rafeememon:
> It's just a devDependency. Doesn't hurt the component itself.

TBH I think this makes it even more silly that it's blocking the upgrade.  :stuck_out_tongue:  Let's hope that your PR gets merged soon.

martijnrusschen:
Agree lets get rid of redbox-react for now :)

rafeememon:
Unfortunately there's yet another issue -- the `react-tether` library now only accepts strings (i.e. query selectors) instead of objects (i.e. DOM elements) for the `renderElementTo` prop.  This will need to be accounted for as well before we can upgrade.

rafeememon:
`renderElementTo` fix: https://github.com/souporserious/react-tether/pull/19

michaelzoidl:
Hey guys, is there some work-around to get the `react-datepicker` working?
In another project just removing `react v0.14` from the `peerDependencies` in the `package.json` worked, seems like this works not for this project.

[...]

PR #451, opened 2016-04-15
Allow react 15

rafeememon:
Re #444, #464

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,9 +8,6 @@
             "transform": "react-transform-hmr",
             "imports": ["react"],
             "locals": ["module"]
-          }, {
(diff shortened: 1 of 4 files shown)


martijnrusschen:
Depends on react-redbox. Opened a PR to support React 15: https://github.com/KeywordBrain/redbox-react/pull/52

Let see how long this takes. It this takes more than a few days, we should remove this dependency.

rafeememon:
Need to remove `react-transform-catch-errors` as well...

rafeememon:
This is finally ready to go!

martijnrusschen:
Awesome! Great work!

PR #453, opened 2016-04-18
Prevent setOpen after clicking the input meanwhile component is disabled

enriquemgc:
Fix issue w.r.t disabled state.

Currently, if you disable the date picker and click on it, nothing happens until the date picker  is enabled again. Then, it shows the calendar without any user interaction.

This fix prevents this from happen, disabling any status change meanwhile the component is disabled.

### DIFF
--- a/test/datepicker_test.js
+++ b/test/datepicker_test.js
@@ -24,6 +24,18 @@ describe('DatePicker', () => {
     expect(datePicker.refs.calendar).to.exist
   })
 
+  it('should not show calendar when focusing on date input and disabled turns false', function () {
+    var divNode = document.createElement('div')
+    var disabledDatePicker = ReactDOM.render(<DatePicker disabled/>, divNode)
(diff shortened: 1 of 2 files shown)


rafeememon:
Hey @enriquemgc, thanks for the fix.  Could you write a unit test for this?

test/datepicker_test.js
```diff
@@ -24,6 +24,18 @@ describe('DatePicker', () => {
     expect(datePicker.refs.calendar).to.exist
   })
 
+  it('should not show calendar when focusing on date input and disabled turns false', function () {
+    var divNode = document.createElement('div')
+    var disabledDatePicker = ReactDOM.render(<DatePicker disabled/>, divNode)
```
rafeememon: Can we use `TestUtils.renderIntoDocument` for this instead?

rafeememon:
Thanks!

PR #456, opened 2016-04-20
Inline calendar

kvasa:
Hi,

I needed inline version of calendar for my project (calendar rendered directly without TetherComponent), so I added new prop `inline` to achieve this behavior for the user and implemented this functionality. I also added new example with `inline` prop.

I can use my enhanced component directly, but maybe you will find this feature useful for other people and you will merge it directly into your code.

Thank You very much for awesome date picker!
Michal

### DIFF
--- a/docs-site/src/example_components.jsx
+++ b/docs-site/src/example_components.jsx
@@ -19,6 +19,7 @@ import DateRange from './examples/date_range'
 import TabIndex from './examples/tab_index'
 import YearDropdown from './examples/year_dropdown'
 import Today from './examples/today'
+import Inline from './examples/inline'
 
 import 'react-datepicker/dist/react-datepicker.css'
 import './style.scss'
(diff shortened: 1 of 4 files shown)


martijnrusschen:
I like the idea. Can we get some test coverage for this?

kvasa:
Sure, I will add some tomorrow.

kvasa:
So I added 3 simple tests mainly to check if TetherComponent is rendered or not. Is it ok like this or should I add some more?

rafeememon:
Hey @kvasa, thanks for the contribution, this looks good.  I'll wait for the final approval from @martijnrusschen before we merge this in.

For SA: I think the way we'll want to handle this in v1.0 is to remove the `inline` prop and instead allow direct consumption of the `Calendar` component.  However this is a good stopgap solution until we get to that world.

PR #469, opened 2016-04-28
Remove react-tether dependency

rafeememon:
Part of React v15 support (#444)

Unfortunately react-tether has changed their API, so we need to inline the component until the old API can be supported.  Perhaps we can revisit adding it back as a dependency in the future.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,5 @@
 {
-  "presets": ["react", "es2015"],
+  "presets": ["es2015", "stage-0", "react"],
   "env": {
     "development": {
       "plugins": [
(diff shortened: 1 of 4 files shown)


martijnrusschen:
Okay lets do this, and move back to the NPM package as soon as your PR is merged.

rafeememon:
> move back to the NPM package as soon as your PR is merged

TBH, I don't think this is a good idea -- the code base quality is poor and has bitten us a couple times in the past.  I think we can maintain a much better `TetherComponent` in this library, and perhaps extract it out after iterating on it.

martijnrusschen:
Sounds good to me!

PR #471, opened 2016-04-30
Add react and react-dom to devDependencies for the docs

martijnrusschen:
I tried installing React 15 and encountered issues with installing the modules. Looks like we need react and react-dom in dev mode for our examples page. Not sure if this was related to React 15 but worth looking into it.


rafeememon:
:+1:

PR #472, opened 2016-05-02
Make input to have context auto-selected on focus

firedev:


rafeememon:
This is a pretty significant UX change so unfortunately I don't think we can include it in this form

firedev:
Sorry, what do you mean exactly? I will make tests pass.

rafeememon:
I mean that this changes the current user experience enough that it should not be the default behavior, and it will break current expectations of many workflows.  If you want to discuss this further please open an issue

PR #474, opened 2016-05-03
onFocus prop is passed to the input

firedev:
It was passed before but it was not explicitly mentioned


rafeememon:
Hey @firedev, it's actually by design that `onFocus` isn't explicitly mentioned; the idea is that the `DateInput` component behaves as closely to a standard `input` as possible by passing through all props with the `...this.props` line.  There are many props not explicitly mentioned, e.g. `placeholder`, `readOnly`

Issue #320, opened 2016-01-25
Option to set default date on open


TheSisb:
Currently in all examples the picker opens to the current date.  I want to open it to another date/year without setting a value.
Any undocumented option for this?

rafeememon:
Hey @TheSisb, there's no hidden option for this but it sounds like a good request.  This could be an `initialDate` prop that, if specified, will set the calendar's view to this date on first load.

uniphil:
We do this by hooking into `onFocus`, eg.:
```js
const SomeComponent = React.createClass({
  getInitialState() {
    return { selected: null };
  },
  handleDateFocus() {
    if (this.state.selected === null) {
      this.replaceState({ selected: doAnyKindOfMagicDefaultLookupHere() });
    }
  },
  handleChange(newDate) {
    this.replaceState({ selected: newDate });
  },
  render() {
    return (
      <DatePicker
          onFocus={this.handleFocus}
          onChange={this.handleChange}
          selected={this.state.selected} />
    );
  }
});
```

Which works well for us since we do some computation at focus time to determine which date should be default. However, for simpler cases this is probably annoying to have to set up.

PetarIvancevic:
You can do this by creating a new date and pass it to moment. If you store it in your component state then you can do this:

```
const defaultDate = moment(new Date('YYYY', 'MM', 'DD'));
this.setState({ startDate: defaultDate });

...

<DatePicker selected={(this.state) ? this.state.startDate : null} />
```

Hope it helps someone.

martijnrusschen:
@PetarIvancevic's answer looks probably like the way to go here. Do we want to document this somewhere?

rafeememon:
@PetarIvancevic's solution sets the selected date, but @TheSisb was looking for a way without doing that.  @uniphil's code gets the right functionality but in an "unsupported" way.

lbntly:
Hi @TheSisb, just opened a PR for this if you want to take a look https://github.com/Hacker0x01/react-datepicker/pull/486

rafeememon:
Fixed in #486

PR #486, opened 2016-05-17
add openToDate prop

lbntly:
Resolves issue https://github.com/Hacker0x01/react-datepicker/issues/320. 

Allows you to specify a date at which to open the date picker initially (without setting a selected value).


rafeememon:
Hey @lbntly, thanks for the contribution, this looks good!  Could you remove the compiled docs files (bundle.js and style.css) from the PR -- we automatically generate those when we make a release.  Other than that, this should be good to go.

lbntly:
@rafeememon Oh right, my bad! I've gone ahead and removed those changes, would you prefer I rebase/squash those commits before merging or leave as is?

martijnrusschen:
Nice work!

rafeememon:
Thanks @lbntly!  We can squash on merge so leaving it as-is is fine

Issue #465, opened 2016-04-28
how to set the datapicker input tag autocomplete off


gentlythink:
how to set the datapicker input tag  autocomplete attribute  "off"?
It happens when i use the component on ie11 browser. could you help me to solve this problem
![image](https://cloud.githubusercontent.com/assets/2908871/14882932/dd98e604-0d6d-11e6-90aa-2555d62ee815.png)

gentlythink:
![image](https://cloud.githubusercontent.com/assets/2908871/14882895/b298b826-0d6d-11e6-98d6-3b2da3013bbb.png)

rafeememon:
We don't currently support the `autocomplete` attribute but a PR is welcome.

rafeememon:
Added in #489

PR #489, opened 2016-05-24
Support autoComplete prop

lbntly:
For issue https://github.com/Hacker0x01/react-datepicker/issues/465

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -15,6 +15,7 @@ var DatePicker = React.createClass({
   displayName: 'DatePicker',
 
   propTypes: {
+    autoComplete: React.PropTypes.string,
     className: React.PropTypes.string,
     dateFormat: React.PropTypes.string,
     dateFormatCalendar: React.PropTypes.string,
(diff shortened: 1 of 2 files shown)


rafeememon:
Thanks @lbntly, LGTM.  I think ideally we will pass all props from the `<DatePicker>` into the `<input>` so that we automatically support all possible attributes, but unfortunately we're a couple refactors away from that

technificentConsulting:
Is this change live?  adding autocomplete="false" does not seem to work.

leahjlou:
@technificentConsulting Did you try camelcase `autoComplete`? I also know sometimes different browsers handle this weird.

technificentConsulting:
@leahjlou thank you and yes I've tried that as well.  Doesn't seem to get carried forward when converted to a div :(

moeloubani:
> @leahjlou thank you and yes I've tried that as well. Doesn't seem to get carried forward when converted to a div :(

try autoComplete='off'

Divinare:
autoComplete='off' doesn't seem to work. Also it seems like there is just autoComplete prop added but nothing is done with it.

ultrinnan:
autoComplete="off" works as expected

PR #490, opened 2016-05-24
react-onclickoutside discontinued support for usage as mixin

balsamiqstefano:
Addresses the following error when using React v15: `Invariant Violation: ReactClass: You're attempting to use a component class or function as a mixin. Instead, just use a regular object`

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -3,8 +3,9 @@ import YearDropdown from './year_dropdown'
 import Month from './month'
 import React from 'react'
 import { isSameDay, allDaysDisabledBefore, allDaysDisabledAfter, getEffectiveMinDate, getEffectiveMaxDate } from './date_utils'
+import onClickOutside from 'react-onclickoutside'
 
-var Calendar = React.createClass({
+var Calendar = onClickOutside(React.createClass({
   displayName: 'Calendar',
 
   propTypes: {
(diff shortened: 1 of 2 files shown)


balsamiqstefano:
Sorry, I can't figure out why the tests are failing.

```
Uncaught TypeError: (0 , _reactOnclickoutside2.default) is not a function
  at /Users/stefano/projects/react-datepicker/test/index.js:34799 <- webpack:///src/year_dropdown_options.jsx:9:967
```

The code seems to work though.

rafeememon:
The dependency version needs to be bumped as well; the API change happened in v5 but we depend on `^4.8.0` (handy semver calculator: http://semver.npmjs.com/)

balsamiqstefano:
Doh, of course!

So, ok, I tried bumping up the revision, but now a whole bunch of other tests fail. :-(

martijnrusschen:
Can you post the log here? Or push the change and let Travis build it?

rafeememon:
@balsamiqstefano, any luck here?  Would love to get this in so that the onclickoutside dependency doesn't fall too far behind

valerybugakov:
Would be great to have this merged!

rafeememon:
Closing for lack of activity

plentz:
#536

PR #492, opened 2016-05-25
Use rem units for size to improve visibility on pages with a larger font

fritz-c:
The rem units I used were calculated such that the calendar would appear the same on a page with `html, body { font-size: 14px; }` as it does currently on a page with any font size.


martijnrusschen:
Thanks for changing this. Going to double check this to make sure it doesn't break our existing font sizes

Issue #498, opened 2016-05-29
Picked dates are timezone agnostic


lgrapenthin:
Currently datepicker (inadvertently?) gives one a date set to the beginning of day in the users current local timezone. I. e. if you live in GMT+2 the moment will be set to two hours before the beginning of day in UTC.

Currently the only way to avoid this is to initialize the object with a moment object that is e. g. set to UTC like passing Moment.utc(); as selected.

I need the user to pick a UTC date and to show him an empty field if he hasn't done, so the above workaround is not an option.

Would it be possible to add a utc flag to the options that forces react-datepicker to interpret and return dates only based on utc so that I can do timezone conversion outside if I wish to?

rafeememon:
Hey @lgrapenthin, it's very much intentional that the date is at the beginning of the day (since the component is time-agnostic it's the only thing that really makes sense).  If we were to add an option it would probably be better to specify an offset rather than a UTC flag ([docs](http://momentjs.com/docs/#/manipulating/utc-offset/)), but another possible workaround is to set UTC globally with `moment.utc()` ([docs](http://momentjs.com/docs/#/manipulating/utc/)) to make the empty case work.

lgrapenthin:
This is why the picker being timezone agnostic highly irritates me:

1. The component is supposed to let the user pick a calendar date, not a point in time. A calendar date is defined by Wikipedia as reference to a particular day within a Calendar system. 

2. Being able to pick a calendar date is the minimum requirement to any datepicker. Everything else, like adding daytime, casting timezones etc. can be done without a datepicker. If the datepicker also has such capabilities that is great, but it shouldn't make this minimum usecase anymore difficult. 

3. At the time react-datepicker makes this very difficult by forcing one to convert the picked date from the users local timezone back.

4. Whether the fact that users pick dates in different timezones is relevant to the _meaning of the date picked_ entirely depends upon on what kind of information that date is supposed to specify and as such should be users choice.

  Consider a birthday - Daytime has no meaning in this context. You don't know in what timezone the user was born, just the one in which he tells you about it. Yet you bake this into the default.

5. I. e. consider user1 has picked his m1 in USA, user2 has picked his m2 in GER, both via react-datepicker - Only knowing each users local timezone allows me to determine the calendar dates those users meant when they picked them in the UI.

rafeememon:
> This is why the picker being timezone agnostic highly irritates me:
>
[...]

PR #506, opened 2016-06-09
Support custom timezone

cristiango:
Expand the concept of **today**.
Today can be two different dates based on timezone.
In some situations you want to set the context timezone different than the client machine and by adding `timeZone: React.PropTypes.string` prop this can be achieved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
     "grunt-webpack": "^1.0.11",
     "highlight.js": "^9.0.0",
     "isparta-loader": "^2.0.0",
-    "json-loader": "^0.5.4",
     "karma": "^0.13.15",
     "karma-chai": "^0.1.0",
     "karma-chrome-launcher": "^1.0.1",
@@ -87,7 +86,6 @@
   "dependencies": {
     "classnames": "^2.2.1",
     "moment": "^2.13.0",
-    "moment-timezone": "^0.5.4",
     "react-onclickoutside": "^4.8.0",
     "tether": "^1.3.2"
   },
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "grunt-webpack": "^1.0.11",
     "highlight.js": "^9.0.0",
     "isparta-loader": "^2.0.0",
+    "json-loader": "^0.5.4",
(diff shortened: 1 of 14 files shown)


rafeememon:
Hey @cristiango, thanks for the contribution.  This seems okay but I'm wondering if this is the wrong solution to your problem; since you mention timezones I wonder if a fix to #498 should be implemented instead.  FWIW I generally prefer fixing the root of the problem rather than giving a workaround, and since this is an API addition I want to make sure it's the right thing to support in the long term.

cristiango:
@rafeememon I've opened that issues a couple of times an read an re-read it over. Not even sure if fixing the UTC problem will solve this problem

> You are in UTC+2 and want to select `today` in UTC-9.

The fact that the controls show information about `today` make is zone dependent

rafeememon:
Sorry, the relevant part of the other bug I meant to highlight was the idea of adding a timezone offset option (which is just a more general case of adding a UTC flag).  What do you think?

cristiango:
Yes indeed adding UTC flag will make more sense in the long run. 
I will try to come up with a solution for this

cristiango:
@rafeememon what do you think about the new solution?

martijnrusschen:
Ping @rafeememon

package.json
```diff
@@ -54,6 +54,7 @@
     "grunt-webpack": "^1.0.11",
     "highlight.js": "^9.0.0",
     "isparta-loader": "^2.0.0",
+    "json-loader": "^0.5.4",
```
rafeememon: I don't think the `json-loader` changes are necessary, are they?

package.json
```diff
@@ -86,6 +87,7 @@
   "dependencies": {
     "classnames": "^2.2.1",
     "moment": "^2.13.0",
+    "moment-timezone": "^0.5.4",
```
rafeememon: I would really prefer to not introduce a dependency on `moment-timezone`.  Instead of taking in the name of a timezone, we can take in an offset instead and do something like what is described here: http://momentjs.com/docs/#/parsing/parse-zone/

PR #508, opened 2016-06-15
Allow data-* attributes in date input

Lacerda:
I am using `react-datepicker` in a form where I need all the inputs to have `data-*` attributes.

With this PR, a DatePicker component created with the following:
```
<DatePicker dataAttributes={{'data-foo': 'bar'}} />
```
Creates the following input element:
```
<input type="text" class="datepicker__input" data-foo="bar">
```

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -17,6 +17,7 @@ var DatePicker = React.createClass({
   propTypes: {
     autoComplete: React.PropTypes.string,
     className: React.PropTypes.string,
+    dataAttributes: React.PropTypes.object,
     dateFormat: React.PropTypes.string,
     dateFormatCalendar: React.PropTypes.string,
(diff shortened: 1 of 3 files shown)


martijnrusschen:
Seems reasonable, @rafeememon what do you think?

deammer:
This would be wonderful to have :)

rafeememon:
Hey @Lacerda, sorry for the delayed response.  This seems like a good idea, but the way the code is currently structured, _any_ attribute can be passed into `dataAttributes` and appear on the component.  I think a better API to enforce `data-*` attributes would be

```
<DatePicker dataAttributes={{foo: 'bar'}} />
```

which turns into

```
<input type="text" class="datepicker__input" data-foo="bar">
```

rafeememon:
Closing due to inactivity

Issue #361, opened 2016-02-16
There's a problem when the popup's position is "top".


yy8597:
When clicking the button of increaseMonth and the height of the popup is change.  The popup of Calendar will be moved when clicking the button again. So I can't fire 'increaseMonth()'.

You can reproduce the bug by scrolling page when pupup is in bottom of page.

rafeememon:
Confirmed; I got the datepicker into a couple weird states
<img width="393" alt="screen shot 2016-02-16 at 10 23 34 am" src="https://cloud.githubusercontent.com/assets/1829450/13073579/794ceae2-d497-11e5-8ab8-3fedb7aa2ba4.png">

<img width="390" alt="screen shot 2016-02-16 at 10 23 51 am" src="https://cloud.githubusercontent.com/assets/1829450/13073574/73af67cc-d497-11e5-9edb-42dfc771988f.png">

taxigy:
I'm having the same problem. I put the `DatePicker` into a relatively positioned and constrained by width element, so the whole thing is

```jsx
<div style={{display: 'block', position: 'relative', width: '200px', zIndex: 3}}>
  <DatePicker
    className='form-control datepicker'
    dateFormat='MM/DD/YYYY'
    popoverAttachment='bottom right'
    popoverTargetAttachment='top right'
    selected={moment(category.published_at)}
    onChange={::this.setPublishDate} />
```

It behaves this way:

![react-datepicker-top-center-positioning-issue-gh-361](https://cloud.githubusercontent.com/assets/1625256/13088222/740ab554-d4fc-11e5-8aef-f8aa3a7188d7.gif)

crunchex:
Having this problem as well. I've been playing around with the tether properties, but haven't come up with a combination that serves as a workaround. If anyone finds one, please lay it down here.

sattaman:
+1 We're having the same issue

benwill:
+1 same

bekerov:
![2016-05-02_14-25-05](https://cloud.githubusercontent.com/assets/2405629/14953397/c0b0c4ba-1071-11e6-911f-08eb9e276f9e.gif)

Rambobafet:
+1

mb21:
[...]

PR #510, opened 2016-06-20
Fix top popup position issue

alextkachuk:
Remove filter isWeekInMonth in month.jsx for showing the same calendar size (6 rows) in different months. It is usual approach for date picker that it has the same size calendar. In this case, there is not the issue like #361 and size is predictable for users. 

Connect to #361

### DIFF
--- a/docs/datepicker.md
+++ b/docs/datepicker.md
@@ -6,6 +6,11 @@ General datepicker component.
 Props
 -----
 
+### `autoComplete`
+
+type: `string`
+
+
 ### `className`
 
 type: `string`
@@ -53,6 +58,11 @@ type: `string`
 type: `array`
 
 
+### `inline`
+
+type: `bool`
+
+
 ### `isClearable`
 
 type: `bool`
@@ -94,6 +104,11 @@ defaultValue: `function() {}`
 
 type: `func`
 defaultValue: `function() {}`
+
+
+### `openToDate`
+
+type: `object`
 
 
 ### `placeholderText`
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "author": "HackerOne",
   "name": "react-datepicker",
   "description": "A simple and reusable datepicker component for React",
-  "version": "0.27.0",
+  "version": "0.28.0",
   "license": "MIT",
   "homepage": "https://github.com/Hacker0x01/react-datepicker",
   "main": "dist/react-datepicker.min.js",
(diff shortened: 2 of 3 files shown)


benwill:
+1, think this is much better than how it currently works

rafeememon:
Hey @alextkachuk, thanks for the PR.  This is a pretty large departure from the current functionality and we need to be very careful when making this change.  I'm not sure that the argument that this is the usual approach is correct; some take our approach (e.g. [jQuery UI datepicker](https://jqueryui.com/datepicker/)) and others take the other (e.g. [bootstrap-datepicker](https://bootstrap-datepicker.readthedocs.io/en/latest/)).  I think at the very least that this should be behind a boolean flag option, and we most certainly need an answer for #296/#417 as well.  Let me know what you think.

rafeememon:
The popover position issue can almost certainly be directly fixed as well -- at a high level what probably needs to happen is that the calendar should be repositioned when the month changes

alextkachuk:
@rafeememon In my view, the best quick solution is setting opacity: 0.x (for example) for days are not in the current month and add my changes (fixed calendar).

Option for fixed or not is not good approach because in this case - not fixed mode will have a strong bug. For example, now I can't use react-datepicker in my current project because some modal window needs only top-view calendar and this bug my users found in production.

On the other hand, if popover position will fix in short period it would be great and in this case option for fixed/not fixed would be additional sugar :-)

rafeememon:
> In my view, the best quick solution is setting opacity: 0.x (for example) for days are not in the current month and add my changes (fixed calendar).

Potentially; this was once the behavior but it was changed because the dates looked too similar to disabled dates.  See the other issues I mentioned for more discussion about that

> Option for fixed or not is not good approach because in this case - not fixed mode will have a strong bug. For example, now I can't use react-datepicker in my current project because some modal window needs only top-view calendar and this bug my users found in production.

Yes, but the not-fixed mode is the current behavior, so it would be no worse than it is now.  Adding the extra rows without an option to turn them off is potentially undesired behavior for many consumers, including myself.

> On the other hand, if popover position will fix in short period it would be great and in this case option for fixed/not fixed would be additional sugar :-)

I definitely think that the ideal outcome here is that the position problem is solved, which would probably make this PR a moot point.  I'm not comfortable merging this yet because of the things we've been talking about, but feel free to take a stab at the position problem, maybe with the approach I mentioned before.

alextkachuk:
@rafeememon It makes sense, but now react-datepicker has the serious bug. The solution needs to change the tether library - it could be a long process. Can I add the option "isCalendarFixed" (or give me another name) and implement this optional logic? I think, anyway this option would be useful.

PR #515, opened 2016-07-01
Add a space between a number and 'pm' if needed

jaredsohn:
If a user types "4:00pm", make sure it parses as "4:00 pm" instead of "4:00 am".

Also, use an existing `moment` instead of passing in a date to remove some warnings.


jaredsohn:
Sorry, wrong repo.

Issue #517, opened 2016-07-02
React 15.2 Unkown Prop Warning


jambonrose:
We're running react-datepicker v0.27 and have just upgraded to React 15.2. We are now raising the following error for each instance of the date-picker.

```console
ERROR: 'Warning: Unknown props `date`, `locale`, `minDate`, `maxDate`, `excludeDates`, `includeDates`, `filterDate`, `dateFormat`, `onChangeDate` on <input> tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
```

The link at the end of the output, for convenience: https://fb.me/react-unknown-prop

The discussion occurring at erikras/redux-form#1249 may be of interest.

martijnrusschen:
Interesting! I haven't used 15.2 myself yet, so can't confirm the issue. Seems we're hitting this error because of reason 2:
> You are using a non-standard DOM attribute, perhaps to represent custom data. Custom data attributes should always have a data- prefix, as described in https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes

@rafeememon Can you confirm this is the case. If so, we should start renaming our props to make sure these are all valid.

spicyj:
As the gist says, you want to change DateInput to "consume" the props it uses instead of blindly spreading all of them onto the input tag.

https://github.com/Hacker0x01/react-datepicker/blob/8da6d8a6f2964fce814a19d594648c25f1e6c44e/src/date_input.jsx#L89

gaearon:
This is problematic because you’re forwarding props that only make sense to your component (e.g. `date`) to the `<input>` component that doesn’t know what to do with them. In future versions of React we plan to pass such props “through” without the whitelist, so we ask that you don’t forward props that don’t make sense to the underlying components anymore.

Here is the code today:

```js
    return <input
        ref='input'
        type='text'
        {...this.props}  <----------- can be anything! not good.
        value={this.state.value}
        onBlur={this.handleBlur}
        onChange={this.handleChange} />
```

you might have something like

```js
    const {
      date, dateFormat , excludeDates, filterDate, includeDates, locale,
      maxDate, minDate, onBlur, onChange, onChangeDate,
      ...rest
    } = this.props;
      
    return <input
        ref='input'
        type='text'
        {...rest} <------------------ we know for sure we aren't passing any DateInput's props
        value={this.state.value}
        onBlur={this.handleBlur}
        onChange={this.handleChange} />
```

[...]

PR #522, opened 2016-07-08
React "Unknown Prop Warning" fix

meggantodd:
https://facebook.github.io/react/warnings/unknown-prop.html

This PR removes unsupported props from the <input>.

### DIFF
--- a/src/date_input.jsx
+++ b/src/date_input.jsx
@@ -83,7 +83,7 @@ var DateInput = React.createClass({
   },
 
   render () {
-    const { date, locale, minDate, maxDate, excludeDates, includeDates, filterDate, dateFormat, onChangeDate, ...rest } = this.props
+    const { date, locale, minDate, maxDate, excludeDates, includeDates, filterDate, dateFormat, onChangeDate, ...rest } = this.props // eslint-disable-line no-unused-vars
     return <input
         ref='input'
         type='text'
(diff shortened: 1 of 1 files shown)


meggantodd:
Closing because of the eslint failure.

martijnrusschen:
Instead of closing the PR, could we try to fix the linter? Since this fixes #517, I'd like to get this trough :)

src/date_input.jsx
```diff
@@ -83,10 +83,11 @@ var DateInput = React.createClass({
   },
 
   render () {
+    const { date, locale, minDate, maxDate, excludeDates, includeDates, filterDate, dateFormat, onChangeDate, ...rest } = this.props
```
martijnrusschen:
```
/*eslint-disable */
const { date, locale, minDate, maxDate, excludeDates, includeDates, filterDate, dateFormat, onChangeDate, ...rest } = this.props
/*eslint-enable */
```

martijnrusschen: This should fix the build. We can revisit this once ES Lint has a better solution for this.

rafeememon: I'd prefer to be much more explicit about what is being disabled and applying it to the line in question rather than a block, just to prevent actual badness from sneaking in.  i.e.

```
... // eslint-disable-line no-unused-vars
```

martijnrusschen: That's even better :)

rafeememon:
LGTM, let's get this merged.  Thanks for the fix @meggantodd!

PR #524, opened 2016-07-11
Move tests enzyme

RSO:
I wasn't sure if this is actually something we want, but I figured I could check out what the hype is all about.

So far I think that enzyme works really well for 90% of the tests, however, things get complicated when I try to test the `TetherComponent`, as that component renders outside of itself. I'll try to figure out how to work around that, but so far I wanted to show what I have. What do you guys think?


martijnrusschen:
I like this, @rafeememon what do you think?

test/calendar_test.js
```diff
@@ -1,132 +1,132 @@
...
+    expect(todayButton.text()).to.equal('Vandaag')
   })
 
   describe('localization', function () {
     function testLocale (calendar, selected, locale) {
       var localized = selected.clone().locale(locale)
 
-      var calendarText = TestUtils.findRenderedDOMComponentWithClass(calendar, 'react-datepicker__current-month')
-      expect(calendarText.textContent).to.equal(localized.format(dateFormat))
+      var calendarText = calendar.find('.react-datepicker__current-month')
```
martijnrusschen: `.find` is a nice method.

test/date_input_test.js
```diff
@@ -1,53 +1,55 @@
...
   })
 
   it('adds autocomplete attribute to input field when autoComplete is passed as prop', function () {
-    var dateInput = TestUtils.renderIntoDocument(
+    var dateInput = shallow(
       <DateInput autoComplete='off' />
     )
 
-    expect(ReactDOM.findDOMNode(dateInput).autocomplete).to.equal('off')
+    expect(dateInput.find('input').prop('autoComplete')).to.equal('off')
```
martijnrusschen: `.prop`is fancy too

martijnrusschen:
We should probably update the readMe to make sure anyone knows that tests you can use Enzyme for tests. Here's some more information: https://github.com/airbnb/enzyme

martijnrusschen:
@rso Is this ready to go or do we need more changes before we land this?

Issue #525, opened 2016-07-15
openToDate doesn't work when minDate and maxDate are set


tylergould:
I'm trying to use react-datepicker with an openToDate on a project and I'm having trouble getting it to work when I have a minDate and maxDate set.

The calendar opens to the beginning of my min/max date range, when I would expect it to open to the openToDate.

I made this barebones project to reproduce the issue.
https://github.com/tylergould/reproduce-opentodate-issue

Thank you.

rafeememon:
Thanks for the bug report; resolved with #526

thangaduraicse:
@rafeememon - Kindly Let us know whether above issue is resolved in current release or if this issue is resolved only in next release then please let us know date for next release?

vetri50ster:
@rafeememon - we are using the same plugin in our project. we are also facing same defect.Please let us know solution for this defect.

martijnrusschen:
This should be fixed with the next release (should happen this week)

thangaduraicse:
@martijnrusschen: Client ask is to update that fix immediately. If you release the next version, this 'll be helpful for us.

rafeememon:
@thangaduraicse, you can easily build your own dist of the component in lieu of a release.

martijnrusschen:
Planning to do a release today. Stay tuned :)

martijnrusschen:
@thangaduraicse Here you go https://github.com/Hacker0x01/react-datepicker/releases/tag/v0.28.2

thangaduraicse:
@martijnrusschen @raffeememon Thank you.. :)

tsteele2013:
I'm having this issue also. Using v0.41.1

The following code still opens to today's date. Any ideas?

````
<DatePicker
  selected={moment(Application.ExpectedRetDate).isValid() ? moment(Application.ExpectedRetDate) : null}
  onChange={value => {
    this.setState({ touched: Object.assign(touched, { ExpectedRetDate: true })});
    Application.ExpectedRetDate = value ? value._d : null;
  }}
  openToDate={moment('1993-09-28')}
  onBlur={value => this.setState({ touched: Object.assign(touched, { ExpectedRetDate: true })})}
  minDate={Authentication.user.eligible_date}
  maxDate={moment().add(5, "years")}
  disabled={disabled}
  fixedHeight
  name="ExpectedRetDate"
  id="FormDatePicker"
  placeholderText="Select Date"
  className='form-control'
/>
````

[...]

PR #526, opened 2016-07-15
Make openToDate work with minDate and maxDate

tylergould:
See #525 

With this PR, when setting an openToDate along with a minDate and/or a maxDate, the calendar will initially open to the openToDate if it falls within the range.


rafeememon:
Lgtm, thanks for the fix!

PR #527, opened 2016-07-17
Add accessibility tags to month and day jsx

kaylieEB:
I was writing a ui test and realized that I can't click the div element containing a specific day text without using XPath. The unique aria-labels will change that while also helping screen readers better understand the elements and their roles as date options.

### DIFF
--- a/src/day.jsx
+++ b/src/day.jsx
@@ -66,7 +66,12 @@ var Day = React.createClass({
 
   render () {
     return (
-      <div className={this.getClassNames()} onClick={this.handleClick}>
+      <div
+        className={this.getClassNames()}
+        onClick={this.handleClick}
+        aria-label={`day-${this.props.day.date()}`}
+        role="option"
+      >
         {this.props.day.date()}
       </div>
     )
--- a/src/day.jsx
+++ b/src/day.jsx
@@ -71,7 +71,7 @@ var Day = React.createClass({
           onClick={this.handleClick}
           aria-label={`day-${this.props.day.date()}`}
           role="option">
-        {this.props.day.date()}
+          {this.props.day.date()}
       </div>
     )
   }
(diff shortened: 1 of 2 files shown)


src/day.jsx
```diff
@@ -66,8 +66,12 @@ var Day = React.createClass({
 
   render () {
     return (
-      <div className={this.getClassNames()} onClick={this.handleClick}>
-        {this.props.day.date()}
+      <div
+          className={this.getClassNames()}
+          onClick={this.handleClick}
+          aria-label={`day-${this.props.day.date()}`}
```
rafeememon: (I'm not super familiar with accessibility properties, so sorry if this is a dumb question...)  Is this required to be a unique label within the `listbox` role?  A month will show days from the previous and next month leading to some duplicate numbers, but not sure if that matters here.

kaylieEB: Thanks for asking! I actually originally thought it should be a label associated with a unique ID, but realized that there would be duplicates so went with `aria-label` instead, which provides more of a description. On the other hand, the roles are for semantic purposes, and elements with `option` roles are required to be within a `listbox` role to mark them as a group of available choices.

martijnrusschen: We could prefix the numbers with the month if that makes sense?

rafeememon: Thanks for the info @kaylieEB!  I think prefixing won't be necessary given that it's not a unique ID, but I'm curious how it affects a screen reader; can you give us a little insight into how a screen reader interprets this?  I don't think this will block the PR, but I'm just curious 😄

kaylieEB:
Sure, you can try using Mac VoiceOver tool or get the WAVE Chrome extension. It'll notify to the screen reader that they're inside a list box and read out the options, like below:

screen shot 2016-07-20 at 5 30 58 pm
screen shot 2016-07-20 at 5 30 29 pm

If we do add the month, it can get a bit nosy - I'm sure there can definitely be other improvements though!

PR #528, opened 2016-07-19
Add scrollable year dropdown option

lf-kiranpariyar:
Please review the code, comment and merge it as soon as possible.

### DIFF
--- a/docs-site/src/examples/year_dropdown.jsx
+++ b/docs-site/src/examples/year_dropdown.jsx
@@ -25,7 +25,8 @@ export default React.createClass({
               {"selected={this.state.startDate}"}<br />
               {"onChange={this.handleChange}"} <br />
               {"showYearDropdown"} <br />
-              {"dateFormatCalendar=\"MMMM\" />"}
+              {"dateFormatCalendar=\"MMMM\""} <br />
+              {"scrollableYearDropdown={true} />"}
         </code>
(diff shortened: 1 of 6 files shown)


docs-site/src/examples/year_dropdown.jsx
```diff
@@ -25,15 +25,17 @@ export default React.createClass({
...
         </code>
       </pre>
       <div className="column">
         <DatePicker
             selected={this.state.startDate}
             onChange={this.handleChange}
             showYearDropdown
-            dateFormatCalendar="MMMM" />
+            dateFormatCalendar="MMMM"
+            scrollableYearDropdown={true} />
```
martijnrusschen: Any reason why we should make this optional? UI wise, I think this should be scrollable by default and there's no need for toggling this currently. @rafeememon thoughts?

lf-kiranpariyar: Yes, you are right but i think scrollable year dropdown mainly needed in case of date of birth field otherwise default date works fine. That's why i made it scrollable.

TryingToImprove:
Is there any reason why this is not merged other than failing css lint? I am needing this for a datepicker for birthdates, and clicking the arrow is to slow..

martijnrusschen:
I created a new PR with the updated code. See #577

PR #531, opened 2016-07-23
Expose open state via prop

schickling:
Expose open state via prop

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -35,6 +35,7 @@ var DatePicker = React.createClass({
     onBlur: React.PropTypes.func,
     onChange: React.PropTypes.func.isRequired,
     onFocus: React.PropTypes.func,
+    open: React.PropTypes.bool,
     openToDate: React.PropTypes.object,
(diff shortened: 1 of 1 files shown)


rafeememon:
Hey @schickling, thanks for the PR.  I think this needs to be a little bit more -- this only covers the initial open state, but I think to fully address #283 the open prop needs to also change the open state after the initial mount.

The way I can see this working is that if the `open` prop is specified, then obey the prop, or else delegate to the original open state behavior.  I'm not sure if the callbacks we currently expose make it feasible.  Imagine you wanted to emulate the old behavior; you would probably want to attach callbacks to `onFocus`, `onBlur`, and potentially others.  However, we would probably need some special handling of these -- e.g. imagine the case where the picker is open and you click on the calendar; technically `onBlur` will fire, but we don't want to close the picker.

I've been putting off addressing #283 because of these complications, but let me know if you want to take a crack at it. :smile:

martijnrusschen:
Closing this out because of inactivity.

PR #534, opened 2016-08-02
Allow multiple formats for date input.

forum-is:
First, thanks for creating react-datepicker!

I would like to suggest allowing multiple formats for date entry. Our users often use the input box for manually typing or pasting dates. If the date provided does not __exactly__ match the specified dateFormat, this will fail. For example, if dateFormat is 'MM.DD.YYYY' and the user inputs '1.1.2016', the date will be rejected as invalid. 

As Moment.js already offers this functionality by simply providing an [array for formats](http://momentjs.com/guides/#/parsing/multiple-formats/). This requires only minimal changes (I hope I didn't miss anything).

If an array of formats is provided, the first format will be used to format the output but all other formats would be used to parse the input.

Would you please consider merging this PR? Thank you very much!

### DIFF
--- a/src/date_input.jsx
+++ b/src/date_input.jsx
@@ -7,7 +7,10 @@ var DateInput = React.createClass({
 
   propTypes: {
     date: React.PropTypes.object,
-    dateFormat: React.PropTypes.string,
+    dateFormat: React.PropTypes.oneOfType([
+      React.PropTypes.string,
+      React.PropTypes.array
+    ]),
     disabled: React.PropTypes.bool,
     excludeDates: React.PropTypes.array,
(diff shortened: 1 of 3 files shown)


martijnrusschen:
Thanks for submitting your PR! Seems like a nice addition to the dateFormat options. Would you mind adding some tests for the new functionality? This way we can make sure this will keep working in the future ;)

forum-is:
I've added a simple test. Somehow my first commit was missing some lines that my GitHub client appears to have swallowed...

jasonmullins:
Has this PR been published to npm?

martijnrusschen:
Not yet, I'll get to it today.

PR #536, opened 2016-08-04
Fix onClickOutside reference

FND:
TLDR: react-onclickoutside [exposes itself](https://github.com/Pomax/react-onclickoutside/blob/v5.3.3/index.js#L222) as `onClickOutside`, not `OnClickOutside`.

Having externalized React Date Picker (i.e. including `react-datepicker/dist/react-datepicker.js` via HTML rather than bundling it with webpack), I end up with the following warning:

> Warning: ReactClass: You're attempting to include a mixin that is either null or not an object. Check the mixins included by the component, as well as any mixins they include themselves. Expected object but got undefined

After some digging¹, I figured out that what's missing is [react-onclickoutside](https://github.com/Pomax/react-onclickoutside) (I'd already externalized React, ReactDOM and Moment.js anyway).

However, including (externalizing²) that as well didn't work. After more digging, it turns out to be due to this simply typo.

---

¹ These external dependencies don't seem to be documented as such, so I had to refer to `webpack.config.js`.

² Required to ensure proper load order: react-onclickoutside must be available upon react-datepicker initialization.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -46,7 +46,7 @@ module.exports = {
     },
     {
       'react-onclickoutside': {
-        root: 'OnClickOutside',
+        root: 'onClickOutside',
         commonjs2: 'react-onclickoutside',
         commonjs: 'react-onclickoutside',
         amd: 'react-onclickoutside'


FND:
Actually, this doesn't seem to suffice - injecting a temporary workaround for react-datepicker v0.28.2:

```html
<script src="react-onclickoutside.js"></script>
<script>
window.OnClickOutside = window.onClickOutside;
</script>
<script src="react-datepicker.js"></script>
<script src="bundle.js"></script>
```

results in

> Uncaught Invariant Violation: ReactClass: You're attempting to use a component class or function as a mixin. Instead, just use a regular object.

That error disappears if I use `onClickOutside.prototype` instead - but that wouldn't help with the webpack configuration being used here, so I'm not sure how to address this properly.

martijnrusschen:
Hi @FND, 

I'm updating this package in #601. This changes the mixin to a regular HOC since the implementation has been changed. What happens if you try this branch with your proposed fix?

jbdeboer:
FYI: This issue is preventing my company from adopting react-datepicker. :-<.

rafeememon:
@jbdeboer, feel free to pitch in on this PR to help out

martijnrusschen:
#651 was merged today. Do we still need this PR after we've upgraded the package?

jochenberger:
Yes, I think so. The global is still named `onClickOutside`.

Issue #529, opened 2016-07-20
Custom input component or element


gpoole:
What do you think about supporting customising the actual `<input />` by allowing the user to pass in their own? Something like:

```jsx
<DatePicker input={InputComponent} />
<DatePicker input={<InputComponent someCustomProp />} />
```

I think it's fairly common to have a base `Input` component which provides extra classes, styles or functionality, so it's useful to be able to use that here too. Obviously whatever's passed in would have to support the props already being passed to the regular `input`.

maxkrieger:
I was very surprised to see this not already available considering most people use this for the picker itself. It's pretty compromising for accessibility when you're stuck with a dummy textbox and can't use an `<input type="date">` or just a more custom solution/button.

gpoole:
Fixed in #551!

bORm:
<CustomInputComponent onClick={ this.props.onClick } .... 
add toggle Visibility calendar

PR #551, opened 2016-08-24
Add ability to pass a custom input

linpe:
Addresses https://github.com/Hacker0x01/react-datepicker/issues/529

Added a `customInput` prop which allows you to pass in your own component, allowing you to add custom functionality and styling.

### DIFF
--- a/docs-site/src/examples/custom_input.jsx
+++ b/docs-site/src/examples/custom_input.jsx
@@ -11,15 +11,10 @@ var ExampleCustomInput = React.createClass({
     value: React.PropTypes.string
   },
 
-  handleFocus () {
-    this.props.onFocus()
-  },
-
   render () {
     return (
       <button
           className="example-custom-input"
-          onFocus={this.handleFocus}
           onClick={this.props.onClick}>
           {this.props.value}
       </button>
--- a/docs-site/src/examples/custom_input.jsx
+++ b/docs-site/src/examples/custom_input.jsx
@@ -1,0 +1,84 @@
+import React from 'react'
+import DatePicker from 'react-datepicker'
+import moment from 'moment'
+
+/* eslint-disable react/no-multi-comp */
+var ExampleCustomInput = React.createClass({
+  displayName: 'ExampleCustomInput',
+
+  propTypes: {
+    onClick: React.PropTypes.func,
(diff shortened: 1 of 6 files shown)


martijnrusschen:
Thanks!

cowboy:
Now that this has been added, can you publish a new release? Thanks!!

i4got10:
Please add ``customInput`` prop to  [docs](https://github.com/Hacker0x01/react-datepicker/blob/master/docs/datepicker.md)

martijnrusschen:
I'll take a look tomorrow.

mtrabelsi:
**REALLY** coverage is 90.93%? you must be kidding : when using this custom input there is a lot of bugs that doesn't work : 

1. onBlur
2. Placeholder 
3. and much more

rafeememon:
@mtrabelsi, feel free to help out to fix those bugs

aij:
@rafeememon I'm fairly sure those are differences of opinions rather than actual bugs. Though FWIW, I found it confusing at first as well...

For example, placeholder is currently set per the `placeholderText` prop. If we were to "fix" that so that placeholder from customInput is used instead, then the users who expect `placeholderText` to be honored would be displeased.

Would it make sense to render customInput as a component instead of cloning an element? That would give users full control over the rendering of custom inputs. (Though we may want to still support cloning for compatibility.)

rafeememon:
@aij, these very well may be things that should be changed, but it was more commentary on the attitude of the comment, not the content =)

docs-site/src/examples/custom_input.jsx
```diff
@@ -0,0 +1,84 @@
...
+          <br />
+          {"<DatePicker"}<br />
+              {"customInput={<ExampleCustomInput />}"}<br />
+              {"selected={this.state.startDate}"}<br />
+              {"onChange={this.handleChange} />"}
+        </code>
+      </pre>
+      <div className="column">
+        <DatePicker
+            customInput={<ExampleCustomInput />}
```
prakashvgr: C(CustomInput) must be caps customInput={<ExampleCustomInput />}

martijnrusschen: Feel free to update the example and submit a PR

Issue #488, opened 2016-05-23
Datepicker inside a modal


reyhansofian:
Hi, I have a datepicker inside a modal. Whenever I clicked on the date, the modal got closed too. Is there a way to avoid this? My code looks like this

```
import React from 'react';
import moment from 'moment';
import { change } from 'redux-form';
import DatePicker from 'react-datepicker';

class ReservationDetailsCheckinContainer extends React.Component {
  constructor(props) {
    super(props);
  }

  handleInputChange(e) {
    console.log(e);
  }

  render() {
    return (
      <div className="pull-left" style={{
        width: '65%'
      }}>
        <label>
          Check-in Date
        </label>
        <div>
          <DatePicker selected={this.state.startDate} onChange={this.handleInputChange}/>
        </div>
      </div>
    );
  }
}

export default ReservationDetailsCheckinContainer;
```

TMUND:
Have you tried e.stopPropagation()?

mezod:
Did you make anything special to make it work inside a modal? It doesn't work for me

leojh:
Anyone find a solution to this?

mayrsascha:
The problem here is that the onChange callback only provides the newly selected date, but not the event associated with it. If we would get the event as lets say second parameter the following code should work:

`handleInputChange(date, e) {  
    e.persist();  
    e.stopPropagation();  
  }`

It would work because AFAIK most modals are adding a click listener to window and then check if the click target (in our case the datepicker calendar) is the modal itself or a children of it.

BUT
React datepicker is using the tether library to connect the date input field with the calendar component and tether is placing the calendar outside the modal before the closing body tag for positioning reasons

<img width="585" alt="bildschirmfoto 2016-07-14 um 10 44 16" src="https://cloud.githubusercontent.com/assets/5527658/16834136/c478e616-49b2-11e6-8491-6d8c01ef57f8.png">

So technically the calendar is not inside your modal and thats why it closes. As I wrote already, exposing the React event in the onChange callback would probably solve this, since then you can stop the event from reaching the window onclick listener

reyhansofian:
Yeah, you're right. Thanks for your answer

Reyhan Sofian
Web Developer

LeoZeda:
Is there a way to get the event?

kiddo13:
+1

shanSyapse:
+1

mateder:
+1

amakhrov:
[...]

PR #558, opened 2016-08-31
Propagate datepicker events up the chain

averted:
This should provide `Event` obj to `onChange` etc.. (Related to #488).

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -94,8 +94,8 @@ var Calendar = React.createClass({
     })
   },
 
-  handleDayClick (day) {
-    this.props.onSelect(day)
+  handleDayClick (day, ev) {
+    this.props.onSelect(day, ev)
   },
 
   changeYear (year) {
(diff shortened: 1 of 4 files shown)


src/calendar.jsx
```diff
@@ -94,8 +94,8 @@ var Calendar = React.createClass({
     })
   },
 
-  handleDayClick (day) {
-    this.props.onSelect(day)
+  handleDayClick (day, ev) {
```
martijnrusschen: I guess `ev` is the event here? Would you mind changing that to `event`?

averted: Changed `ev` everywhere in PR into `event`.

kiddo13:
when this pull request will be merged?

kiddo13:
I could not get this update from npm install.

iknowcss:
I believe the change won't be available until a new version is published to npm (maybe v0.30.0)

martijnrusschen:
We'll release a new version soon.

PR #572, opened 2016-10-05
added transform-es2015-destructuring to .babelrc

github1:
After upgrading to React 15.3.2 I started seeing a bunch of 'Unknown props' warnings (shown below).  Apparently babel does not handle 'rest' destructuring unless the transform-es2015-destructuring plugin is enabled.  Adding this transform ensures the 'Unknown props' shown below are excluded from the props set on the 'input' element (src/date_input.jsx:90).

warning.js:36Warning: Unknown props `date`, `locale`, `minDate`, `maxDate`, `excludeDates`, `includeDates`, `filterDate`, `dateFormat`, `onChangeDate` on <input> tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -9,7 +9,8 @@
             "imports": ["react"],
             "locals": ["module"]
           }]
-        }]
+        }, 
+        "transform-es2015-destructuring"]
       ]
     }
   }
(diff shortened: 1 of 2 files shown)


rafeememon:
Hey @github1, the destructuring plugin should already be included since we use the [ES2015 preset](https://babeljs.io/docs/plugins/preset-es2015/).  There were some fairly recent fixes regarding this error; what version of the datepicker are you using?

github1:
Ah yea I was on 0.27.0.  It is working in at least 0.29.0+. Cool.

PR #573, opened 2016-10-05
Add autoFocus prop

### LINKED ISSUE #276: How can I set focus on the input element from the outside?
I'd like to autofocus the input element and show the calendar once the DatePicker is rendered, in order to allow the user to select a date w/o forcing him to click on the input field first. Is there a way?

MarkReeder:
Fixes #276

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -93,12 +93,6 @@ var DatePicker = React.createClass({
     }
   },
 
-  componentDidMount () {
-    if (this.props.autoFocus) {
-      this.setFocus()
-    }
-  },
-
   setFocus () {
     this.refs.input.focus()
   },
@@ -235,6 +229,7 @@ var DatePicker = React.createClass({
         ref='input'
         id={this.props.id}
         name={this.props.name}
+        autoFocus={this.props.autoFocus}
         date={this.props.selected}
         locale={this.props.locale}
         minDate={this.props.minDate}
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -93,10 +93,6 @@ var DatePicker = React.createClass({
     }
   },
 
-  setFocus () {
-    this.refs.input.focus()
-  },
-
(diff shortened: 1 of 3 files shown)


src/datepicker.jsx
```diff
@@ -92,6 +93,16 @@ var DatePicker = React.createClass({
     }
   },
 
+  componentDidMount () {
+    if (this.props.autoFocus) {
```
rafeememon: Did you try simply passing the `autoFocus` prop to the `input` element?  It would be nice to delegate this to native HTML if possible.

test/datepicker_test.js
```diff
@@ -298,4 +298,13 @@ describe('DatePicker', () => {
     var result = data.callback.args[0][0]
     expect(result.format(data.testFormat)).to.equal(data.copyM.format(data.testFormat))
   })
+  it('should autofocus the input given the autoFocus prop', (done) => {
+    var div = document.createElement('div')
+    document.body.appendChild(div)
+    ReactDOM.render(<DatePicker autoFocus />, div)
+    defer(() => {
```
rafeememon: Out of curiosity, can this `defer` go away?  Not sure if focusing happens sync or async

MarkReeder: Good call- the test failed previously, but using the autoFocus prop, it's not necessary. I'll remove.

src/datepicker.jsx
```diff
@@ -92,6 +93,10 @@ var DatePicker = React.createClass({
     }
   },
 
+  setFocus () {
```
rafeememon: (sorry for the back-and-forth, hard to give a complete CR on the train =P)  It looks like this part of the change can probably go away

rafeememon: Agreed that it might be useful, but that should be addressed in another PR to separate concerns =)

rafeememon:
lgtm, thanks!

PR #519, opened 2016-07-07
Add option for fixed size of the calendar

alextkachuk:
Lets to set fixed size of the calendar (6 rows of weeks). In addition, lets to prevent issue with top position of the calendar.

Without fixed size:
react-datepicker-before

Fixed size:
react-datepicker-fixed

PR #574, opened 2016-10-05
Fix positioning of calendar when month is changed

umurgdk:
For some months there is an extra row of days. Tether component calls
position method of tether instance but at the moment that method called
react doesn't update the real dom. As well as fixing this asynchronous
problem, there is an another bug in tether library, which can not
recalculate the position for the first call of `position()` method. I
observed if you call that method many times (between 1 - 10) it finally
recalculate correctly.

PS: I'm aware of how `setTimout` and `for loop` looks ugly. If you have a better idea i would like to hear.

### DIFF
--- a/src/tether_component.jsx
+++ b/src/tether_component.jsx
@@ -74,7 +74,7 @@ var TetherComponent = React.createClass({
   },
 
   position () {
-    this._tether.position()
+    this._updatePosition()
   },
 
   _destroy () {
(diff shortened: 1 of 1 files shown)


src/tether_component.jsx
```diff
@@ -124,20 +124,27 @@ var TetherComponent = React.createClass({
...
+        this._tether.setOptions(tetherOptions)
+      }
+
+      this._updatePosition()
+    }, 100)
+  },
 
-    this._tether.position()
+  _updatePosition () {
+    for (var i = 0; i < 10; i++)
```
rafeememon: Unfortunately this seems like too much of a hack to include, and it may incur a performance hit as I suspect `position()` is not cheap to call.

rafeememon:
Hey @umurgdk -- thanks for the fix, but unfortunately I don't think it's a good idea to include this.  If this is a problem with upstream (i.e. `tether` or `react-tether`) then the bug should be fixed there instead.  If you're looking for a workaround, take a look at #519 and https://github.com/Hacker0x01/react-datepicker/issues/361#issuecomment-218100483

umurgdk:
@rafeememon you're right ^_^ thanks for pointing me the fixedHeight that works for us. latest version in npm registry does includes that feature?

rafeememon:
Yep!

PR #577, opened 2016-10-06
Scrollable year dropdown

martijnrusschen:


src/year_dropdown_options.jsx
```diff
@@ -83,8 +85,13 @@ var YearDropdownOptions = React.createClass({
   },
 
   render () {
+    let dropdownClass = classNames({
+      'react-datepicker__year-dropdown': true,
+      'scrollable-year-dropdown': this.props.scrollableYearDropdown
```
rafeememon: To follow proper BEM conventions can we change this class name to `react-datepicker__year-dropdown--scrollable`?

PR #578, opened 2016-10-06
Correct proptypes to reflect updated docs

martijnrusschen:
In 441edf887e76f51e631527d6c03c45bda673c046 we manually updated the docs, this won’t hold if we’re releasing new docs. I updated the code that’s used to generate docs. Please make sure this is correct.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -25,7 +25,7 @@ var Calendar = React.createClass({
     dateFormat: React.PropTypes.oneOfType([
       React.PropTypes.string,
       React.PropTypes.array
-    ]),
+    ]).isRequired,
     dropdownMode: React.PropTypes.oneOf(['scroll', 'select']).isRequired,
     endDate: React.PropTypes.object,
     excludeDates: React.PropTypes.array,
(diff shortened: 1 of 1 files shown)


martijnrusschen:
@grumd @rafeememon Can you verify this is correct?

grumd:
Seems like this is correct, see this file for reference https://github.com/Hacker0x01/react-datepicker/blob/master/src/date_input.jsx

src/calendar.jsx
```diff
@@ -56,6 +59,7 @@ var Calendar = React.createClass({
 
   getDefaultProps () {
     return {
+      dateFormat: 'L',
```
rafeememon: This is passed in from the `DatePicker` component, which has a default of `'MMMM YYYY'`.  We should only define this in one place

src/calendar.jsx
```diff
@@ -22,7 +22,10 @@ var Calendar = React.createClass({
   displayName: 'Calendar',
 
   propTypes: {
-    dateFormat: React.PropTypes.string.isRequired,
+    dateFormat: React.PropTypes.oneOfType([
+      React.PropTypes.string,
+      React.PropTypes.array
+    ]),
```
rafeememon: `.isRequired`

PR #582, opened 2016-10-08
Prevent browsers autocomplete suggestions.

silvajonathan:
Using react-datepicker i noticed that the browser still show autocomplete suggestions based on previous values used on the field. This suggestion is a dropdown and overlaps the calendar.


micabe:
autoComplete="off"

Issue #585, opened 2016-10-11
openToDate prop not triggering change after getInitialState()


arnireynir:
First I would like to say thanks for creating this, this is so awesome package and super friendly, easy to use and to build around. 

I have small problem that I would like to solve with this package and that is to show two calendars side-by-side like is now possible but the other calendar is always showing next month ( like booking on Airbnb ).

I would like to submit a PR that fixes problem.
The prop "openToDate" is only triggered on getInitialState() therefore when changing date the second calendar that should be showing one month ahead won't keep up.

arnireynir:
Also it would be nice if this will be merged that you can label it "Hacktoberfest" :)

PR #586, opened 2016-10-11
Fix for openToDate bug

arnireynir:
Fix for #585

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -73,6 +73,11 @@ var Calendar = React.createClass({
         date: this.localizeMoment(nextProps.selected)
       })
     }
+    if (nextProps.openToDate && nextProps.openToDate !== this.props.openToDate) {
+      this.setState({
+        date: this.localizeMoment(nextProps.openToDate)
+      })
+    }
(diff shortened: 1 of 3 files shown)


src/calendar.jsx
```diff
@@ -73,6 +73,11 @@ var Calendar = React.createClass({
         date: this.localizeMoment(nextProps.selected)
       })
     }
+    if (nextProps.openToDate && nextProps.openToDate !== this.props.openToDate) {
```
rafeememon: * Can you change this to an `else if`?  It makes it more clear what should happen if both `selected` and `openToDate` change, where `selected` should take precedence to be consistent with `getDateInView`\
* You should use the `isSameDay` helper to avoid some unnecessary state updates

arnireynir: Yes, no problem will make the changes.

rafeememon:
Looks good, thanks @arnireynir!

arnireynir:
@rafeememon fantastic, do you know when next npm release will be made?
also could you add "Hacktoberfest" label for me :)

Issue #590, opened 2016-10-14
Comply With Event Spec On Date Input Change Handler 


kintendo:
I currently get an error because my Custom Input is not passing a jQuery event. Could you change the event handler to something like the following?
```
handleChange (event) {
    if (this.props.onChange) {
      this.props.onChange(event)
    }
    if (event.defaultPrevented === false || (typeof event.isDefaultPrevented === 'function' && !event.isDefaultPrevented())) {
      this.handleChangeDate(event.target.value)
    }
  }
```

rafeememon:
Seems reasonable to just change it to `if (!event.defaultPrevented) { ...` since it complies with both  the [react](https://facebook.github.io/react/docs/events.html) and [normal web](https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented) specs; PRs welcome

kintendo:
Ah, I left `isDefaultPrevented` in there because it's possible with custom inputs that `defaultPrevented` could be undefined. According to [jQuery docs](https://api.jquery.com/category/events/event-object/) and [W3 docs](https://www.w3.org/TR/2003/WD-DOM-Level-3-Events-20030331/ecma-script-binding.html) it doesn't look like `defaultPrevented` is a property on the event object.

rafeememon:
[These W3 docs](http://www.w3schools.com/jsref/event_defaultprevented.asp) show that `defaultPrevented` is valid.  Also, from the jQuery docs:

```
jQuery's event system normalizes the event object according to W3C standards. The event object is guaranteed to be passed to the event handler. Most properties from the original event are copied over and normalized to the new event object.
```

which I interpret to mean that `defaultPrevented` is copied over.  I'm also dubious of building around the jQuery API, as it's a separate ecosystem that generally is not relevant for React apps.

kintendo:
Sounds good, I'll make the change.

rafeememon:
FIxed by #591

PR #591, opened 2016-10-14
support other specs for preventDefault

kintendo:
With custom inputs, sometimes you could get a DOM event vs. a JS event. This change accommodates both.


rafeememon:
Please change `defaultPrevented` as discussed on https://github.com/Hacker0x01/react-datepicker/issues/590

rafeememon:
Thanks!

PR #596, opened 2016-10-19
Jkyro/multi month calendar

jkyro:
Initial support for showing multiple months in calendar via 'months' option. If this is combined with dropdowns they are shown only on the first month. Includes some examples and tests.


docs/datepicker.md
```diff
@@ -195,3 +195,6 @@ type: `string`
 
 type: `string`
 
+### `months`
```
martijnrusschen: You can ignore this, this document will be auto generated on every release.

src/calendar.jsx
```diff
@@ -56,7 +57,8 @@ var Calendar = React.createClass({
 
   getDefaultProps () {
     return {
-      utcOffset: moment.utc().utcOffset()
+      utcOffset: moment.utc().utcOffset(),
+      months: 1
```
martijnrusschen: What about something more descriptive, like: `monthsShown`. Not the best naming but open for suggestions.

src/datepicker.jsx
```diff
@@ -83,7 +84,8 @@ var DatePicker = React.createClass({
           attachment: 'together'
         }
       ],
-      utcOffset: moment.utc().utcOffset()
+      utcOffset: moment.utc().utcOffset(),
+      months: 1
```
martijnrusschen: Same here.

src/datepicker.jsx
```diff
@@ -218,6 +220,7 @@ var DatePicker = React.createClass({
         utcOffset={this.props.utcOffset}
         outsideClickIgnoreClass={outsideClickIgnoreClass}
         fixedHeight={this.props.fixedHeight}
+        months={this.props.months}
```
martijnrusschen: And here.

jkyro:
Thanks for the comments! I think the build fails because the added code lowers the coverage percent somewhat. Additional test cases should do the trick. As for the parameter naming, monthsShown sounds good to me, I'll fix that.

martijnrusschen:
Thanks for adding this!

PR #600, opened 2016-10-24
Fix month arrow visibility issue

MarkReeder:
When manually testing my other pull request, I noticed that the month navigation arrows were missing. I added a z-index to the react-datepicker__navigation class to fix.


martijnrusschen:
Thanks!

PR #601, opened 2016-10-24
Update react-onclickoutside

martijnrusschen:


martijnrusschen:
This seems the way to go. However, 2 tests where this method is used are failing because we moved from a mixin to HOC. Anyone wants to look in this? cc @rafeememon

magsout:
Something like 
```js
module.exports = onClickOutside(Calendar) 
```

No ?

martijnrusschen:
@magsout That could be, but it doesn't magically fixed the tests. If you have time looking into this, that would be appreciated.

plentz:
#536

jochenberger:
I think that the tests fail because they try to access components' internal state. Now those components are wrapped inside a HOC so the state is "one level deeper".

martijnrusschen:
@jochenberger Correct, I'd love to get some help on this. I don't have time to fix this in the next few weeks.

jochenberger:
See #651.

jochenberger:
Even though I created #651, I'll comment here anyway, so you can decide which approach you like better.
First of all, you'll need to change `calendar_test.js` to use `mount` instead of `shallow` inside `getCalendar()`, line 14. Then you need to replace `calendar.state()` by `calendar.instance().getInstance().state`. That fixes the first batch of tests.
Things get more complicated on `year_dropdown_options_test.js` since it tries to simulate events on refs. We cannot work with `instance().getInstance()` there because we're no longer in enzyme space then so we cannot use `simulate`.
I can try to dig in deeper, but I'd like to know whether you want to take this approach or #651.

martijnrusschen:
Code wise, I think this implementation is a bit cleaner over the one you created in #651, but since your implementation saves a lot of issues with fixing all the tests, I think it's worth the tradeoff here since it's still understandable and doesn't look that bad.

martijnrusschen:
Going with #651

PR #614, opened 2016-11-02
Fix prop name in examples from months to monthsShown

OAGr:


OAGr:
Previously there was an issue with the documentation, resulting in the 'monthsShown' examples to now work. I believe this was the intended functionality.

https://github.com/Hacker0x01/react-datepicker/pull/600

image

PR #617, opened 2016-11-06
Migrate to Yarn

martijnrusschen:
I want to replace NPM with Yarn in this repo. First step is adding a yarn.lock file.

I ran `yarn install` and this is the result.


avindra:
Any feedback for this?


Not sure if Yarn.lock should be there but the rest looks good

martijnrusschen:
We can probably simplify the Travis setup a bit since it supports Yarn out of the box now: https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn

ehkasper:
@avindra as long as I know it's a good idea to commit .lock files so installation is deterministic. See gemfile.lock in ruby, composer.lock in PHP, etc.

avindra:
IMO, it should only be done if absolutely necessary. Lock files are great for apps (you want deterministic builds there), but I can't think of a case where this makes sense for a library. Otherwise, libraries are dictating to apps exactly what versions of their dependencies should be bundled, which goes too far, imho

ehkasper:
@avindra great point! I havent thought of it this way. Makes sense, but i need to read further. Many thanks on your response :)

martijnrusschen:
Having a lock file in your repository shouldn't hurt since it's only used for local development AFAIK. Even projects like https://github.com/facebook/react are using lock files.

martijnrusschen:
Also, according to the Travis docs, you'll need a lock.file so that Travis can detect that you're using Yarn.

ehkasper:
Just to add info on top of this: https://yarnpkg.com/en/docs/yarn-lock#toc-check-into-source-control. Yarn only checks the top level, so libs should include it.

Issue #487, opened 2016-05-20
Feature request: Callback for month change or year change


alexstroukov:
When using the datepicker in a scenario where you need to request available dates asynchronously its often inefficient to request all possible dates at once.

It would be beneficial to be able to hook into year and month changes to know when to make requests for the next set of dates

PR #626, opened 2016-11-15
Callback for month change, increase, decrease or year change

lbrdar:
Added 4 callbacks called:
- onMonthIncrease (newDate)
- onMonthDecrease (newDate)
- onMonthChange (newDate)
- onYearChange (newDate)

Where newDate is a moment object representing the new date.

Pretty sure this solves issue #487


src/calendar.jsx
```diff
@@ -127,13 +135,13 @@ var Calendar = React.createClass({
   increaseMonth () {
     this.setState({
       date: this.state.date.clone().add(1, 'month')
-    })
+    }, this.props.onMonthDecrease(this.state.date.month()) )
```
magsout: 1) it's dangerous to use this.state, because `setState` is asynchronous, so the result is not the same as above, better to use const with one result
2) why did you passed func in setState ?

lbrdar: You are completely right, that is exactly why I changed this in later commit. setState allows callback as second argument, which ensures to return new state as this.state.
So  () => this.props.onMonthIncrease(this.state.date) will call onMonthIncrease function (that user sent as prop) with new date as argument, allowing user to do whatever he wants with this info.

intpp:
When planned to push it in the release?

alex-shamshurin:
Could you resolve conflicts and squash the commits into one? Do not include `docs-site/bundle.js` and `docs-site/style.css` into commit please to avoid conflicts.

PR #629, opened 2016-11-17
Calendar pop-up window does not close if using Macbook touchpad [fix #628]

### LINKED ISSUE #628: Calendar pop-up window does not close if using Macbook touchpad
Yes, it sounds strange but this is how it is.
I have Macbook Pro 2011, and when i click on some date(day number), window doesn't close(it closes for a millisecond, then it appears again).

I recorded a video.
https://monosnap.com/file/iC6HuXQjChtF0QVNGaFAQH5Xfz7bxv

It only happened when i touched the touchpad. If i 'clicked' it or used a mouse everything was fine.
I wonder that there were not such issue. I tested it on my friend's Macbook Pro 2014 and there was the same issue.

I figured out that it happened because onFocus event ran after onSelect. I fixed it by prohibiting onFocus event for 40 milliseconds after onSelect event. It fixed this problem, and i don't think it caused anything else to break.

ignatif:
https://github.com/Hacker0x01/react-datepicker/issues/628

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -93,7 +93,8 @@ var DatePicker = React.createClass({
 
   getInitialState () {
     return {
-      open: false
+      open: false,
+      preventFocus: false
     }
   },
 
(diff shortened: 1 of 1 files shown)


Kadrian:
Looking good to me! Is there anything else to be done / can I help somehow?

ignatif:
@Kadrian 
Well, I guess the solution i've provided isn't most accurate one.
It looks more like a crutch.

If you can figure out how to fix it more "pretty" - you are welcome :)

PR #608, opened 2016-10-27
Save time for selected date

zouxuoz:
resolve #606

PR #633, opened 2016-11-23
fix regression when clearing input field

fxfilmxf:
A regression was introduced in this pull request #608

DateInput calls it's `onChange` prop with a `null` date if the input value is and empty string (https://github.com/Hacker0x01/react-datepicker/blob/master/src/date_input.jsx#L64). That value is then parsed by moment into an 'Invalid Date' in the DatePicker (https://github.com/Hacker0x01/react-datepicker/blob/master/src/datepicker.jsx#L149) and then passed as a parameter to it's `onChange` prop.

This pull request fixes the regression and passes `null` instead. This is the same behavior that DatePicker had before #608

### DIFF
--- a/test/datepicker_test.js
+++ b/test/datepicker_test.js
@@ -329,4 +329,24 @@ describe('DatePicker', () => {
     datePicker.setFocus()
     expect(div.querySelector('input')).to.equal(document.activeElement)
   })
+
+  it('should correctly clear date with empty input string', () => {
+    var cleared = false
+    function handleChange (d) {
+      // Internally DateInput calls it's onChange prop with null
+      // when the input value is an empty string
+      if (d === null) {
+        cleared = true
+      }
+    }
+    var datePicker = TestUtils.renderIntoDocument(
+      <DatePicker
+          selected={moment('2016-11-22')}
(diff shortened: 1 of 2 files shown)


test/datepicker_test.js
```diff
@@ -329,4 +329,24 @@ describe('DatePicker', () => {
     datePicker.setFocus()
     expect(div.querySelector('input')).to.equal(document.activeElement)
   })
+
+  it('should correctly clear date with empty input string', () => {
```
fxfilmxf: I verified that this test fails without the fix.

test/datepicker_test.js
```diff
@@ -329,4 +329,24 @@ describe('DatePicker', () => {
...
+    var cleared = false
+    function handleChange (d) {
+      // Internally DateInput calls it's onChange prop with null
```
fxfilmxf: I don't know if this comment is helpful or not.

martijnrusschen:
Thank you!

test/datepicker_test.js
martijnrusschen: TDD :)

PR #649, opened 2016-12-01
emit errors for unused prop types

jochenberger:
Since the PropTypes are the only documentation for a lot of features, they should be correct. Otherwise, they only cause confusion.


martijnrusschen:
Awesome!

martijnrusschen:
Fully agree btw!

martijnrusschen:
could you fix the build so we can land this?

jochenberger:
done

PR #651, opened 2016-12-02
Upgrade react-onclickoutside

jochenberger:
I moved the wrapping code outside of the actual class implementations to be able to unit-test those.


jochenberger:
I'm not entirely sure that this is the way it should be done, one could argue that the onClickOutside behavior belongs to the component and thus should be contained in the same file. That makes unit testing a bit more difficult. See also https://github.com/Pomax/react-onclickoutside/issues/133

Issue #646, opened 2016-12-01
Layout is broken for showYearDropdown=false showMonthDropdown=true


jochenberger:
`.react-datepicker__header__dropdown react-datepicker__header__dropdown--select` has a margin of `-16px` (https://github.com/Hacker0x01/react-datepicker/commit/95a870ac91b3cc76c3f9f3fca32975c2a2825e22#diff-9da5657b359639ae904fe748a1fb2b4cR43). This was probably added to "remove" the additional margin that is added when using a year dropdown (https://github.com/Hacker0x01/react-datepicker/blame/95a870ac91b3cc76c3f9f3fca32975c2a2825e22/src/stylesheets/datepicker.scss#L59). But if the date picker does not have a year dropdown, there is no additional margin to remove and the month picker ends up 16px too high.

jochenberger:
See https://codepen.io/anon/pen/GNyJPN

jochenberger:
It also breaks if you only specify `dropdownMode="select"`: https://codepen.io/jochenberger/pen/jVYbwX

martijnrusschen:
Interesting, would you be able to fix this and create a PR?

jochenberger:
Actually, I'm not sure how. It looks pretty complicated to me the way it is. Absolute positioning combined with magic-number margins. I think we should get rid of those altogether but since there are so many possible combinations of the dropdown options, I'm not sure how to fix it without breaking something else.

Issue #533, opened 2016-07-29
the year should not disappear when showYearDropdown is false.


solsos:
When I set the showYearDropdown is false. 
We don't want the user select the year. 

But base on current function. 
1. the first show is right, we the month and year in title.
![before click the year](https://cloud.githubusercontent.com/assets/17923026/17262450/6493056c-55a9-11e6-8b1c-deb2f62992a7.png)
2. after click, the year is disappear and the year drop down list is in display none. 
![after click the year](https://cloud.githubusercontent.com/assets/17923026/17262449/6491eec0-55a9-11e6-9348-f8e717d35c9b.png)
So, that not a good user experience..

jochenberger:
I cannot reproduce this with 0.34.0: https://codepen.io/jochenberger/pen/xRpQOK

PR #657, opened 2016-12-02
add month dropdowns

jochenberger:
The styles are horribly broken, but apart from that, it works.
Oh, and it has no tests...


jochenberger:
It probably makes sense to review #646 first and revisit this afterwards.

jochenberger:
@martijnrusschen this should be ready for review

martijnrusschen:
Would you mind adding an example to the docs on how this works? Similar to the year dropdown.

jochenberger:
Done. But I noticed that the year is not dsplayed in the header if `showMonthDropdown ` is set. This is probably a bug. Might be related to #533.

martijnrusschen:
Awesome, thanks for the help!

PR #671, opened 2016-12-13
After utcOffset property change, update date input value

### LINKED ISSUE #670: Date input value does not match selected date in calendar after changing utcOffset property
The date input doesn't get updated when the utcOffset property changes and the selected date remains set to the same date/time.  This appears to be an issue because the componentWillReceiveProps function checks that the new and current date are the same day.  Two dates can be exactly the same day, but have different dates displayed in YYYY-MM-DD format given different timezone offsets.

Ex: "2016-12-06T00:00:00Z" is the same as "2016-12-05T18:00:00-06:00".  The date input should display 2016-12-06 in one case, 2016-12-05 for the other.

In my use case, the calendar component does update correctly.  I'm not sure why, as it also checks for the same day in the componentWillReceiveProps function, but perhaps something else changes that triggers the re-render.  The date input does not update.

I'll see if I can figure out a fix and put in a pull request, but would appreciate confirmation that this is indeed an error, replicated by others.

amityb:
Allow a change to the utcOffset property to update the date input value even when the selected date has not changed.  A change to the utcOffset may change the actual representation of the same date, especially when time is included in the date format.  This change expects the utcOffset to be applied to the new date coming in as a property to the date-input component.
Fixes #670


jochenberger:
The UTC offsets are not constant, e.g. "Europe/Berlin" has +01:00 or +02:00 depending on DST (https://en.wikipedia.org/wiki/Daylight_saving_time)

amityb:
Yes, UTC offsets are not constant for a given timezone.  I'm not exactly clear on how that applies to this commit.  Can you elaborate?

jochenberger:
Sorry, i was too quick with my response. I thought you had added a timezone picker to the component and used constant offsets for that.
I suggest changing the option labels to not include locations. It might lead to issues for other users who copy your code without having DST in mind.

amityb:
Removed locations from options labels in timezone example

amityb:
As far as I can tell, codecov/patch check is failing because of a change to an import line.  I am unfamiliar as how to cover this.  I have written tests to cover the new object that was imported.  Can anyone explain how I'm supposed to improve coverage on an import line in order to get codecov/patch check to pass?  I don't know how to cover an import line.

martijnrusschen:
Thanks for helping with this!

PR #677, opened 2016-12-16
strip propTypes declarations in production bundle

jochenberger:


jochenberger:
See https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types.
I would have liked to do this via `.babelrc`, but unfortunately I haven't found a way to set the babel `env` from grunt.

PR #690, opened 2017-01-09
Render Calendar children

aij:
Allows adding extra content within the calendar.

Also, fix stylesheet so &lt;pre&gt; works.


martijnrusschen:
Thanks!

PR #694, opened 2017-01-11
Verify we have input 'ref' before calling 'focus()' to avoid NPE

travisbt:
If you're wrapping a `stateless functional component` you will not have a ref.

### DIFF
--- a/src/date_input.jsx
+++ b/src/date_input.jsx
@@ -83,7 +83,9 @@ var DateInput = React.createClass({
   },
 
   focus () {
-    this.refs.input.focus()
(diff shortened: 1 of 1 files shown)


travisbt:
@martijnrusschen I can also update the `refs` into the preferred ref usage in another PR, per [https://facebook.github.io/react/docs/refs-and-the-dom.html](https://facebook.github.io/react/docs/refs-and-the-dom.html).

martijnrusschen:
Looks good to me, can you add some tests to cover this?

martijnrusschen:
Closing due inactivity.

Issue #695, opened 2017-01-12
Datepicker on mobile device - Suggest using a portal/overlay


ro-savage:
I have been testing out the datepicker for a project I am working on, but have found the user experience on mobile not to be very good.

<img src="https://cloud.githubusercontent.com/assets/9244507/21877229/8065dcd8-d8ee-11e6-8cb5-26feda0ff77f.png" width="400" />

I took a look at how [airbnb's react-dates](https://github.com/airbnb/react-dates) handle mobile and found they use a portal/overlay.

<img src="https://cloud.githubusercontent.com/assets/9244507/21877296/df81d492-d8ee-11e6-8abc-cbf325067654.png" width="400" />

**I suggest making a similar overlay for react-datepicker.**

ro-savage:
I have created PR #696  the adds the the prop `withPortal` which creates the calendar in a portal/overlay.

Here it is running on my project.

<img src="https://cloud.githubusercontent.com/assets/9244507/21877642/48b82ad6-d8f1-11e6-86b6-9ca0e8a45563.png" width="400" />

Happy to hear any feedback.

martijnrusschen:
Pushed 0.40.0 with this included.

leomyth330:
`withPortal` works great on desktop browser.
When I test on mobile, it doesn't prevent outside touch events and if I click button behind overlay, it's click event works.
And browser console says 

> Unable to preventDefault inside passive event listener invocation

PR #696, opened 2017-01-12
Add portal overlay for datepicker

ro-savage:
This PR adds a new prop which creates a Portal/Overlay version of the calendar, and fixes issue #695 

This helps if the calendar is being used by a mobile device, or other small screen.

If the prop `withPortal` is passed, it will create a portal version of the calendar.

Example:
```js
<DatePicker
onChange={this.handleChange}
withPortal />
```

The docs have been updated. No tests have been updated as this did not affect any, and no new tests added as the logic is very straight forward.

Happy to hear feedback/changes.

### DIFF
--- a/docs-site/bundle.js
+++ b/docs-site/bundle.js
@@ -480,7 +480,7 @@
 
 /***/ },
 /* 6 */
-[634, 7],
+[637, 7],
 /* 7 */
 /***/ function(module, exports) {
 
@@ -5500,7 +5500,7 @@
 
 /***/ },
 /* 47 */
-[634, 32],
+[637, 32],
 /* 48 */
 /***/ function(module, exports, __webpack_require__) {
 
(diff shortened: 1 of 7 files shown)


rafeememon:
I'm pretty skeptical of introducing a new dependency for this; unfortunately I don't think this should be included.  @martijnrusschen thoughts?

ro-savage:
`react-portal` is the same module `react-dates` uses to create their portal. 

The alternative I had originally played with was to create a div/portal that contained the calendar set to `display: none;` then apply a new class when the calendar was opened that made the div visible.

This has a few issues that I didn't get around to resolving but nothing that looked like a show stopper. However, when I ran into them I checked the code for airbnb's dates module and saw they used react-portal. So decided to run with that encase there was more edge cases I wasn't aware off.

ro-savage:
I have created PR #698 which removes the dependancy.

rafeememon:
Closing in favor of #698

PR #697, opened 2017-01-12
Upgrade from scss-lint to sass-lint

aij:
sass-lint is pure JS, so this removes the Ruby dependency.

I'm not sure this is actually something you want, but it would make things more convenient for people I work with at least.


martijnrusschen:
Looks good to me. Can you make sure the build succeeds?

PR #698, opened 2017-01-12
Add portal for mobile support

ro-savage:
This PR adds a new prop which creates a Portal/Overlay version of the calendar, and fixes issue #695 

This helps if the calendar is being used by a mobile device, or other small screen.

If the prop `withPortal` is passed, it will create a portal version of the calendar.

Example:
```js
<DatePicker
onChange={this.handleChange}
withPortal />
```

The docs have been updated. No tests have been updated as this did not affect any, and no new tests added as the logic is very straight forward.

Happy to hear feedback/changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@ dist
 lib
 tmp
 coverage
+docs-site/bundle.js
+docs-site/style.css
 
 docs/calendar.md
 docs/date_input.md
--- a/docs-site/src/example_components.jsx
+++ b/docs-site/src/example_components.jsx
@@ -31,6 +31,7 @@ import CustomInput from './examples/custom_input'
 import MultiMonth from './examples/multi_month'
 import MultiMonthDrp from './examples/multi_month_drp'
 import Children from './examples/children'
+import Portal from './examples/portal'
 
 import 'react-datepicker/dist/react-datepicker.css'
 import './style.scss'
@@ -110,6 +111,10 @@ export default React.createClass({
     {
       title: 'Configure Popover Placement',
       component: <Placement />
+    },
+    {
+      title: 'Portal version',
+      component: <Portal />
     },
     {
       title: 'TabIndex',
--- a/docs-site/src/examples/portal.jsx
+++ b/docs-site/src/examples/portal.jsx
@@ -1,0 +1,38 @@
+import React from 'react'
+import DatePicker from 'react-datepicker'
+import moment from 'moment'
+
+export default React.createClass({
+  displayName: 'With Portal',
+
+  getInitialState () {
+    return {
+      startDate: moment()
+    }
+  },
+
+  handleChange (date) {
+    this.setState({
+      startDate: date
+    })
+  },
+
+  render () {
+    return <div className="row">
+      <pre className="column example__code">
+        <code className="jsx">
+          {'<DatePicker'}<br />
+              {'selected={this.state.startDate}'}<br />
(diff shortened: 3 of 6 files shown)


rafeememon:
> no new tests added as the logic is very straight forward.

This is not a good reason to not add tests; please include tests, even for basic functionality =)

docs-site/style.css
```diff
@@ -287,6 +287,39 @@
 
 .react-datepicker__tether-element {
   z-index: 2147483647; }
+
+.react-datepicker__portal {
```
rafeememon: What does this look like on the docs site without any of these overridden styles?  I'd prefer if the docs site reflects exactly what this will look like out-of-the-box

ro-savage: I am not sure what you mean by this?

If the portal is not shown, then it will use default styles.

If the portal is shown, it overwrites styles to make the text/calendar slightly bigger / more readable on a mobile device.

You can see what it looks like by scrolling down to the portal example and clicking the calendar.

ro-savage:
@rafeememon I have added tests for checking if withPortal is correctly working.

My testing experience is limited, so you might want to check if those 2 tests cover enough.

docs-site/style.css
rafeememon: These styles are added to the demo site but are NOT in the datepicker stylesheet.  Why are they necessary, and what does the demo site look like if you remove them from `docs-site/style.css`?

rafeememon: Oops, looks like this file is generated.  Can you remove the built files from your PR -- we automatically build them when we make a release.

ro-savage: Yep I can remove generated files (just need to figure out how!)

Would it make sense to have them (`doc-site/`) in `.gitignore` then? 
As every time it's someone makes a change and builds the files, it'll show has changed/updated in git.

ro-savage: Files now removed.

martijnrusschen:
@ro-savage Thanks for creating this. How's your progress with the internal testing?

ro-savage:
@martijnrusschen @rafeememon 

There is a bug with iOS Safari in landscape mode which is causing issues. (**Only on iPhone & in Landscape mode**. iPad appears fine).

It appears when React creates the new element, iOS safari renders the div without doing a full redraw/applying CSS styles and therefore the calendar doesn't appear on screen.

I tried a few things to force a redraw but couldn't find out how to reliably do it.

ios calendar bug

This is an example, if we change tabs and come back it forces a redraw/render. Changing from landscape->portrait->landscape works as well. 

Internally we are going to push ahead despite that bug, and revisit it later.

**So if thats still ok for you, it's ready to merge.** 

If the iOS bug is a show stopper. Then I can update the PR in the future. But it might be a couple weeks away. If the merge is accepted, I'll submit a new PR with a fix whenever it becomes a priority for our product.

ro-savage:
Alternative strategies would be:
- Use the react-portal (although I haven't tested if that actually fixes the issue)
- Render the portal + calendar (or maybe just portal without calendar) on every div, with display: none. However this will mean every input has its own calendar rendered and overlay, which isn't ideal.

martijnrusschen:
Seems fine for now, we can revisit later and if you find the fix in the mean time let me know!

bshirah89:
Question Is there a way to customize the class of 'react-datepicker__portal', I am trying to lighten up the overlay for my use case.

PR #702, opened 2017-01-13
Add an option that disabled navigation keys

dpwrussell:
Optionally disables arrows, page up/down, home and end keys

While I think it is nice to have the keyboard interaction available, it is also good to be able to turn it off.

The motivation for this is that we found that users in equal part either wanted to edit the date as text with keyboard functions (e.g. arrow keys) or wanted to click a date. So in our case, disabling the navigation only allows the former without affecting the latter. This PR makes that possible.

### DIFF
--- a/docs-site/src/examples/disabled_navigation.jsx
+++ b/docs-site/src/examples/disabled_navigation.jsx
@@ -1,0 +1,40 @@
+import React from 'react'
+import DatePicker from 'react-datepicker'
+import moment from 'moment'
+
+export default React.createClass({
+  displayName: 'DisabledNavigation',
+
+  getInitialState () {
+    return {
+      startDate: moment()
+    }
+  },
+
+  handleChange (date) {
+    this.setState({
+      startDate: date
+    })
+  },
+
+  render () {
+    return <div className="row">
+      <pre className="column example__code">
+        <code className="jsx">
+          {'<DatePicker'}<br />
+              {'selected={this.state.startDate}'}<br />
+              {'onChange={this.handleChange}'}<br />
+              <strong>{'disabledNavigation />'}</strong><br />
+              {'placeholderText="This has disabled navigation"'}
+        </code>
+      </pre>
+      <div className="column">
+        <DatePicker
+            selected={this.state.startDate}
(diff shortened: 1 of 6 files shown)


docs-site/src/examples/disabled_navigation.jsx
```diff
@@ -0,0 +1,40 @@
...
+              {'onChange={this.handleChange}'}<br />
+              <strong>{'disabledNavigation />'}</strong><br />
+              {'placeholderText="This has disabled navigation"'}
+        </code>
+      </pre>
+      <div className="column">
+        <DatePicker
+            selected={this.state.startDate}
+            onChange={this.handleChange}
+            disabledNavigation
```
martijnrusschen: Would `disabledKeyboardNavigation` be better? Or do I miss something?

docs/datepicker.md
```diff
@@ -43,6 +43,12 @@ type: `bool`
 defaultValue: `false`
 
 
+### `disabledNavigation`
```
martijnrusschen: You can remove this, docs will be automatically updated on every release.

src/datepicker.jsx
```diff
@@ -184,30 +186,33 @@ var DatePicker = React.createClass({
...
-      event.preventDefault()
-      this.setSelected(copy.add(1, 'months'))
-    } else if (event.key === 'Home') {
-      event.preventDefault()
-      this.setSelected(copy.subtract(1, 'years'))
-    } else if (event.key === 'End') {
-      event.preventDefault()
-      this.setSelected(copy.add(1, 'years'))
+    }
+    if (!this.props.disabledNavigation) {
```
martijnrusschen: Maybe add some tests to cover your change?

dpwrussell:
Thanks for reviewing. This should address all the comments. I decided that it made sense to drop the checks for the effect of the keyboard navigation (e.g.):

```js
data.copyM.add(1, 'years')
expect(result.format(data.testFormat)).to.equal(data.copyM.format(data.testFormat))
```

...as the fact that it is not calling the callback handler at all makes this moot.

martijnrusschen:
Thanks!

Issue #705, opened 2017-01-17
utcOffset prop's default is documented incorrectly


dylanscott:
The datepicker's `offsetProp`'s default [is documented as](https://github.com/Hacker0x01/react-datepicker/blob/master/docs/datepicker.md#utcoffset) `moment().utc().utcOffset()` (which is always zero, regardless of the browser's timezone). However, [looking at the code](https://github.com/Hacker0x01/react-datepicker/blob/master/src/datepicker.jsx#L95), the default is actually set to `moment().utcOffset()` (which is the browser's timezone, as reported by `new Date().getTimezoneOffset()`).

dylanscott:
I've created a PR for this #706, but I'm not sure if this is the right way to change the documentation, since the previous commits to touch that doc file have "automated commit" in their messages so I'm not sure if this documentation is generated from elsewhere.

martijnrusschen:
Should fixed now. Thank!

PR #706, opened 2017-01-17
fix documentation for utcOffset prop default #705

dylanscott:


martijnrusschen:
@dylanscott Thanks for making this change. However, our documentation is auto generated on each release. This will be fixed with the next release.

Issue #716, opened 2017-01-23
Whitespace in pasted in dates.


MikaelCarpenter:
I happen to have a use case where a lot of users paste in dates. Occasionally an extra space gets pasted in, which causes the date to not be recognized as valid. This seems like a trivial thing to take care of. Will make a PR.

PR #717, opened 2017-01-23
Trim Whitespace from Date

MikaelCarpenter:
**Issue:** https://github.com/Hacker0x01/react-datepicker/issues/716

### What
Trim the whitespace from date values before checking if `.isValid()`

### Why
If a date gets pasted in with extra whitespace, it should still be recognized as valid.

**BEFORE**
whitespace_old

**AFTER**
whitespace


jochenberger:
Not sure I like this. Isn't this the kind of thing you should do with a `customInput`?

MikaelCarpenter:
hmm I don't think so? It's actually odd enough that the presence of trailing spaces makes a date not recognized as valid. I don't think this is something that is necessary in just certain use cases, I think it's a fundamental issue with the `.isValid` check and should be handled for all use cases.

martijnrusschen:
I'm a little in doubt here. But it seems a fairly simple solution. Can you add a test case to make sure this doesn't break once someone tries to refactor this and doesn't know about this case?

MikaelCarpenter:
@martijnrusschen Added a test that makes sure a date with trailing whitespace gets treated as a valid date. It passes locally for me, but it's failing here? Do you happen to see anything wrong with it? 

Also as a side note I had to use `mount` rather than `shallow` to access the props. Shallow's props gives the props for the `input` not the `DateInput` for some reason.

rafeememon:
From travis

```
Running "karma:unit" (karma) task
25 01 2017 00:00:27.801:INFO [karma]: Karma v1.3.0 server started at http://localhost:9876/
25 01 2017 00:00:27.804:INFO [launcher]: Launching browser Firefox with unlimited concurrency
25 01 2017 00:00:27.809:INFO [launcher]: Starting browser Firefox
25 01 2017 00:00:29.887:INFO [Firefox 31.0.0 (Linux 0.0.0)]: Connected on socket /#p3oy7_xb8j_b6wUrAAAA with id 17824565
...........................................
Firefox 31.0.0 (Linux 0.0.0) DateInput should recognize a date with trailing whitespace as valid and call onChangeDate with the trimmed date FAILED
	must be called once with expectedDate
	AssertionError@node_modules/chai/chai.js:5553:5
	[3]</module.exports/Assertion.prototype.assert@node_modules/chai/chai.js:210:1
	[6]</module.exports/chai.assert@node_modules/chai/chai.js:2192:1
	@webpack:///test/date_input_test.js:103:4 <- test/index.js:49857:6
```

MikaelCarpenter:
@rafeememon Yea I saw that and I get that the test I added is failing, but it doesn't say why. I can only assume that either the callback wasn't called or that it was called but with a different value. 

However, this test passes for me locally so I'm not sure what would make it fail now.

jochenberger:
The only case I can imaging this going wrong is if the *pattern* has leading/trailing spaces, but this is probably not a very common usage. It might be worth adding a warning/check for that too. Maybe only `trim()` if `input.length > pattern.length` or something like that?

rafeememon:
I don't think comparing the input's length to the pattern's is a good option -- a format can produce a string greater in length than the pattern, e.g. `MMMM` becomes `January`

jochenberger:
Yes, of course. I should have had more coffee before chiming in here. ;-)

MikaelCarpenter:
So does anyone else have any idea why it might be failing? I'm having some coworkers run it locally to see if it fails for them, but it might be helpful for one of you guys to try that too.

mgrdevport:
The sinon function withArgs(args) has to be used on the spy creation.

Working:
```javascript
var callback = sinon.spy().withArgs(expectedDate)
...mount...
expect(callback.calledOnce).to.be.true
```

If you want to create a generic spy:
```javascript
var callback = sinon.spy()
...mount...
expect(callback.calledOnce).to.be.true
expect(callback.calledWithExactly(expectedDate)).to.be.true
```

Sometimes its even better to compare the arguments with equal to be able to lookup the expected and actual value on failure:
```javascript
var callback = sinon.spy()
...mount...
expect(callback.calledOnce).to.be.true
expect(moment(callback.args[0][0]).format(dateFormat)).to.equal(expectedDate.format(dateFormat))
```

MikaelCarpenter:
Thanks @mgrdevport that `.withArgs` must have been causing problems. Modified the test to check the two separately

martijnrusschen:
Awesome, thanks!

PR #721, opened 2017-01-25
Respect min/max dates on keyboard navigation

### LINKED ISSUE #720: Keyboard Navigation ignores min/max dates
If you set a min and max date, the datepicker tries to prevent you from selecting dates outside of that range. However, the new keyboard navigation feature is able to ignore the min and max dates.

In the gif, I have min date set to `1/24/2017` and then proceed to use my arrow keys to select a date before the min date.
keyboard-min-date

mgrdevport:
Fixes #720 
- change preSelected date on keyboard navigation only if new selection is between minDate and maxDate (if props are present)


package.json
```diff
@@ -92,9 +93,9 @@
     "tether": "^1.3.2"
   },
   "scripts": {
-    "build": "NODE_ENV=production grunt build",
+    "build": "cross-env NODE_ENV=production grunt build",
```
martijnrusschen: Just curious, why is adding cross-env needed here?

package.json
```diff
@@ -37,6 +37,7 @@
     "babel-preset-stage-0": "^6.5.0",
     "chai": "^3.5.0",
     "codecov.io": "^0.1.6",
+    "cross-env": "^3.1.4",
```
martijnrusschen: Can you please run `yarn` so this package is added to the lock file as well?

package.json
mgrdevport: I only had a windows machine to build at this time and cross-env enabled me to set the env-param there. You don't need it on unix systems, but on windows the npm scripts would fail -> I've removed the dependency and restored the old scripts.

martijnrusschen:
Happy to merge this, can you resolve the merge conflicts so I can merge this?

mgrdevport:
@martijnrusschen 
Done.

Issue #731, opened 2017-02-06
PreventFocus timeout is not cleared


TheWalruzz:
During development of my project I stumbled upon issue with [this line](https://github.com/Hacker0x01/react-datepicker/blob/master/src/datepicker.jsx#L162) - timeout was unable to setState on components that were already unmounted.
From what I've seen, simple clearTimeout in componentWillUnmount is sufficient in that case.

I will create a PR with the fix.

PR #732, opened 2017-02-06
PreventFocus timeout is now cleared when datepicker is unmounted #731

TheWalruzz:
This PR fixes [issue](https://github.com/Hacker0x01/react-datepicker/issues/731) with datepicker trying to set state after being already unmounted thus resulting in React warning.

Things that were done:
- timeout id is now stored inside datepicker component,
- whenever component wants to unmount, timeout is cleared if it's created


martijnrusschen:
Thanks!

PR #735, opened 2017-02-11
Upgrade to Webpack 2

martijnrusschen:


martijnrusschen:
Todo:
- Fix the build
- Replace codecov, or leave it out?

skimi:
Don't forget to enable tree shaking with
```
    new webpack.LoaderOptionsPlugin({
      minimize: true,
      debug: false
    }),
```

Note that this might minimize everything too. If you don't want that I think you need to give each loader a minimize false argument.

PR #740, opened 2017-02-17
className prop should always be applied to the outermost element, old className prop renamed to inputClassName

Qrysto:
When I pass the `className` prop, intuitively I always expect it to be applied to the outermost element rendered, not some element deep inside. For example your DatePicker is in a `flex` element, you need to control the `order` or `flex-grow` of it, you cannot achieve the same effect when apply the styles into the input. Or when you need the component to be full width, setting `width: 100%` on the input is pointless if the outer `div` is `inline-block` and cannot be changed (it currently is).

This pull request changes the way `className` prop works, making it always be applied to the outermost element rendered, and the old `className` prop is renamed to `inputClassName` to be exact. Example docs site is modified accordingly.

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -41,6 +41,7 @@ var DatePicker = React.createClass({
     id: React.PropTypes.string,
     includeDates: React.PropTypes.array,
     inline: React.PropTypes.bool,
+    inputClassName: React.PropTypes.string,
     isClearable: React.PropTypes.bool,
     locale: React.PropTypes.string,
     maxDate: React.PropTypes.object,
@@ -325,7 +326,7 @@ var DatePicker = React.createClass({
   },
 
   renderDateInput () {
-    var className = classnames(this.props.className, {
+    var className = classnames(this.props.inputClassName, {
       [outsideClickIgnoreClass]: this.state.open
     })
     return <DateInput
@@ -375,7 +376,7 @@ var DatePicker = React.createClass({
 
     if (this.props.withPortal) {
       return (
-        <div>
+        <div className={this.props.className}>
           <div className="react-datepicker__input-container">
             {this.renderDateInput()}
             {this.renderClearButton()}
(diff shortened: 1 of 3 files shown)


src/datepicker.jsx
```diff
@@ -325,7 +326,7 @@ var DatePicker = React.createClass({
   },
 
   renderDateInput () {
-    var className = classnames(this.props.className, {
+    var className = classnames(this.props.inputClassName, {
```
rafeememon: As unfortunate as the existing behavior is, this is a breaking change.  Even though semver dictates that we are allowed to break things since we're 0.x, I'm extremely hesitant to make this change.

Qrysto: then please keep this in mind and when you have other major changes, you can consider taking this up along with them to make all breaking changes at once, does that make sense?

src/datepicker.jsx
```diff
@@ -399,7 +400,7 @@ var DatePicker = React.createClass({
           targetOffset={this.props.popoverTargetOffset}
           renderElementTo={this.props.renderCalendarTo}
           constraints={this.props.tetherConstraints}>
-        <div className="react-datepicker__input-container">
+        <div className={classnames('react-datepicker__input-container', this.props.className)}>
```
Gurkenschreck: I second that. Being able to overwrite react-datepicker__input-container would be nice.

martijnrusschen:
Lets get this rebased and merged. I'm cleaning stuff up, so we can make the last breaking changes before we announce a 1.0.

src/datepicker.jsx
simkesd: @rafeememon Is it possible then to add other `div` with `this.props.componentWrapperClass` or smth like that? I'm also having problem with this as I can't easily change DatePicker to inline element.

joenot443:
Is there any chance this will end up being merged? It would be a huge help for my project.

j33n:
Is there any reason this was put on hold? Looks like it was a needed change

PR #750, opened 2017-02-26
Improve the datepicker for using inline and portal props together.

Ch-Valentine:
I need open the datepicker in a portal after click on an icon in a my custom field.


Ch-Valentine:
What do i need to do that this pull request will merge ?

Issue #749, opened 2017-02-25
Launching V1.0


martijnrusschen:
About a year ago, we did our latest try on launching a v1.0 version. I think it would be good to define a v1.0 version so that we can apply breaking changes when needed and show the maturity of this component. 

Our last effort in defining what should be done before we launch v1.0 was documented in https://github.com/Hacker0x01/react-datepicker/milestone/2. What do you guys think of this? Should be continue with this, create a new Milestone, or just call it a day and launch v1.0 in the current state?

Todos:
- [ ] Webpack2?
- [x] Remove grunt?
- [ ] Decide if we leave the API as it is?

martijnrusschen:
Adding some recent contributors for their opinion on this: @mgrdevport, @jochenberger, @ro-savage, @rafeememon.

jochenberger:
I haven't used react-datepicker in a while because I had too many issues with it, so I don't have a strong opinion.
You might want to consider doing some alpha/beta releases before 1.0.

ro-savage:
Unfortunately the portal doesn't work as expected on iphone iOS 10.

When you click the datepicker, the portal should be fixed, take up the entire screen, and remove the ability to scroll. 

This works on all devices we tested including iPad iOS10 but does not work on iPhone iOS10, instead the portal appears at the top of the page, and the viewport stays focused on the the input.

My guess is this is something to do with `focus` being on the input at the same time it tries to change the viewport, and by not focusing on the input when clicked (or focusing on the calendar) it may fix it.

We worked around it by using the built-in date picker when we detected an iPhone.

Unfortunately, I am not at the company anymore and no longer have access to an iPhone to test.

Maybe release with a known issue, and offer the work around until its fixed?

rafeememon:
IMO the most important thing to settle on is the API, which includes the components we will expose, the props that they'll take in, and possibly even the styles.  I think it stands to be cleaned up pretty substantially, and there are a few breaks that we've wanted to make but deferred on.  Once we've settled on the API I think we'll have a clear path forward to getting to a releasable v1.

ro-savage:
General feedback:

- We used quiet a bit of the API and found it to be good and completed. I wouldn't see any major breaking changes from where it is now.
- `withPortal` sounds like a stupid name to me, but I used the same name as [react-dates](https://github.com/airbnb/react-dates)
- I may be wrong, but I think you have to pass the date picker a moment object. Would be nice to pass a regular JS Date object and it get's handled.
[...]

PR #757, opened 2017-03-03
Remove lodash and replace with inbuilt native functionality

ro-savage:
As discussed in Issue #749, this removes Lodash from the datepicker.
Takes file size from 77kb to 57kb.

Lodash is still in dev deps because it uses in some of the tests.

Before:
image

After:
image

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -1,5 +1,4 @@
 import moment from 'moment'
-import find from 'lodash/find'
 import YearDropdown from './year_dropdown'
 import MonthDropdown from './month_dropdown'
 import Month from './month'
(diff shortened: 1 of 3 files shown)


src/calendar.jsx
```diff
@@ -14,9 +13,9 @@ const DROPDOWN_FOCUS_CLASSNAMES = [
 
 const isDropdownSelect = (element = {}) => {
   const classNames = (element.className || '').split(/\s+/)
-  return !!find(DROPDOWN_FOCUS_CLASSNAMES, (testClassname) => {
+  return DROPDOWN_FOCUS_CLASSNAMES.filter((testClassname) => {
```
rafeememon: How about using `.some` instead?  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

ro-savage: Yep. Good idea. Updated.

martijnrusschen:
This is amazing!

martijnrusschen:
Can you also remove lodash from the package.json or are we still using this internally?

ro-savage:
@martijnrusschen lodash is still used in a few places
https://github.com/Hacker0x01/react-datepicker/search?utf8=%E2%9C%93&q=lodash
in a couple tests, in the gruntfile and the server.

I didn't remove them because they don't hurt the package size and lodash can be easier to reason with than some of the code that may replace it.

Although I can do a 2nd PR to remove lodash from the entire repo if you'd like. But I wont be able to work on that for at least a few days.

rafeememon:
That's fine, we can address removing it in the supporting code later, if we do it at all.  Thanks!

Issue #768, opened 2017-03-11
Accessibility issue:  Invalid date string gets overwritten when leaving input field


greg-a-smith:
When the input field loses focus and the field contains an invalid date string, it is clearing the field and restoring the previously-selected date (or blank if none was previously selected).  This is an accessibility issue because when the field is overwritten back to the previous date, there is nothing that would notify a screen reader that it occurred.

We would like to propose a propType to disable the auto-correction of invalid dates.  Our consumers of this DatePicker would handle their own date validation and could mark the field as "invalid" which would alert a screen reader of the state change.

**NOTE:  We have a proposed solution in PR #769 .**

aij:
I think any fix for #547 will be also fix this issue in a more general way. Would you mind giving #773 a spin?

IMHO, it would also be nicer to have the less annoying behavior by default rather than having to explicitly enable it via a prop. Is it even worth making the current behavior of losing user input an option?

greg-a-smith:
@aij I checked out your pull request #773 and this seems to be a good solution for this issue.  My current pull request #769 could be closed if your pull request got approved and merged.

As for whether it is default behavior or enabled via a prop -- default is always nice, but it doesn't matter to me either way.  We use a component in our library that wraps the react-datepicker component so it would be invisible to our consumers.

aij:
Fixed in #773

PR #769, opened 2017-03-12
Changes to support disabling auto correction of dates

greg-a-smith:
These are the suggested changes to address issue #768 .

### DIFF
--- a/docs-site/src/example_components.jsx
+++ b/docs-site/src/example_components.jsx
@@ -3,6 +3,7 @@ import hljs from 'highlight.js'
 import Default from './examples/default'
 import CodeExampleComponent from './code_example_component'
 
+import DisableDateAutoCorrection from './examples/disable_date_auto_correction'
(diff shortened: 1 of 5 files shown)


greg-a-smith:
Looking at the code coverage report, it doesn't appear there are any additional new lines that can be tested.  There are some additional lines defining the prop and its default values which seem to be causing the overall line coverage percentage to drop.  I am not quite sure how to address this.

Please let me know how to proceed.  Thanks!

osbornm:
@martijnrusschen Any idea on the Code Coverage issue @greg-a-smith is talking about? And is there anything else needed for this PR?

martijnrusschen:
The coverage reports aren't always helpful. I'm fine with merging in this case.

Trend20:
Is there any solution for this when using an Angular Datepicker? I have got the same problem but I am using Angular.

PR #710, opened 2017-01-19
WIP: Controlled input

aij:
I'm trying to fix the issue where user input gets reformatted as the user is typing (when the selected dateFormats share a prefix) #560 

Some tests broke after rebasing, but I'm not sure I understand how the new utcOffset changes are supposed to work... Was that meant to be an API breaking change?

Issue #560, opened 2016-09-06
Custom Input/Multi Date Format


Haemoglobin:
Hi, 

I'm thinking of using the new custom input functionality (https://github.com/Hacker0x01/react-datepicker/issues/529) in order to support multiple date formats (https://github.com/Hacker0x01/react-datepicker/issues/448). 

I need the ability to enter a variety of things like 1/9/2016 01/09/2016 01092016 01-09-2016 etc. 

First question, is there a reason that moment is set to strict mode parsing at this line https://github.com/Hacker0x01/react-datepicker/blob/v0.25.0/src/date_input.jsx#L56 ? If it was relaxed that might be all I need.   

However, with that in, I'm thinking of parsing/reformatting the date myself in a custom input element and raising onChange using the format that react-datepicker is configured to use? Do you think that will work? 
If so, do you know when the next release will be published that includes custom input? 

Thanks, 
Hamish

luggage66:
You can accept multiple import formats by providing an array to the `dateFormat` property. The first format will be the displayed format.

e.g.

```js
<DatePicker dateFormat={[ 'D/M/YYYY', 'DDMMYYYY', 'DD-MM-YYYY' ]} />
```

Note that as soon as the user types a value that is valid in **any** of those formats, it'll reformat and put the cursor at the end.

luggage66:
Nevermind, I don't thing this is a solution. If you accept for M/D/YY and M/D/YYYY then as soon as you type two digits for a year, it assumes the whole year. If you tried to type "1999" if would assume you meant "2019".

Haemoglobin:
I managed to get it working with the custom input in the end, but it wasn't entirely easy to get working due to various issues, the cursor problem you mentioned for example

aij:
For me, the problem is that as soon as the string is parseable as a date (Eg: 10/10/19) it sends on onChange event, and we feed the resulting moment back into DatePicker via the selected prop. That highlights the date (2019-10-10) in the mini-calendar (which is good), but it also updates the value prop of the input field (which is bad if the user wasn't done typing, as they end up with Eg: 10/10/201999).

aij:
Hmm, I was trying to fix this, but part of the issue is that `DateInput` is used neither as a controlled input nor an uncontrolled input, but something kind of in between. 

[...]

Issue #547, opened 2016-08-16
allow datepicker to be a controlled component


bbmreed:
date_input.jsx does not always call `props.onChange`, eg if the date is in valid or not within a valid range.

This means that a parent component using date_picker as a controlled component will not always be aware of the user's input, which has the following pitfalls:
- the user's invalid data can not be stored and re-displayed (aka it disappears on any re-render)
- the user's input can not be inspected to provide warnings (aka "out of range" / "invalid") 

This is especially jarring when typing in a syntactically valid date that is not within range - the typed in value disappears, and we can't give a warning as to why.

example: https://gyazo.com/c3404b9dcfdc8497a92466bd4dc77717.gif

I believe this was the original commit:

https://github.com/Hacker0x01/react-datepicker/pull/346/commits/b8fc931399f50eecf27203cb7d976483b7f18c54

More on what I mean by a controlled component: https://facebook.github.io/react/docs/forms.html#controlled-components

aij:
Does the new `onChangeRaw` solve this?

PR #773, opened 2017-03-15
Fix two issues related to uncontrolled components

aij:
First is a rebase of #710. (Fixes input getting reformatted as the user is typing)

Second is a fix to allow the user of the library to control the value in the input field. (via the `value` prop)


bcullman:
Are the tests in the deleted file 'date_input_test.js' still valid for the datepicker? Should they be relocated?

martijnrusschen:
Thanks @aij for putting this together. Can you maybe explain a bit on what's happening in this PR? Also, I merged some other Pull Requests so a rebase is welcome.

aij:
@martijnrusschen The main problem I was trying to solve, is that the date picker loses (or throws away) user input. (Which has resulted in various bug reports.)

I think @bbmreed explained the problem reasonably well in #547. 

#560 is basically a special case of the same problem. (User input is lost as soon as it can be parsed.)

#768 is another special case. (User input is lost on blur, which is especially unfriendly to users with disabilities.)

I assume you are familiar with controlled vs. uncontrolled components.

The root of the problems is that `DateInput` is not a controlled component (though it does not behave like a fully uncontrolled component either). Because of that, `DatePicker` isn't controlling the input, which means the application can't either.

So, the obvious next step is to make `DateInput` a controlled component. When I went to do that, I realized it would no longer serve much of a purpose. Back when it was first created in 9fd86b2aa, it's main purpose seemed to be to handle keyboard events.  Now that has moved to `DatePicker`, and all `DateInput` does is store the current value of the input field, and convert the string to and from moments. Changing it to receive and emit raw strings (and not keep state internally any more) would leave it with very little purpose. (It could still be in charge of converting strings to dates and emitting moment events, but having a component for that seems like unnecessary complexity IMHO. If you disagree it could be kept though.)

The only major merge conflict now is with commit a95cd8666, which was just merged, and as mentioned in the bug solves a subset of the same problem in a more specialized way...  I'm not sure how you would want the two to be merged. (I'd be inclined to revert the special-case fix in favor of the more general fix, but since you just merged it, I wouldn't expect that to be well received without discussion first.) So... What are your thoughts on reverting that change since it would no longer be needed?

I'll add some line-comments to try to explain specific changes, but if there's anything else you have questions about, feel free to ask.

src/date_utils.js
```diff
@@ -64,3 +64,14 @@ export function getEffectiveMaxDate ({ maxDate, includeDates }) {
     return maxDate
   }
 }
+
+export function parseDate (value, { dateFormat, locale }) {
```
aij: I originally factored out these functions from date_input. Currently, they are only used in datepicker, but they seemed more general purpose. If you decide we should keep the current date_input API (in case anyone is using it directly), then I think it makes sense to factor these out in order to be able to maintain consistency, but otherwise they could be inlined into datepicker.jsx if you prefer.

src/datepicker.jsx
```diff
@@ -24,7 +23,7 @@ var DatePicker = React.createClass({
     children: React.PropTypes.node,
     className: React.PropTypes.string,
     customInput: React.PropTypes.element,
-    dateFormat: React.PropTypes.oneOfType([
+    dateFormat: React.PropTypes.oneOfType([ // eslint-disable-line react/no-unused-prop-types
```
aij: `eslint-disable-line` is needed here because eslint is not picking up on the fact that the prop is actually used via `parseDate` and `safeDateFormat`.

src/datepicker.jsx
```diff
@@ -74,11 +73,13 @@ var DatePicker = React.createClass({
     title: React.PropTypes.string,
     todayButton: React.PropTypes.string,
     utcOffset: React.PropTypes.number,
+    value: React.PropTypes.string,
```
aij: This is the new value prop, that now allows the app to control the value of the input, even when it needs to be set to something other than a valid date in the default format.

src/datepicker.jsx
```diff
@@ -166,6 +167,20 @@ var DatePicker = React.createClass({
     if (this.props.withPortal) { event.preventDefault() }
   },
 
+  handleChange (event) {
```
aij: This is where the raw string events are now handled. (Basically replacing `DateInput.handleChange` and `DateInput.handleChangeDate`.)

src/datepicker.jsx
```diff
@@ -166,6 +167,20 @@ var DatePicker = React.createClass({
...
+    if (this.props.onChangeRaw) {
+      this.props.onChangeRaw(event)
+      if (event.isDefaultPrevented()) {
+        return
+      }
+    }
+    this.setState({ inputValue: event.target.value })
```
aij: The input value is now maintained as state here, for backwards compatibility with existing apps which do not control it themselves via the `value` prop.

src/datepicker.jsx
```diff
@@ -199,9 +214,11 @@ var DatePicker = React.createClass({
           preSelection: changedDate
         })
       }
-
       this.props.onChange(changedDate, event)
     }
+    if (!keepInput) {
```
aij: This avoids overriding user input as soon as the date is parseable, which solves #560 without requiring the app to control the input. (Though if you want tho require the app to control the input value when enabling multiple formats, this could be dropped.)

src/datepicker.jsx
```diff
@@ -329,34 +346,33 @@ var DatePicker = React.createClass({
...
-        readOnly={this.props.readOnly}
-        required={this.props.required}
-        tabIndex={this.props.tabIndex}
-        customInput={this.props.customInput} />
+
+    const customInput = this.props.customInput || <input type="text" />
+    const inputValue =
+      typeof this.props.value === 'string' ? this.props.value
+        : typeof this.state.inputValue === 'string' ? this.state.inputValue
+        : safeDateFormat(this.props.selected, this.props)
```
aij: When the datepicker is being used as a controlled component, the input value is determined strictly via the `value` prop.

Otherwise, if the user has entered text, that string is maintained as component state, and we should show it. Otherwise, render the `selected` prop as a string in the appropriate format.

src/datepicker.jsx
```diff
@@ -329,34 +346,33 @@ var DatePicker = React.createClass({
...
-        disabled={this.props.disabled}
-        autoComplete={this.props.autoComplete}
-        className={className}
-        title={this.props.title}
-        readOnly={this.props.readOnly}
-        required={this.props.required}
-        tabIndex={this.props.tabIndex}
-        customInput={this.props.customInput} />
+
+    const customInput = this.props.customInput || <input type="text" />
```
aij: One of the things `DateInput` did was provide the default input type, though it did it via an if/else listing all the props to pass on both branches. This allows us to avoid that duplication.

martijnrusschen:
Thanks for the extensive comment! I think having a more generic solution for the problem we solved in the recently merged PR should be fine. I'm not really against that, I think I didn't realize that this was solving the problem as well when I merged that PR. Lets try getting this PR merged and see where we are after that. I'd really like focussing on getting in a more stable state again and it seems this would help moving to a more generic solution for multiple problems.

Thanks again!

aij:
@martijnrusschen Ok, rebased and squashed, though I haven't gotten a chance to re-test it yet.

aij:
Let me know if you want any other changes.

martijnrusschen:
Sorry for the delay on this. Thanks for the useful explanation. I'm going to merge this and hope nothing serious brakes :)

PR #784, opened 2017-03-21
'OnSelect' callback

alex-shamshurin:
In my case I need it in inlined mode in modal window when datepicker need some callback to close modal when user clicks on any date, even on the current `selected` date. The `onChange` fires only when value has changed.


alex-shamshurin:
This is with merge commit and should not be merged

PR #791, opened 2017-03-26
Ignoring yarn.lock

mberneti:


martijnrusschen:
Can you tell me why you want to ignore this?

mberneti:
The problem is, once a developer run yarn install on its own computer and have the yarn.lock file pushed to git, all packages are resolved to "https://registry.yarnpkg.com/*" which is not accessible on out CI machine, so even when our CI machine has a yarn config set registry {out-local-registry}, yarn tries to fetch packages from registry.yarnpkg.com which causes network errors!

alex-shamshurin:
The typical workflow is to INCLUDE `yarn.lock` everywhere into git. I'm not sure we should break it for this specific case.

martijnrusschen:
Agree, it's best practice to have the yarn.lock file checked into git.

PR #801, opened 2017-03-30
Make Sass variables overridable

jakezatecky:
Changing the colors used in this library is somewhat painful. Either you have to copy over the `.scss` files and change `variables.scss` yourself or you have to override the colors with the same CSS rules. Instead, this PR makes the variables default, allowing you to import the stylesheet and set the colors easily:

``` scss
$datepicker__selected-color: #990c41;

@import "node_modules/react-datepicker/src/stylesheets/datepicker";

...
```

Because the variable names are relatively simple and could conflict with existing variables, I went ahead and modified the variables to add a `datepicker__` prefix, as there was already a variable called `datepicker__margin`. Let me know if you would like for me to revert that change.

### DIFF
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -3,11 +3,11 @@
 
 .react-datepicker {
   font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
-  font-size: $font-size;
+  font-size: $datepicker__font-size;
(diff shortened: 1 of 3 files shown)


martijnrusschen:
Thanks!

ziweiwu:
Thanks. I had trouble with override scss. This issue helped me the most.

PR #811, opened 2017-04-10
Option to set number of years on year dropdown

### LINKED ISSUE #809: Year dropdown improvement ("infinite" scroll / show more than one on arrow click)
Is there a way of achieving this with the current API? So far I haven't found it and I'll probably try to implement it otherwise. 

(Thanks for building the component!)

Cheers

caffeine-based-lifeform:
Fixes #809


oneRice:
Hello, what's the status of this PR? I really need it quick.

caffeine-based-lifeform:
* Fixed conflicts.
* Updated functionality to show the number passed as prop both below and above the selected date (previously showing one less below).

rafeememon:
@caffeine-based-lifeform @martijnrusschen Is there any practical difference between this change, and having the year dropdown respect the min/max dates?  I would prefer to stick to just having it obey min/max, because limiting the number of years is effectively specifying a min and max.

caffeine-based-lifeform:
Our team came across this problem when we intended to use the datepicker to allow the users to specify a birthdate. Since clicking on the arrows only displays one more item in the list, it would be a problem to select a date far in the past. 
By passing the min/max as a prop we can set it to a normal human lifespan and avoid this issue.

I believe it will still obey the min/max dates as default, so it shouldn't be a problem to continue the normal use of the components.

rafeememon:
Got it, I think I misunderstood the purpose of this PR, apologies.

PR #821, opened 2017-04-12
Use JavaScript ES6 class

navinesh:
This fixes React.createClass deprecation warning.


bostonaholic:
Any reason all of these were closed without merging?

Issue #813, opened 2017-04-10
Deprecation warning with React 15.5


obartra:
React is showing the following deprecation warning when using the latest published version of react-datepicker (0.44.0):

>  DateInput: React.createClass is deprecated and will be removed in version 16. Use plain JavaScript classes instead. If you're not yet ready to migrate, create-react-class is available on npm as a drop-in replacement.

There's information about the deprecations [here](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html)

rafeememon:
cc @Pomax re: https://github.com/Pomax/react-onclickoutside/commit/5b023e90b09ea148285296c31041e35d7aa5b8dd

obartra:
Hey there @rafeememon happy to contribute, what should be the approach here, do you want to use `create-react-class`, ES6 classes or something else?

If using ES6 classes, I see you have stage-0 plugins added on babelrc, have you considered using [transform-class-properties](https://babeljs.io/docs/plugins/transform-class-properties/)? It could make migration a bit simpler/cleaner

martijnrusschen:
@obartra ES6 classes seem to be the new standard. Welcoming help to get the codebase migrated!

liady:
@rafeememon any news on this PR?

ltfschoen:
Yes, I simply had to change this:
```
import React from 'react';

const Hello = React.createClass({
    render() {
        return <h1>Hello, world</h1>;
    }
});

export default Hello;
```

to either:
```
import React from 'react';

export default class Hello extends React.Component{
    render() {
        return <h1>Hello, world</h1>;
    }
}
```

or:
```
import React from 'react';

const Hello = () => {
    return <h1>Hello, world</h1>;
};

export default Hello;
```

PR #822, opened 2017-04-13
fix(upgrade): remove deprecation warnings

obartra:
Upgrade to React 15.5.4 and fixed deprecation warnings.

Closes https://github.com/Hacker0x01/react-datepicker/issues/813

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,8 +29,10 @@
   },
   "devDependencies": {
     "babel-core": "^6.22.1",
+    "babel-eslint": "^7.2.2",
     "babel-loader": "^6.2.10",
     "babel-plugin-react-transform": "^2.0.2",
+    "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-react-remove-prop-types": "0.3.2",
     "babel-preset-airbnb": "^2.2.3",
     "babel-preset-es2015": "^6.22.0",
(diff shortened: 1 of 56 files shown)


package.json
```diff
@@ -90,6 +92,7 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "moment": "^2.17.1",
+    "prop-types": "^15.5.8",
     "react-onclickoutside": "^5.9.0",
```
sebastiandeutsch: @obartra if you want to get rid of all deprecation warnings you'll need to update `react-onclickoutside` to `5.11.1`.

sebastiandeutsch:
@obartra do you need any helping hands with this one? I've seen that you transitioned all the classes in your fork. Anything else to do?

obartra:
I need to take another pass at it. I broke a couple tests in `day.jsx` but should be easy to fix.

My main concern is on making sure we are ok using `transform-class-properties`. It's a stage 2 feature but help us get rid of `.bind` everywhere.

sebastiandeutsch:
Awesome! I'm using `transform-class-properties` in my projects too - but eventually @Hacker0x01 has to decide. Ping me if you need help with tedious tasks like putting the binds in the constructor in countless classes :-)

obartra:
@martijnrusschen can you guys take a look see if this along the lines of what you are looking for?

martijnrusschen:
Looks great! Thanks for the contribution!

PR #825, opened 2017-04-15
Prevent 'open' from being called when datepicker is shown inline.

Amauris:
Currently there's an issue with inline calendar, where picking a date, not belonging to the current month, will trigger the calendar to render itself with the wrong month showing. 
This is caused because datepicker.jsx's 'open' is re-writing handleSelect's 'preSelection' modification.

Change includes preventing 'open(false)' from being called if datepicker is an inline calendar.

### DIFF
--- a/docs-site/bundle.js
+++ b/docs-site/bundle.js
@@ -65,14 +65,14 @@
 	return modules;
 }([
 /* 0 */
-/***/ function(module, exports, __webpack_require__) {
+/***/ (function(module, exports, __webpack_require__) {
 
 	module.exports = __webpack_require__(1);
 
 
-/***/ },
+/***/ }),
 /* 1 */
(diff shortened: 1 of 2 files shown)


cozmo:
What would it take to get this PR merged? This is causing us to have to lock ourselves to an older version which is suboptimal.

janbuerling:
What about this error? Is a time near the fixes in sight or is there a workaround?

alex-shamshurin:
Please exclude `build.js` from the commit and resolve other conflicts.

martijnrusschen:
Sorry I merged another PR that's conflicting with yours. Can you double check your changes and rebase against the most recent master? I'll merge it directly afterwards.

martijnrusschen:
Closing this since it's a duplicate of #941 941

PR #835, opened 2017-04-20
Add prop to enable adding custom classes to days based on days date

martinek:
This adds `dayClassName` prop. This can be set to function which gets called for each day of rendered calendar. Function can return any object supported by classnames utility and it will be added to classnames of that day. Function gets rendered day's date as an argument.

This for example enables completely custom highlighting without the need to specify highlighted days upfront.

If function returns undefined, no class is added.

### DIFF
--- a/docs-site/bundle.js
+++ b/docs-site/bundle.js
@@ -65,14 +65,14 @@
 	return modules;
 }([
 /* 0 */
-/***/ function(module, exports, __webpack_require__) {
+/***/ (function(module, exports, __webpack_require__) {
 
 	module.exports = __webpack_require__(1);
 
 
-/***/ },
+/***/ }),
 /* 1 */
-/***/ function(module, exports, __webpack_require__) {
+/***/ (function(module, exports, __webpack_require__) {
 
 	'use strict';
 
@@ -92,18 +92,18 @@
 
 	_reactDom2.default.render(_react2.default.createElement(_root2.default, null), document.getElementById('app'));
 
-/***/ },
+/***/ }),
 /* 2 */
-/***/ function(module, exports, __webpack_require__) {
+/***/ (function(module, exports, __webpack_require__) {
 
 	'use strict';
 
 	module.exports = __webpack_require__(3);
 
 
-/***/ },
+/***/ }),
 /* 3 */
-/***/ function(module, exports, __webpack_require__) {
+/***/ (function(module, exports, __webpack_require__) {
 
 	/**
 	 * Copyright 2013-present, Facebook, Inc.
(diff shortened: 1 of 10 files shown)


utsavdevsharma:
This is exactly what we need in advanced usage. I would like to request, it will be really helpful if we can merge #835 soon.

aij:
@martinek Please avoid committing changes to build outputs like docs-site/bundle.js. It creates needless conflicts.

Other than that, this looks good.

utsavdevsharma:
Big Thanks ! 👍 💯

Rachomir:
@martinek example in docs is not very clear for myself is there any chance you can explain to me how for example you would give a class depending on previous button click. 

Let say you have three buttons "red" "green" and "blue" and depending on chosen button you would give className to a perticular day - .react-datepicker__day-green or .react-datepicker__day-red or .react-datepicker__day-blue. Everytime you click on a different day you have an option to click a button which change color of that day in datepicker.

Thanks in advance

martinek:
Well, you would have to somehow build up and object where you would store which day is which color. Then you would use something like
```
dayClassName={date => dayColors[date]}
```

In dayColors you would have something along

```
dayColors = {
    '09-10-2017': 'red',
    '10-10-2017': 'blue'
}
```

or something like that

Rachomir:
@martinek first of all thanks for your help.

I tried to do it like you said unfortunately without success. Any other ideas? I thought maybe date format should be different but I tried many things and still doesn't work.

martinek:
Here is a working example that should help you out:
https://codesandbox.io/s/825l6m0zr2?module=%2FExample.js

Rachomir:
@martinek thanks a lot for your patience and apologies for beeing pain in the ass :)
Do you think it is possible to keep those "coloured dates" in database and render datepicker with days already coloured (accordingly to data fetched from db)? plus add new on click, like you did with state?

Rachomir:
I think I know how to do it. I will keep the data in db than on componentWillReceiveProps I will setState with date + colour. Em I right?

Rachomir:
My concern is also this bit 

`<style dangerouslySetInnerHTML={{__html: `
          .red { color: red }
          .green { color: green }
          .blue { color: blue }
        `}} />` 

How would I use it if I fetch data from DB? instead of clicking

Rachomir:
@martinek would you be able to help me?

martinek:
The example linked intended to show you how the prop on DatePicker works. I'm positive you can keep the data from state in database. I don't know what you mean with new click.

Also if you are designing you own application, you don't need `<style dangerouslySetInnerHTML...>`. That was something i used just for the example.

I'm not sure what you expect of me? You are apparently creating some project and without whole picture it is hard to recommend any solution. Also it is kind of out of scope for pull request discussion.

You might get more info on stackoverflow.

PR #839, opened 2017-04-23
Select time feature

shadrech:
Feature allows users to select a time along with the date. So moment object returned will have specific time
Also includes time filtering with minTime, maxTime and excludeTimes


shadrech:
Will work on tests at the weekend. Will creating tests be all thats needed to fix the codecov errors?

martijnrusschen:
@shadrech Thanks, yes. Codecov only complains about the fact that the current test coverage of this PR is below the target of 75%. Currently the overall coverage floats around 95% for the complete project.

docs/datepicker.md
```diff
@@ -313,3 +313,30 @@ type: `string`
 type: `bool`
 defaultValue: `false`
 
+
```
martijnrusschen: You can remove this submissions. This will be autogenerated when releasing a new version.

server.js
```diff
@@ -22,11 +22,11 @@ app.use(require('webpack-hot-middleware')(compiler))
 
 app.use(express.static('docs-site'))
 
-app.listen(8080, 'localhost', function (err) {
+app.listen(9000, 'localhost', function (err) {
```
martijnrusschen: Do we need this change?

src/datepicker.jsx
```diff
@@ -75,12 +75,17 @@ export default class DatePicker extends React.Component {
...
+    timeIntervals: PropTypes.number,
+    minTime: PropTypes.object,
+    maxTime: PropTypes.object,
+    excludeTimes: PropTypes.array
   }
 
   static get defaultProps () {
     return {
-      dateFormat: 'L',
+      dateFormat: 'LLL',
```
martijnrusschen: Why's this changed?

server.js
shadrech: Sorry forgot to change that back. That port was occupied with a different process when I was developing the datepicker. That change isn't necessary at all

src/datepicker.jsx
shadrech: "LLL" is more verbose. It displays the time as well. Maybe I should set a if statement if this.props.showTimeSelect is instantiated before changing the format to display time as well

spiroski:
Any progress on this issue?

shadrech:
My bad. Been really busy at work. Will work on tests this weekend hopefully 🤞🏾

shadrech:
@martijnrusschen Sorry for hiatus. Added tests but codecov still not happy. Please advise on best way forward? What am I missing here?

docs-site/style.css
```diff
@@ -84,6 +84,8 @@
   border-top-right-radius: 0.3rem;
```
martijnrusschen: Can you please exclude changes to this file from the PR? This will be automatically updated when I'm going to run a new release.

docs-site/bundle.js
```diff
@@ -1116,7 +1116,7 @@
```
martijnrusschen: Can you please exclude changes to this file from the PR? This will be automatically updated when I'm going to run a new release.

docs/datepicker.md
```diff
@@ -68,4 +68,4 @@ General datepicker component.
 |`value`|`string`|||
```
martijnrusschen: Can you please exclude changes to this file from the PR? This will be automatically updated when I'm going to run a new release.

shadrech:
Yh I can add this to README. I had just included examples of how to use the new feature in docs-site/src/examples (exclude_time_period.jsx, exclude_times.jsx, show_time.jsx). But yh I will add some info in README

shadrech:
What of the tests? @martijnrusschen

martijnrusschen:
Not sure why codecov isn't liking them. Need to look into that, but for now this seems fine. @aij Any thougths on this PR?

martijnrusschen:
@alex-shamshurin ^

docs/datepicker.md
shadrech: I'm pretty sure I didn't edit these files. They must have been edited when running `yarn start` and editing files in src/ directory

docs-site/src/examples/exclude_times.jsx
```diff
@@ -0,0 +1,39 @@
...
+        <strong>{'  dateFormat="LLL"'}</strong><br />
+        {'/>'}
+        </code>
+      </pre>
+      <div className="column">
+        <DatePicker
+            selected={this.state.startDate}
+            onChange={this.handleChange}
+            showTimeSelect
+            excludeTimes={[moment().hours(17).minutes(0), moment().hours(18).minutes(30), moment().hours(19).minutes(30), moment().hours(17).minutes(30)]}
```
aij: Making this a function/predicate instead of an array might be more flexible, especially if whole time ranges need to be excluded, or if the exclusions need to depend on the date. (Eg, if it is based on available appointment times from a DB, or varies with DST.)

If we expect a fixed/predefined list to be much more common this could be easier to use though. Thoughts?

docs-site/style.css
```diff
@@ -142,6 +146,51 @@
...
+    display: inline;
+    border: 1px solid #aeaeae;
+    border-radius: .3rem;
+    position: absolute;
+    right: -72px;
+    top: 0; }
+  .react-datepicker__time-container .react-datepicker__time {
+    position: relative;
+    background: white; }
+    .react-datepicker__time-container .react-datepicker__time .react-datepicker__time-box {
```
aij: Indentation here and below.

Also seems like a lot of special casing.

docs/datepicker.md
aij: @shadrech Yes, they're build outputs. 

@martijnrusschen Is there a way we could improve the release process so everyone doesn't need to manually avoid committing these sorts of changes?

docs-site/src/example_components.jsx
```diff
@@ -193,8 +195,16 @@ export default class exampleComponents extends React.Component {
     component: <RawChange/>
   },
   {
-    title: 'Don\'t hide calendar on date selection',
-    component: <DontCloseOnSelect/>
```
aij: Why was this example removed?

test/exclude_time_period_test.js
```diff
@@ -0,0 +1,32 @@
...
+    var datePicker = mount(
+      <DatePicker
+          showTimeSelect
+          selected={moment()}
+          onChange={() => null}
+          minTime={moment().hours(17).minutes(0)}
+          maxTime={moment().hours(18).minutes(0)} />
+    )
+    var times = datePicker.find('li.react-datepicker__time-list-item')
+    expect(times).to.exist
```
aij: Wouldn't this test pass even if minTime and maxTime are ignored?

src/date_utils.js
```diff
@@ -31,6 +31,30 @@ export function isDayDisabled (day, { minDate, maxDate, excludeDates, includeDat
...
+      return true
+    }
+  }
+
+  return false
+}
+
+export function isTimeInDisabledRange (time, { minTime, maxTime }) {
+  if (!minTime || !maxTime) {
+    throw new Error('Both minTime and maxTime props required')
```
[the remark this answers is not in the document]
shadrech: Only applicable if user decides to have a time range. Both values required to create time window. Since time loops at 24 (24 hr clock), setting a lower bound with no upper bound for example won't create a finite window (not good at explaining, hopefully that makes sense)

docs/datepicker.md
shadrech: @aij Add files ending in .js/.md/.css in directories docs/ & docs-site/ to .gitignore? Since files are auto generated

martijnrusschen: Problem is that we need the .js and .css files in the gh-pages branch for the docs. It seems pretty hard to ignore something in one branch but have it checked in into the other branch. I read something about it here: https://stackoverflow.com/questions/1836742/using-git-how-do-i-ignore-a-file-in-one-branch-but-have-it-committed-in-another but haven't found the silver bullet yet. Ideas are welcome!

We can't ignore the .md files since those are served in the docs folder in the master branch.

src/tether_component.jsx
```diff
@@ -0,0 +1,149 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import ReactDOM from 'react-dom'
+import Tether from 'tether'
```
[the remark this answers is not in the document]
martijnrusschen: Yeah seems an artifact of recent changes. We don't use Tether anymore.

docs-site/style.css
martijnrusschen: Auto generated file, should be removed from this PR.

shadrech:
Will work on requests at the weekend

docs-site/src/example_components.jsx
shadrech: My bad will put it back

docs-site/src/examples/exclude_times.jsx
shadrech: I find an array to be more simpler. I imagine a scenario where an api call returns a json array of moment.toISOString strings, and all client has to do is convert these to moment() objects and plug the array into excludeTimes. Even maybe extend excludeTime that first a check is made to see if objects received are strings, and automatically convert to moment() objects. That way client can just pass the json array directly to excludeTimes via a simple JSON.parse?

docs-site/style.css
shadrech: Yh most of it removed. Main function of special casing was for sizing the time component. Now thats done by calculating the month container height via clientHeight (passed via refs)

src/time.jsx
```diff
@@ -0,0 +1,114 @@
...
+      return
+    }
+
+    this.props.onTimeChange(time)
+  }
+
+  ulClasses = () => {
+    let classes = ['react-datepicker__time-list']
+
+    if (this.props.showYearDropdown || this.props.showMonthDropdown) {
```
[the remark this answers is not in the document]
shadrech: Removed. Used to calculate height. Like mentioned above, now its calculated via month container clientHeight

test/exclude_time_period_test.js
shadrech: Lol yh my bad. At first I wanted to test the number of .react-datepicker__time-list-item they'll be; there should be 3 with configuration above. I tried testing with expect(times).to.have.length(3) but kept getting a fail as length always comes back as 0. Any ideas what I'm missing here? Maybe theres another test I can do that'll achieve desired result

src/tether_component.jsx
```diff
@@ -0,0 +1,149 @@
+import React from 'react'
```
martijnrusschen: Please remove the entire file. This isn't used.

shadrech: 👍

dola:
Overall looking pretty cool. I'm looking forward to using this!
However I would advise to only merge to master when the feature is due to be released. Otherwise the master Readme shows documentation for a feature that is not yet available. (confuses new users)

actraiser:
Thanks @dola - that was exactly the case for me. I was wondering why adding _showTimeSelect_ to the props did not show up anything  when I tested react-datepicker 0.54.0 for the first time.

actraiser:
PS: The just released 0.55.0 now behaves as expected.  Thx!

martijnrusschen:
Sorry for the confusion, forgot to release this earlier.

PR #840, opened 2017-04-25
Added highlightDates2 and highlightDates3 ranges

alex-shamshurin:
Sometimes we need diffrent highlighted date ranges with diffrent styles. I added `highlightDates2` and `highlightDates2` props. It can be very useful to apply different styles.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -31,6 +31,8 @@ export default class Calendar extends React.Component {
     filterDate: PropTypes.func,
     fixedHeight: PropTypes.bool,
     highlightDates: PropTypes.array,
+    highlightDates2: PropTypes.array,
+    highlightDates3: PropTypes.array,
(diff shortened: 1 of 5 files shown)


alex-shamshurin:
Ok, I think it might be an array. I'll do it.

rafeememon:
Cool, keep in mind that the API will need to be backwards-compatible

alex-shamshurin:
Yep, `this.highlightedDates` may be an array of `moment` or multidimensional array of `moment`s

rafeememon:
That could work; it's not obvious how you might specify the class name for the additional date ranges in a sane way, but maybe you can find something that works well =)

alex-shamshurin:
I think `className + arrayIndex`, like ''react-datepicker__day--highlighted2''. Beside this, have no idea how to make it better.

PR #844, opened 2017-04-27
contain day in element

baoist:


baoist:
Sorry about this, it was unintentional. The base was auto-set to the source's master and I didn't catch it in time for me spamming 'open'.

PR #859, opened 2017-05-14
HighlighedDates prop now can be an array of object with class names as a key

alex-shamshurin:
1) In our project we need some date ranges with different color highlighting. For example  - last week, current week, next week with different colors. This PR solve it. 

2) And another change - annoying `bundle.js` and `style.css` were removed from git index according to .gitignore files.

Here is an example how to use highlighted ranges with custom class:
```
this.highlightWithRanges = [
      { 'react-datepicker__day--highlighted-custom-1': [
        moment().subtract(4, 'days'),
        moment().subtract(3, 'days'),
        moment().subtract(2, 'days'),
        moment().subtract(1, 'days') ]
      },
      { 'react-datepicker__day--highlighted-custom-2': [
        moment().add(1, 'days'),
        moment().add(2, 'days'),
        moment().add(3, 'days'),
        moment().add(4, 'days') ]
      }
    ]
```
 And inside render()

```
<DatePicker 
  onChange={this.handleChange}
  highlightDates={this.highlightWithRanges}
  placeholderText="This highlight two ranges with custom classes" />
```
The previous behavior (array of moments) is supported.


alex-shamshurin:
So?

martijnrusschen:
Looks good! Can you resolve your merge conflict so I can merge this?

alex-shamshurin:
Conflict is with only these two build files. They must be excluded from git. If you have access permissions you can resolve it right here.

martijnrusschen:
@alex-shamshurin I'm not completely sure how to fix these build files. We'll need them in the gh-pages branch, but it doesn't seem possible to exclude files just from 1 branch.

alex-shamshurin:
Why we cannot it exclude from this branch? On each build these files will be recreated. Yes, it prevents from easy switching to the branch which tracks these files, but we can delete these files ant then checkout.

alex-shamshurin:
It ir really hard to work with build files in git index. It's terrible. Best solution exclude them from all branches.

martijnrusschen:
> Why we cannot it exclude from this branch? On each build these files will be recreated. Yes, it prevents from easy switching to the branch which tracks these files, but we can delete these files ant then checkout.

It's currently ignored from the master branch and all feature branches. The problems is that to completely fix it, we need to exclude it from gh-pages as well. But that means we need to fix our deployment of the demo site in a different way. Open to suggestions, but it seems unrelated to this PR.

alex-shamshurin:
Hmm, ok, got it. Do you mean I should include these two files built with all my changes?

martijnrusschen:
Yeah, it's a bit complicated but if you take out the compiled files from the PR those will be re-generated when doing a new release.

MHultman:
+1
I'd love to see this in the next release. Have the exact same need!

alex-shamshurin:
I have updated build files, so please merge it!

Issue #736, opened 2017-02-13
`openToDate` not working in `v0.41.1`


yp:
Currently the prop `openToDate` does not seem to work in `v0.41.1`.
See, for example, https://hacker0x01.github.io/react-datepicker/#example-25

In my opinion, the bug is an unintended consequence of commit 418b8d33594b72f97f05c2ba37558305a969ad24 which introduced the `preSelection` prop in `Calendar` (https://github.com/Hacker0x01/react-datepicker/commit/418b8d33594b72f97f05c2ba37558305a969ad24#diff-53b2cf631ea5d028d238854eaa70ce16R107) that is filled with today if no date is `selected` hence `if (initialDate)` is always true.

mgrdevport:
I'm working on a PR.
I think we can remove the openToDate prop on the calendar due to the preSelection prop and move the validation, handling and test cases to the datepicker class.

yp:
Thanks for looking into this.
Are you planning to remove `openToDate`?
I found it really convenient for showing a (multi-month) inline calendar which stylistically matches the datepickers used in other places of the application. Actually, I am directly using the `Calendar` component now to show the inline calendar without using the `DatePicker` with the inline prop set to true. Is this use "supported" (potentially, at least) or discouraged?

shogan94:
Wanted to know if there was any word on `openToDate` or a work around to get that functionality? I see that setting `selected` works but I don't want to have the field pre-selected. Thanks

GautierT:
Any news on the fix for `openToDate` ? Thanks !

shogan94:
I found a quick fix (for now) just downgrading the version to `v. 0.40.0`

GautierT:
aha yes, i just did that too ! ;)

martijnrusschen:
Hmm, it seems 0.41 introduced some regressions that weren't in 0.40. Hopefully someone is able to find the root cause, so this can be fixed.

donysukardi:
Found out the issue.

In datepicker.jsx, [`preSelection: this.props.selected ? moment(this.props.selected) : moment()`](https://github.com/Hacker0x01/react-datepicker/blob/master/src/datepicker.jsx#L109). `preSelection` is defaulted to `moment()` and hence initialDate will always be set in [`getDateInView`](https://github.com/Hacker0x01/react-datepicker/blob/master/src/calendar.jsx#L102) in calendar.jsx.

[...]

PR #860, opened 2017-05-14
'openToDate' prop is now take a precedence over 'selected' value

alex-shamshurin:
I think that current behavior when `selected` and `openToDate` both specified and `selected ` take a precedence over `openToDate` is totally incorrect. Every time calendar is open it must be opened at `openToDate` if specified regardless of other values. It's should be treated as a constant value at which calendar will open every time. It's not a selection of any kind, it's just a page on which calendar will be opened.
See this issue #736 and my arguments for and some use cases.


alex-shamshurin:
It was two weeks ago! Any comments?

alex-shamshurin:
@martijnrusschen Please check it out and merge.

PR #863, opened 2017-05-15
Remove Clickoutside

malectro:
NOTE: This PR is incomplete and more of a proof of concept.

Clickoutside is not native and does not play nice with any clicks that invoke `stopPropagation`. This is especially apparent when using `react-select` and `react-datepicker` on the same page.

Instead of refocusing after every blur of the datepicker input and only hiding the calendar `onClickOutside`, we can hide the calendar when the datepicker is blurred except in the case where the blur is caused by a calendar `mousedown` where we set a flag that tells the datepicker to refocus and continue to show the calendar.


malectro:
oops, i meant to open this on my own fork

PR #866, opened 2017-05-17
Better Popover Triangle Orientation

malectro:
This allows the triangle to be positioned relative to the right side of the popover if the user passes `popoverAttachment="right"` as a prop.

### DIFF
--- a/docs-site/src/examples/placement.jsx
+++ b/docs-site/src/examples/placement.jsx
@@ -23,8 +23,8 @@ export default class Placement extends React.Component {
 <DatePicker
     selected={this.state.startDate}
     onChange={this.handleChange}
-    popoverAttachment="bottom center"
-    popoverTargetAttachment="top center"
+    popoverAttachment="bottom right"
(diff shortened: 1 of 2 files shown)


martijnrusschen:
Thanks!

malectro:
Great lib. Glad to help out!

lukastong:
@malectro i used `popoverAttachment="right"` as a prop, but threw error `Failed prop type: Invalid prop `attachment` of value `right` supplied to `TetherComponent`, expected one of ["top left","top center","top right","middle left","middle center","middle right","bottom left","bottom center","bottom right"].`
i m using version 0.29.0(the latest version download from npm)
did V.0.51.0 published on npm?

martijnrusschen:
According to https://www.npmjs.com/package/react-datepicker 0.51.0 is the latest published version.

lukastong:
@martijnrusschen thx~ but why can't i V0.51.0? even if i use `npm install react-datepicker@0.51.0 --save`...

PR #881, opened 2017-06-09
updating doc to show all props in a table form

### LINKED ISSUE #880: update datepicker.md for better readibility
having a table format for the props makes it easier to read/comprehend then a linear format.

hozefaj:
fixes #880 

below is how the updated doc page will looks. _(image does not capture all props....but the docs has them)_

screen shot 2017-06-09 at 4 13 59 pm

### DIFF
--- a/docs/datepicker.md
+++ b/docs/datepicker.md
@@ -1,315 +1,61 @@
-`datepicker` (component)
-========================
-
-General datepicker component.
-
-Props
------
-
-### `autoComplete`
-
-type: `string`
-
-
-### `autoFocus`
-
-type: `bool`
-
-
-### `calendarClassName`
-
-type: `string`
-
-
-### `children`
-
-type: `node`
-
-
-### `className`
-
-type: `string`
-
-
-### `customInput`
-
-type: `element`
-
-
-### `dateFormat`
-
-type: `union(string|array)`
-defaultValue: `'L'`
-
-
-### `dateFormatCalendar`
-
-type: `string`
-defaultValue: `'MMMM YYYY'`
-
-
-### `disabled`
-
-type: `bool`
-defaultValue: `false`
-
-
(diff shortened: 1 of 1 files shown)


rafeememon:
Hey @hozefaj, thanks for the work here, but unfortunately the API docs are autogenerated, so this will be wiped out the next time a release is made.  I think a proper fix here needs to involve changing the autogeneration process.

hozefaj:
Oh man. Can you point me to the auto generation process. I can maybe modify the code to generate a table then.

hozefaj:
Thats 30min of work down the drain.... :(

rafeememon:
I'm not particularly familiar with it, but the script is https://github.com/Hacker0x01/react-datepicker/blob/master/scripts/generateMarkdown.sh. It looks like it's written by Facebook and I'd generally prefer to avoid a custom solution, but maybe there is another doc generator that looks better.

hozefaj:
Let me have a look. In general I also prefer against custom stuff. But here when I was reviewing docs I realized they can be much better in the form of a table.

martijnrusschen:
If you're able to modify the scripts, I'm happy to take a look. We use react-docgen from Facebook to generate docs based on the proptypes.

hozefaj:
Basically this https://github.com/Hacker0x01/react-datepicker/blob/master/scripts/generateMarkdown.sh#L47-L56 LOC needs to be updated to create a tabular format.

Only thing is that, if there is a way to add some description for some of the props. That would be useful.

martijnrusschen:
> Only thing is that, if there is a way to add some description for some of the props. That would be useful.

See https://github.com/reactjs/react-docgen#example on how to add additional information to a prop.

hozefaj:
@martijnrusschen @rafeememon Created #898 to update `generateMarkdown.sh`...

PR #898, opened 2017-06-19
updating doc generation script to create props as table

### LINKED ISSUE #880: update datepicker.md for better readibility

hozefaj:
fixes #880 

Currently the `generateMarkdown.sh` creates a name-value pair for the props. This is not very readable since the page can get long. eg: https://github.com/Hacker0x01/react-datepicker/blob/master/docs/datepicker.md

updating the script, changes it to a table layout which makes the props docs more readable. eg: https://github.com/hozefaj/react-datepicker/blob/generate-docs/docs/datepicker.md

I have not generated the new docs, since this will done during the release process.

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -96,12 +96,7 @@ export default class DatePicker extends React.Component {
       popoverAttachment: 'top left',
       popoverTargetAttachment: 'bottom left',
       popoverTargetOffset: '10px 0',
-      tetherConstraints: [
-        {
-          to: 'window',
-          attachment: 'together'
-        }
-      ],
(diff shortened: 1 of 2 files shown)


src/datepicker.jsx
```diff
@@ -96,12 +96,7 @@ export default class DatePicker extends React.Component {
       popoverAttachment: 'top left',
       popoverTargetAttachment: 'bottom left',
       popoverTargetOffset: '10px 0',
-      tetherConstraints: [
```
hozefaj: since cannot it a multiline code into the table cell, converted this into single line.

martijnrusschen:
Nice work!

PR #901, opened 2017-06-21
Update datepicker.md

kolupaevri:
Fix markup for "|" character

### DIFF
--- a/docs/datepicker.md
+++ b/docs/datepicker.md
@@ -12,11 +12,11 @@ General datepicker component.
 |`children`|`node`|||
 |`className`|`string`|||
 |`customInput`|`element`|||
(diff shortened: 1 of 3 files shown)


martijnrusschen:
These docs are autogenerated by this file: https://github.com/Hacker0x01/react-datepicker/blob/master/scripts/generateMarkdown.sh. We should try to fix the problem here, to keep your fixes after the next release :)

kolupaevri:
Ok.
If I get it correctly, problem in this line.
https://github.com/Hacker0x01/react-datepicker/blob/master/scripts/generateMarkdown.sh#L36

martijnrusschen:
That should work.

Issue #885, opened 2017-06-13
Weird focus-event related problem after upgraded to 0.47.0


philiptzou:
Hi everyone. I just encountered a weird focus-event related problem after I upgraded the `react-datepicker` to 0.47.0. I tested the following steps on both Chrome and Firefox and it seems both browsers get the same problem. Here are the simple steps to reproduce the problem:

1. Open https://hacker0x01.github.io/react-datepicker/
2. Scroll down to "Default" example. Left click the datepicker input box, then select a random date.
3. Press "tab" key on the keyboard.

Then you will see the browser enters a dead loop and keeps switching between the two datepicker input box.

The problem is gone after downgraded to 0.46.0.

Any idea why this problem happened? I checked the [comparison between 0.46.0 and 0.47.0](https://github.com/Hacker0x01/react-datepicker/compare/v0.46.0...v0.47.0) but didn't find anything interesting.

alex-shamshurin:
Really weird. Looks like with wrong `ref` or `key` somewhere.

conlanpatrek:
I don't know that it's an incorrect ref...

It looks like an issue the `deferFocusInput` call (https://github.com/Hacker0x01/react-datepicker/blob/master/src/datepicker.jsx#L175). When you blur from input A (with an open datepicker) by focusing on input B, you set a timeout that refocuses input A, but by the time that cb gets called, input B's datepicker is popped open. So now, when input B blurs, it will call `deferFocusInput` which primes it to steal focus back from input A. Wash, rinse, repeat until infinity.

My guess is that this never got noticed before because you're also closing the date picker on mouse down somewhere (probably in `react-onclickoutside`), so clicking into a second datepicker doesn't trigger the loop.

conlanpatrek:
I should also mention that I can recreate the issue with 0.46.0

philiptzou:
@conlanpatrek Really? 0.46.0 works fine on my computers. Are there any different steps you used?

philiptzou:
I think I found the reason.

In version 0.46.0, [`datepicker.jsx`](https://github.com/Hacker0x01/react-datepicker/blob/v0.46.0/src/datepicker.jsx#L265-L267) has following three lines:

```javascript
      if (/^Arrow/.test(event.key)) {
        this.onInputClick()
      }
```

[...]

PR #904, opened 2017-06-21
Fix re-focusing problem related with usage of `Tab` (#885)

philiptzou:
This pull-request addressed the solution I proposed in #885.

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -258,23 +258,26 @@ export default class DatePicker extends React.Component {
   }
 
   onInputKeyDown = (event) => {
+    const eventKey = event.key
     if (!this.state.open && !this.props.inline) {
-      this.onInputClick()
+      if (eventKey !== 'Enter' && eventKey !== 'Escape' && eventKey !== 'Tab') {
+        this.onInputClick()
+      }
       return
     }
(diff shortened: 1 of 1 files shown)


src/datepicker.jsx
```diff
@@ -259,7 +259,9 @@ export default class DatePicker extends React.Component {
 
   onInputKeyDown = (event) => {
     if (!this.state.open && !this.props.inline) {
-      this.onInputClick()
+      if (!(/^(Tab|Enter)$/.test(event.key))) {
```
rafeememon: Can we avoid a regex here?  It adds some unnecessary obfuscation

alex-shamshurin: I agree, no regex here, overhead.

philiptzou: Well I guess it's true. I'll made the change.

philiptzou:
Addressed the code review comment and also prevented to re-open the calendar when pressed "ESC" twice.

cattermo:
Please release this code. This bug is stopping me from upgrading to React 15.6 :-(
@rafeememon @martijnrusschen

martijnrusschen:
I’ll release this as soon as I get back from my vacation. Probably Monday.

Issue #905, opened 2017-06-22
Replace `tether` with `popper` package for calendar popups


tohjustin:
This is a follow up from the discussion at #899.

Since the project status for [Tether](https://github.com/HubSpot/tether/) is [up in the air](https://github.com/HubSpot/tether#rotating_light-project-status-rotating_light), [Popper.js](https://popper.js.org/) seems like a better alternative & even the maintainers of Tether recommended it.

tohjustin:
@alex-shamshurin @martijnrusschen @rafeememon Would be great if you guy can assign this issue to me =)

rafeememon:
I don't think we use assignments in this repo but you're free to start working on this!

FezVrasta:
Take a look at react-popper for a battle tested React integration with Popper.js

tohjustin:
@FezVrasta Thanks for the heads up, will look at it over the weekend!

alex-shamshurin:
Waiting for PR to test it!

tohjustin:
@alex-shamshurin @rafeememon Just a quick question, do you think we should use `popper.js` directly or just use `react-popper`? Cause I'm a bit skeptical about how 'battle tested' this wrapper module is when there's no form of testing implemented =/

Also, let's say we're using `react-popper` and there's some bug in `popper.js` or `react-popper` itself, won't we have to wait for the guys w/ the `react-popper` to fix the stuff before we can update our dependencies & ship the fix on our end?

FezVrasta:
I use react-popper in production and it works wonderfully.

I know there's an issue open to add tests, rather than create a new project consider add tests to the existing one maybe?

tohjustin:
@FezVrasta Thanks for the quick reply! Didn't realize that you're also the creator of `popper.js`, I guess I shouldn't be worried about this issue then xD

tohjustin:
@alex-shamshurin @rafeememon `Popper.js` uses a different way of specifying the behavior of the popups & there's some differences in functionality...

Do you guys intend to maintain the same API & all tether related props? (`classPrefix`, `attachment`, `targetAttachment`, `targetOffset`, `renderElementTo`, `constraints`)

PR #910, opened 2017-06-28
Migrate from tether.io to popper.js for handling calendar popups

tohjustin:
This is a follow up from issue #905 

Since the project status for Tether is up in the air, Popper.js seems like a better alternative & even the maintainers of Tether recommended it.


src/datepicker.jsx
```diff
@@ -55,11 +55,9 @@ export default class DatePicker extends React.Component {
     openToDate: PropTypes.object,
     peekNextMonth: PropTypes.bool,
     placeholderText: PropTypes.string,
-    popoverAttachment: PropTypes.string,
-    popoverTargetAttachment: PropTypes.string,
+    placement: PropTypes.string,
```
aij: Minor nitpick: This should use `PropTypes.oneOf` (like `PopperComponent`) to avoid misleading warnings when an invalid string is passed in. (The old code has the same issue.)

tohjustin:
@aij Thanks for reviewing, glad to contribute! =) 

Right now this is just a rough draft, still waiting for you guys to chip in about the types of props that we want to expose in our `<datepicker />` component. (So that the devs can customize the popper's behavior according to their use cases)

I'm currently thinking of providing the following props:
- placement (position of popper)
- popperClass (will be attached to the root div of the popper component)
- modifier (obj to configure popper behavior, see popper.js docs)

martijnrusschen:
I’m not able to review this PR due to my vacation. If you guys feel this is good to merge, go ahead and merge it as soon as you feel it’s a good replacement for what we have today.

tohjustin:
@aij I'm not really familiar with how codecov measure coverage, any idea on how to fix that `codecov/patch — 60% of diff hit (target 85.33%)`?

I believe this commit is ready for a formal review =)

aij:
I think this should be suitable for 99% of uses, though users currently passing tether-specific props will need to update. The removed `renderCalendarTo` shouldn't be needed as much (or at all?), since the calendar now renders more naturally within the DOM rather than at `<body>`.

If anyone does still need `renderCalendarTo`, I expect we can come up with a workaround.

willbuckingham:
Had a problem somewhere between 0.47 and 0.52 that I believe might have been caused by some of these changes.  Our datepicker was in a container that was overflow:scroll, and in the old incarnation it was fine because because the popup got appended to body, but now it stays inside the container and was affected by the overflow.  Now I can probably removed the overflow scroll, or perhaps I can change some options from: https://hacker0x01.github.io/react-datepicker/#example-20

rmdavenp:
^I second this problem

rafeememon:
@willbuckingham, @rmdavenp, if there is a problem please file a bug and/or submit a PR with a fix

willbuckingham:
Sorry, was paying enough attention kinda thought I was tagging on another issue. Not a PR. Will file correctly. Thanks.

Issue #920, opened 2017-07-03
Highlighting of "today" date not always correct when using utcOffset


greg-a-smith:
When passing a `utcOffset` prop, the date for "today" is not always identified correctly.  This occurs when no date value is passed (empty field) and it only applies to adding the `react-datepicker__day--today` class to the appropriate day on the calendar.  It will highlight the day that corresponds to the current local time of the browser.  Clicking on the "Today" button at the bottom does select the proper date even if a different one is highlighted in the calendar.  Once a date has been selected, the highlighting of "today" seems to work properly.

To reproduce:

1. Go to the examples/demo page and find the TimeZone example.
2. Select a value from the Timezone Offset dropdown that would match the same date as your local system.
3. Open calendar and check which calendar day receives the `react-datepicker__day--today` class (bold text) -- it is the date that matches your local system time.
4. Select a value from the Timezone Offset dropdown that would change the date forward a backward a day from your local system.
5. Open calendar and check which calendar day receives the `react-datepicker__day--today` class (bold text).  Even though "today" should have changed, it still shows the previous date highlighted -- the date that matches your local system time.
6. Click on the "Today" link/button at the bottom of the calendar and verify it selected the proper date.
7. Open the calendar and now the highlighted "today" date shows the proper date -- based on `utcOffset`.

### Video
![datepicker today issue](https://user-images.githubusercontent.com/11407353/27805973-91ae4cbe-5ffd-11e7-8ae8-ab5b85d97862.gif)

#### Video Annotation

_NOTE: This was recorded on July 3rd on a system that is GMT-5:00 at approximately 19:30 GMT._

The first example selects a `utcOffset` of -4 which is correctly shown highlighting July 3rd and also selects July 3rd when clicking the "Today" link/button.

The second example selects a `utcOffset` of +8 which should be showing July 4th, but still shows July 3rd highlighted.  When clicking the "Today" link/button, July 4th is correctly selected.

PR #921, opened 2017-07-03
Take utcOffset into account when creating a defaultPreSelection date

greg-a-smith:
Fixes issue #920 .


greg-a-smith:
Since the issue depends on a local environment's current date/time, I'm not quite sure how to replicate that in a unit test.  Passing a date in doesn't work because the fix is only hit when no date is passed in.

PR #933, opened 2017-07-10
Adding a 'weekLabel' property to replace the hash in the calendar view.

sentient06:
Example:

    <DatePicker
        showWeekNumbers
        weekLabel="Wks"
    />


test/calendar_test.js
```diff
@@ -183,6 +183,12 @@ describe('Calendar', function () {
     expect(calendar.state().date.isSame(todayInAuckland, 'day'))
   })
 
+  it('should set custom week label if weekLabel is provided', () => {
```
martijnrusschen: Might be good to check if there's a # displayed if there's no weeklabel prop is passed.

sentient06: Sure, I'll add a new test above that one.

PR #935, opened 2017-07-10
Changing Enter behaviour to avoid select current date

buhuitou:


buhuitou:
Sorry. This is a mistake to open PR here.

Issue #843, opened 2017-04-26
After update from version 0.40.0 calendar not update selected date


AntonovAv:
Calendar display old selected date  if i put new `selected` prop to calendar.
```
<DatePicker
        ref={this.onDatePickerRef}
        openToDate={this.state.openDate}
        onMonthChange={this.onMonthChange}
        customInput={<SelectedDate/>}
        popoverTargetOffset='0 0'
        dateFormat={isSelectedCurrentYear ? DATE_FORMAT_FOR_CURRENT_YEAR : DATE_FORMAT}
        dateFormatCalendar='MMMM'
        {...this.props}
        selected={this.state.selectedDate}
        onChange={this.onSelectedDateChange}
        filter={this.datesFilter}
        forceShowMonthNavigation={true}
      >
        <YearsSelector onChange={this.onYearChange} year={this.state.openDate.year()}/>
      </DatePicker>
```
prev selected date 01 april
![image](https://cloud.githubusercontent.com/assets/11208532/25433020/917e8dce-2a8f-11e7-8c0a-9d7f8e7cfb21.png)

prev selected date 30 april
![image](https://cloud.githubusercontent.com/assets/11208532/25433040/acbc6728-2a8f-11e7-94d5-43b91fe40e15.png)

How you seen. On filter prop calendar reacts and disable days

JorgenEvens:
We are seeing something similar in our application.

malyfko:
@JorgenEvens Thanx, you saved me life! 😃

alex-shamshurin:
`OpenToDate` property always take precedence over `selected` value. Does not specify `openToDate` 

> The calendar switches back to the old date

it's normal behaviour, calendar will scroll on `openToDate`

malyfko:
@alex-shamshurin I'm using 'inline' option, `OpenToDate` shouldn't even be considered

alex-shamshurin:
if so, do not use it, otherwise it will be scrolled to `openToDate` value.

JorgenEvens:
@alex-shamshurin So what you are suggesting is that the behaviour on the [inline example](https://hacker0x01.github.io/react-datepicker/#example-25) is intended behaviour?

To be clear I was literally using the component as specified earlier (`<DatePicker inline={true} selected={selectedDate} />`), not setting `openToDate` to any value at all when I ran into this issue.

[...]

Issue #743, opened 2017-02-21
inline mode does not redisplay properly


MaloneyT:
In the latest release (0.41.1) in inline mode (\<DatePicker inline />) the calendar does not refresh properly after multiple date selections.

Steps to reproduce bug:
1.  open datePicker and click on a day:  the day is shown as selected
2.  move to another month (forwards or backwards) and click on a day:  the day clicked is NOT highlighted as expected

version 0.40.0 works properly

martijnrusschen:
Hmm, it seems 0.41 introduced some regressions that weren't in 0.40. Hopefully someone is able to find the root cause, so this can be fixed.

Amauris:
I created a pull request, although I'm honestly I'm not entirely sure if it's a proper fix: https://github.com/Hacker0x01/react-datepicker/pull/825

raurfang:
So, to add a bit more detail:
It looks like this bug was introduced in #673.

This:
https://github.com/Hacker0x01/react-datepicker/pull/673/files#diff-f662ae56197d5c888897a02b982f0ad3R119
causes problems when called from here:
https://github.com/github1/react-datepicker/blob/2444feea4f5d940f0de96a52fb329554c27399e3/src/datepicker.jsx#L164

The problem is caused by the `setSelected()` state changes not being applied before the `setOpen()` call is made. `setOpen()` uses `this.state` which still has the old `preSelection` value, because React's `setState()` (called in `setSelected()`) is async.

Another way to fix this is by calling `setOpen` inside a `setState` callback:
```
this.setState({}, () => this.setOpen(false))
```
This makes sure that the `setOpen()` call is made after all the previous `setState` calls finish updating the state.

martijnrusschen:
Should be fixed now

PR #941, opened 2017-07-11
Fix inline datepicker month switch bug

### LINKED ISSUE #877: Previous Month on Date Select
On inline calendar, if I follow the following steps: 
1. Advance to the next month (July) and select a date (Jul 15)
2. Advance to next month (August) and select a new date (Aug 15)

I expect to have Aug 15 highlighted, but the calendar navigates to the previous month (July). Aug 15 is selected in data and is highlight if I navigate back to August. The datepicker erroneously navigates to the previous month. Appears to be a propagation or click handling issue, but I haven't narrowed down where it is happening.

### LINKED ISSUE #927: Datepicker don't switch to new range in inline mode
Hi! I faced with a problem that I use DatePicker across the site and in one place I need it just for showing a quarter of year like the calendar.
Steps to reproduce:
1. Add inline datepicker
2. Change selected date (or min max dates like in my example)
3. Expect to switch calendar to the new range but it works just after click somewhere on the page.

I try to find own solution and I think it relates to #743 
But it strange that after click somewhere it works fine. Could I somehow forceUpdate datepicker to implicitly call `setOpen()` or whatever what makes it work :)

Here is the link to screencast of issue -> http://take.ms/oSAc8

I show the debugger statement that doesn't fire after the last click.

Thank you.

Odrin:
Fix #927 #877 #843 #743 etc.

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -207,8 +207,8 @@ export default class DatePicker extends React.Component {
         return this.preventFocusTimeout
       }
     )
-    this.setSelected(date, event)
     this.setOpen(false)
+    this.setSelected(date, event)
   }
 
   setSelected = (date, event, keepInput) => {
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -207,8 +207,11 @@ export default class DatePicker extends React.Component {
         return this.preventFocusTimeout
       }
     )
-    this.setOpen(false)
+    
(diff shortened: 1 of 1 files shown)


rafeememon:
A unit test for this guy would be great =)

Odrin:
Just noticed, that it's duplicate of #825, so you may close this and merge #825

PR #949, opened 2017-07-13
Added check on utcOffset to see if it was passed with a zero value

### LINKED ISSUE #920: Highlighting of "today" date not always correct when using utcOffset

greg-a-smith:
During a previous pull request (#921), I added a check for `utcOffset` when computing the `defaultPreSelection`.  This _almost_ worked.  As it turns out, it works for any non-zero `utcOffset`, however, if the `utcOffset` is passed as `0` (GMT), the check returns falsey and does not evaluate properly.  This change also checks for a value of `0`.

Fixes #920

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -116,7 +116,7 @@ export default class DatePicker extends React.Component {
       this.props.openToDate ? moment(this.props.openToDate)
       : this.props.selectsEnd && this.props.startDate ? moment(this.props.startDate)
       : this.props.selectsStart && this.props.endDate ? moment(this.props.endDate)
-      : (this.props.utcOffset || this.props.utcOffset === 0) ? moment.utc().utcOffset(this.props.utcOffset)
+      : (typeof this.props.utcOffset !== 'undefined') ? moment.utc().utcOffset(this.props.utcOffset)
       : moment()
     const minDate = getEffectiveMinDate(this.props)
     const maxDate = getEffectiveMaxDate(this.props)
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -116,7 +116,7 @@ export default class DatePicker extends React.Component {
       this.props.openToDate ? moment(this.props.openToDate)
       : this.props.selectsEnd && this.props.startDate ? moment(this.props.startDate)
       : this.props.selectsStart && this.props.endDate ? moment(this.props.endDate)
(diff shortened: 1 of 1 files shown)


greg-a-smith:
Similar to #921 , since the issue depends on a local environment's current date/time, I'm not quite sure how to replicate that in a unit test.  Passing a date in doesn't work because the fix is only hit when no date is passed in.

martijnrusschen:
This seems like an uncommon way of fixing this. Isn't there a javascript function that can check if there's something supplied that's a number? I'm  afraid that fixes like these are getting refactored at some point and will be gone because nobody can remember why this was added.

greg-a-smith:
@martijnrusschen Fair point.  I will look at making the change a better way.

aij:
I think the `=== 0` would handle `null` better. Treating `null` and `undefined` differently here seems surprising.

`moment.utc().utcOffset(null)` returns `0` rather than a moment object.

PR #960, opened 2017-07-19
Change selection of inline picker. fix #927.

### LINKED ISSUE #927: Datepicker don't switch to new range in inline mode

m1n0s:
Also remove semicolon by eslint checking.

fixes #927

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -111,7 +111,7 @@ export default class DatePicker extends React.Component {
   }
 
   componentWillReceiveProps (nextProps) {
-    if (this.props.inline && !this.props.selected.isSame(nextProps.selected)) {
+    if (this.props.inline && !this.props.selected.isSame(nextProps.selected, 'month')) {
       this.setPreSelection(nextProps.selected)
     }
   }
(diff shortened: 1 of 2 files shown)


m1n0s:
Hi @martijnrusschen!
I kindly ask you to check this PR as soon as you can and release it because my project is very depending now on this feature.

I hope I've done check and updating state in proper place and it not - I hope you get the idea and can do it better.

Anyway thank you a lot for this awesome and very flexible picker.

martijnrusschen:
Any chance we can add a regression test to avoid breaking this again?

rafeememon:
I think this definitely needs a test, @m1n0s if you need to use this immediately you can build your forked repo and use that in your project.

m1n0s:
@martijnrusschen @rafeememon 
I've added test case for my changes. Please review.

Leo

PR #963, opened 2017-07-22
Same day in highlightDates

gomo:
Hi!

Specifying the same day multiple times in `highlightDates` only applies the first class. I think that all classes should be applied, what do you think?


alex-shamshurin:
This way you want to collect all styles for a particular day? I think it's a good idea

PR #965, opened 2017-07-24
Adding error emitter for user input error

buhuitou:
This pull request is to address the needs to manage keyboard input error outside date picker component by emitting validation error when Enter, Tab, or Esc is pressed.

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -14,6 +14,8 @@ const WrappedCalendar = onClickOutside(Calendar)
  * General datepicker component.
  */
 
+const INPUT_ERR_1 = 'Date input not valid.'
+
 export default class DatePicker extends React.Component {
   static propTypes = {
     allowSameDay: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


aij:
Why only when Enter, Tab, or Esc are pressed?

buhuitou:
Hi! Thanks for your question.

As I understand, Enter, Tab or Esc key press gives the signal that user typing-in is ended hence eventual error events should be emitted.

I hope my answer is clear.

aij:
@buhuitou Some users like to (over)use a mouse. It seems weird to handle Tab differently from clicking to the next field.

Most users also have a multi-tasking OS these days. I don't know whether switching to a different app should be considered the end of typing or just a pause in typing though. I expect there is no right answer there, and it will be wrong no matter what we do.

buhuitou:
@aij I think your points are reasonable.

As to the first point, I definitely agree that the event should be emitted when the field loses focus instead of on Tab pressed.

As to the second point, I think the proposal implements only optional event and they may be simply ignored when not desired. For this reason, they are not intrusive but give flexibility of control.

By the way, at least on Mac, Cmd-Tab does not trigger Blur event. I think it's the same on Linux but I am not sure on Windows.

If you agree, I can adjust the implementation for the first point.

martijnrusschen:
@aij Thoughts? Good to merge?

aij:
@martijnrusschen Sure.

martijnrusschen:
@buhuitou This should be good to merge when the merge conflicts are resolved. Can you take a look, so that we can merge this one?

martijnrusschen:
Closing due inactivity.

PR #981, opened 2017-08-03
Avoid crash when inline and selected is null

mhaylock:
This addresses a problem introduced in #960 (not yet part of a release version) where if a date picker is `inline` then it will crash when `selected` is null, due to the check added in #960 assuming that it will not be null.


rafeememon:
Looks good, thanks!

mhaylock:
Thanks @rafeememon, do you have an idea of when the next release might be? I'm keen to be able to use this and the fix from #960 in a released version.

PR #992, opened 2017-08-11
add shouldCloseOnSelect prop to allow the calendar to remain open on …

### LINKED ISSUE #989: Allow to click on the calendar without closing it
I want to use the DatePicker with range (startDate and endDate)

image

The user will click twice to select the start and end date. the problem is that each click will close the datePicker.

I found an ugly hack by using ref:

```
<ReactDatepicker
          ref={reactDatePicker => this.reactDatePicker = reactDatePicker}
          onChange={val => //if firstClick the set startDate and don't close calendar, else set endDate and close calendar
          customInput={this.renderInput()}
          onSelect={() => setTimeout(() => this.r.reactDatePicker.setOpen(true), 0)}
          startDate={startDate}
          endDate={endDate}
          onSelect={() => setTimeout(() => this.reactDatePicker.setOpen(true), 0)} ///hack!!!!
/>
```

I can track the clicks and to know if this is first click or a second one, but I cant find a good way to leave the calendar open.
Even this hack is just opening the calendar just after it gets closed.

Can I create a pr to add a new dynamic prop like **shouldCloseOnSelect** ?
This will be a boolean, and I will set it to true when it is second clidk, and to false when this is going to be the first click.

This prop will be used here, in order to prevent the calendar to automatically close:
https://github.com/Hacker0x01/react-datepicker/blob/master/src/datepicker.jsx#L225

lbelinsk:
This pr will close #989 

What have changed:
1. When selecting a day with mouse click or Enter press the datePicker will remain open if a shouldCloseOnSelect prop will be provided with a false value.

2. When ```shouldCloseOnSelect = false``` after selecting a day, the calendar will remain open and then the state.preselection value date will be updated to allow the expected UX behaviour.
For example.

### DIFF
--- a/docs-site/src/example_components.jsx
+++ b/docs-site/src/example_components.jsx
@@ -38,6 +38,7 @@ import Children from './examples/children'
 import Portal from './examples/portal'
 import InlinePortal from './examples/inline_portal'
 import RawChange from './examples/raw_change'
+import DontCloseOnSelect from './examples/dont_close_onSelect'
 
 import 'react-datepicker/dist/react-datepicker.css'
 import './style.scss'
@@ -190,6 +191,10 @@ export default class exampleComponents extends React.Component {
   {
     title: 'Get raw input value on change',
     component: <RawChange/>
+  },
+  {
+    title: 'Don\'t hide calendar on date selection',
+    component: <DontCloseOnSelect/>
   }]
 
   renderExamples = () =>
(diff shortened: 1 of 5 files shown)


martijnrusschen:
Awesome!

lbelinsk:
Thanks, that was quick!
Any estimation when a new version will be released?

martijnrusschen:
Can do it later today

PR #1022, opened 2017-09-12
Adding example for OpenByDefault

jakehasler:
When using react-datepicker, I had a tricky time finding how to make the picker open by default, so I thought it'd be helpful to create an example for it.

Thanks --


jakehasler:
@martijnrusschen Just updated with master, and it should be good to go 👍 Thanks!

PR #1026, opened 2017-09-18
feat(calendar): add prop hideHeader

### LINKED ISSUE #1008: Allow to hide the Calendar's title
@martijnrusschen  I want to hear whether it will be ok to add a new boolean prop, something like `displayCalendarTitle` that will hide the header (August 2017 in this example):

image

Because it is a bit redundant to display it when you have a year and month dropdowns.

This is an example for what I am aiming for:

image

image

I am not sure, but maybe the title frame should be configurable?

Let me know what do you think, I will prepare a pr if it will be ok with you.

Thanks...

justame:
Related Issue:
https://github.com/Hacker0x01/react-datepicker/issues/1008#event-1253104
138

Closes #1008


test/hide_header_test.js
```diff
@@ -0,0 +1,33 @@
...
+describe.only('Calendar', () => {
+  function getCalendar (extraProps) {
+    return mount(
+      <Calendar
+          dateFormat="MMMM YYYY"
+          onSelect={() => {}}
+          onClickOutside={() => {}}
+          hideCalendar={() => {}}
+          hideHeader
+          dropdownMode="scroll"
```
lbelinsk: why do we need that?

test/hide_header_test.js
```diff
@@ -0,0 +1,33 @@
...
+          {...extraProps}/>
+    )
+  }
+  it('should show header specified in hideHeader props', () => {
+    var calendar = getCalendar({
+      hideHeader: false
```
lbelinsk: I would add a test that checks that the header is visible by default, not only when you pass false. This is important because it will make sure we didn't change any behaviour by default

test/hide_header_test.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
```
lbelinsk: I think that since it is not a component like "day", "week", "month", a new file might not be needed, just add a describe block in the datepicker test. up to you..

martijnrusschen:
Unsure why the coverage tool isn't picking up the tests, but a 50% coverage drop seems weird for this change.

justame:
@martijnrusschen The 50% drop is very weird, do you have any idea why does it happen?

martijnrusschen:
Not sure, but looks like the coverage tool is just a bit buggy. Happy to merge this if @lbelinsk approves it as well.

lbelinsk:
Actually after this discussion, we kind of agreed that the css solution might be better than our original idea.
We added it in our code:

image

https://github.com/wix/wix-style-react/commit/b5649904798f25ed5fbad363ee6d3ef868542159

Maybe this code snippet can be added here and the prop `hideHeader` can affect the css classes, instead of doing it in our code..

@justame @martijnrusschen wdyt?

martijnrusschen:
Sounds like plan!

lbelinsk:
Cool, we will take care of it next week, we have a holiday here until Sunday :)

martijnrusschen:
Awesome, thanks for the help!

Issue #1034, opened 2017-09-21
Custom format prop for showTimeSelect


actraiser:
Currently the format for the TimeSelect-feature is hard-coded to "hh:mm A" which is not popular in 24h-regions. It would be great if there was a timeFormat-Prop, e.g.

```
<DatePicker
   dateFormat="LLL"
   showTimeSelect
   timeFormat="HH:mm"
 />

```
-act

PR #1035, opened 2017-09-21
Added support for timeFormat prop #1034

ninio:
Also:
- Added a test case
- Updated the docs


namezero:
not available in version 0.55.0

GautierT:
Hello.
When this will be available ?
Thanks for this feature btw

thomax:
It seems `props.timeFormat` gets passed all the way down to `Time.jsx`, but internally, this component tries to read `props.format`, instead of `props.timeFormat`. Can submit a PR if you want?

Bug right here, unless I'm mistaken: https://github.com/Hacker0x01/react-datepicker/blob/master/src/time.jsx#L67

Issue #382, opened 2016-03-03
Provide a simple moment-less API


rafeememon:
In one of my projects I'm passing dates to server API calls as strings, and I also receive dates back as strings which are stored in redux state.  In the spirit of keeping my state clean I keep the strings as the source of truth rather than moment objects; this also makes it possible to pre-generate my state on the server, which would get ugly if it contained moment objects.  This led me to write a wrapper datepicker component that would simply translate between strings and moment objects so that they can be passed as props to the datepicker.  This got me thinking that maybe we should provide a way of dealing with strings natively in this component.  I can think of a couple approaches here:
1. **Provide wrapper component in datepicker** -- Only functionality would be to wrap/unwrap moment objects and pass them to the real datepicker component.  Probably the easiest to maintain, cleanest code-wise; but calling something `SimpleDatePicker` seems a little silly
2. **Provide a `simple` prop** -- When set all props (`selected`, `minDate`, etc, the value passed to `onChange`) will be strings formatted according to `dateFormat`.  A little dirtier code-wise; adding a prop to control how other props are handled seems kind of anti-react to me
3. **Automatically determine behavior** -- Determine the intended behavior by what is passed into `selected`: if a string, everything is a string, and if a moment, everything is a moment.  By far the simplest API; dirtier code-wise, and I'm not sure how it should behave when `selected` is `null`

What does everyone think?

rafeememon:
For posterity here is what my wrapper component looks like:
```
import React from 'react'
import DatePicker from 'react-datepicker'
import moment from 'moment'

const DATE_FORMAT = 'MM/DD/YYYY'

export default class SimpleDatePicker extends React.Component {
  static propTypes = {
    selected: React.PropTypes.string,
    startDate: React.PropTypes.string,
    endDate: React.PropTypes.string,
    minDate: React.PropTypes.string,
    maxDate: React.PropTypes.string,
    onChange: React.PropTypes.func.isRequired
  };

  getMoment (string) {
    return string && moment(string, DATE_FORMAT)
  }

  onChange (date) {
    this.props.onChange(date && date.format(DATE_FORMAT))
  }

  render () {
    return <DatePicker
      {...this.props}
      dateFormat={DATE_FORMAT}
      selected={this.getMoment(this.props.selected)}
      startDate={this.getMoment(this.props.startDate)}
      endDate={this.getMoment(this.props.endDate)}
      minDate={this.getMoment(this.props.minDate)}
[...]

PR #1046, opened 2017-09-30
Optional moment.js Part 1

prayerslayer:
Part 1 of https://github.com/Hacker0x01/react-datepicker/issues/382

* Moves all calls on moment or instances of moment into `date_utils`. (That I found, at least. Since most date properties are still instances of moment I might have missed some.)
* Discussable if there are too many functions and how the moment-agnostic Date API should look like.
* No efforts regarding order of functions were made, this I'll do later but I still wanted to give you the change to look at it.

After this PR I would work on replacing the actual moment instances.


prayerslayer:
> I don't think we could completely remove moment while maintaining backwards compatibility

Yeah me neither, except you'd distribute separate builds for versions with and without moment. If that's not an option we'd need a flag *somewhere* to know which date utils to use.

prayerslayer:
Any idea what could be wrong with the failing test? It seems to be occuring also on `master` (see this build https://travis-ci.org/prayerslayer/react-datepicker/builds/281940187), so I assume it's a dependency issue?

I'll continue on a branch based on this PR instead of master.

aij:
@prayerslayer It appears to be a fragile test. It looks like it will always fail on the first 2 days of the month.

montogeek:
Any updates on this?

prayerslayer:
@montogeek There's this https://github.com/Hacker0x01/react-datepicker/pull/1327, but it's quite old by now, and has some blocking open questions.

PR #1067, opened 2017-10-11
custom time select caption

alalpod:
Hi, added timeSelectCaption prop to suit other languages.

### DIFF
--- a/docs-site/src/examples/show_time.jsx
+++ b/docs-site/src/examples/show_time.jsx
@@ -26,6 +26,7 @@ export default class ShowTime extends React.Component {
     <strong>{`    showTimeSelect
     timeFormat="HH:mm"
     timeIntervals={15}
+    timeSelectCaption="Caption"
     dateFormat="LLL"
 />
 `}</strong>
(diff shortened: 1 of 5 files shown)


docs-site/src/examples/show_time.jsx
```diff
@@ -38,6 +39,7 @@ export default class ShowTime extends React.Component {
             showTimeSelect
             timeFormat="HH:mm"
             timeIntervals={15}
+            timeSelectCaption="Caption"
```
martijnrusschen: Maybe say "Time" here?

docs-site/src/examples/show_time.jsx
```diff
@@ -26,6 +26,7 @@ export default class ShowTime extends React.Component {
     <strong>{`    showTimeSelect
     timeFormat="HH:mm"
     timeIntervals={15}
+    timeSelectCaption="Caption"
```
martijnrusschen: Maybe say "Time" here?

alalpod: "Time" is default prop, so I put a "Caption" here that everyone could see what it does. But if you think there should be "Time" i will change it.

martijnrusschen:
Should be good to go when merge conflicts are resolved.

martijnrusschen:
Closing due inactivity.

Issue #962, opened 2017-07-20
Clear button is not working when date is given as input


bid007:
I am using react datepicker in one of my project. I have been facing this issue. When I enter the date as input and try to clear it, it won't work. There is a state called "InputValue" which gets set when date is given as input. On clear it has to be set as null but does not. 

Steps to replicate:
1. Go to https://hacker0x01.github.io/react-datepicker/#example-18
2. Find the example "Clear datepicker input"
3. Input a date eg. "07/10/2017".
4. Click on clear button. It does not work. 

**Gave an input date**
<img width="709" alt="screen shot 2017-07-20 at 3 11 23 pm" src="https://user-images.githubusercontent.com/3582804/28436788-ddc3ecc2-6d5d-11e7-8884-635a1294bba2.png">

**Clicked on clear button**
<img width="652" alt="screen shot 2017-07-20 at 3 11 34 pm" src="https://user-images.githubusercontent.com/3582804/28436806-ed2494be-6d5d-11e7-8ae4-0a79e4d7e588.png">

MarkusLund:
Any progress on this issue?

rafeememon:
@MarkusLund this doesn't appear to be an issue on the demo site using the repro steps, do you have any more info?

bid007:
The issue can be replicated in demo. If you enter the date manually and try to clear, it won't work.

MarkusLund:
@rafeememon As @bid007 are stating, this is an issue in both my local version and the demo.

rafeememon:
Cool, PRs to fix it are welcome

rnauthon:
The clear button is still not working on Firefox.

muhyidin3222:
so far is still not working any platform

bronz3beard:
this works its not so pretty but it works
    
    ```ts
    
    const [startDate, setStartDate] = useState<Date | null>(value ? new Date(value) : null);

    useEffect(() => {
      document.addEventListener('click', (event: Event) => {
        if (
          event.target &&
          (event.target as HTMLButtonElement).className! === 'react-datepicker__close-icon'
        ) {
          setStartDate(null);
          onChange(''); // react-form-hook
        }
      });

      return () => {
        document.removeEventListener('click', () => {});
      };
    }, [onChange]);
    
    ```

PR #1071, opened 2017-10-12
Clear button is not working when date is given as input #962

mkovrigovich:
solves https://github.com/Hacker0x01/react-datepicker/issues/962


stevenperezGL:
@mkovrigovich Thanks for this PR ! :)

PR #1074, opened 2017-10-13
Optionally wrap the popper element with a user-supplied container

### LINKED ISSUE #969: Datepicker behavior inside container with overflow: scroll
Had a problem somewhere between 0.47 and 0.52 that I believe might have been caused by the transition to tether. 

Our datepicker was in a container that was overflow:scroll, and in the old incarnation (0.47) it was fine because the calendar widget part got appended to body, but now it stays with the input dialog inside the container and was subsequently affected by the overflow (which makes sense from a CSS point of view, but not as useful for the calendar to go over the top). 

Now I can probably remove the overflow scroll as this may be intended behavior (but for people who cannot remove this, but would still like the old behavior), or perhaps I can change some options from: https://hacker0x01.github.io/react-datepicker/#example-20 to affect a better solution.

Or perhaps

Thanks,

pmacmillan:
react-popper will place the popover in the same constraints as the
parent div. For cases where the parent div is somehow constrained -- a
scrollable div -- the popper will appear inside the div and will be
constrained by it to.

There are cases where you want the popover to be unconstrained it's
parent. To do this, you must place the popover in a container outside of
the constrained container.

demo: https://codesandbox.io/s/1ozy5ol5vl

Adding a popperContainer would allow a user to implement the following

  function CalendarContainer (props) {
    const containerId = document.getElementById('container-outside-root')

    return (
      <Portal container={containerId}>
        {props.children}
      </Portal>
    )
  }

  // somewhere else

  <DatePicker ... popperContainer={CalendarContainer} />

Which will then permit rendering the control inside the div and the
calendar outside the div (which the popper library will connect through
the magic of javascript)

### DIFF
--- a/src/datepicker.jsx
+++ b/src/datepicker.jsx
@@ -87,6 +87,7 @@ export default class DatePicker extends React.Component {
     openToDate: PropTypes.object,
     peekNextMonth: PropTypes.bool,
     placeholderText: PropTypes.string,
+    popperContainer: PropTypes.func,
     popperClassName: PropTypes.string, // <PopperComponent/> props
     popperModifiers: PropTypes.object, // <PopperComponent/> props
     popperPlacement: PropTypes.oneOf(popperPlacementPositions), // <PopperComponent/> props
@@ -534,6 +535,7 @@ export default class DatePicker extends React.Component {
               {this.renderClearButton()}
             </div>
           }
+          popperContainer={this.props.popperContainer}
           popperComponent={calendar}
           popperPlacement={this.props.popperPlacement}/>
     )
--- a/src/popper_component.jsx
+++ b/src/popper_component.jsx
@@ -26,6 +26,7 @@ export default class PopperComponent extends React.Component {
     popperComponent: PropTypes.element,
     popperModifiers: PropTypes.object, // <datepicker/> props
     popperPlacement: PropTypes.oneOf(popperPlacementPositions), // <datepicker/> props
+    popperContainer: PropTypes.func,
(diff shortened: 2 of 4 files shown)


pmacmillan:
Fix #969

iamandrewluca:
I am using `reactstrap`. Just added this
```js
import Portal from 'reactstrap/lib/Portal'
// ...
<DatePicker
  // ...
  popperContainer={Portal}
  // ...
/>
```
creation and removal of nodes are handled by `Portal`. I laughed. Thanks!

Rafael-Ramblas:
First, thanks for the PR, it really helped me, but, when i open the date-picker inside a modal, it renders the picker under the modal, so its clipped where the modal ends.
captura de tela de 2018-09-12 15-09-51

thiagofront:
works fine

``` jsx
import ReactDOM from 'react-dom';

const $el = document.querySelector('#yourElement');

const PopperContainer = ({ children }) => (
  ReactDOM.createPortal(
    children,
    $el,
  )
);

<DatePicker
  {...anotherConfigs}
  popperContainer={PopperContainer}
/>
```

activenode:
> works fine
> 
> ```js
> import ReactDOM from 'react-dom';
> 
> const $el = document.querySelector('#yourElement');
> 
> const PopperContainer = ({ children }) => (
>   ReactDOM.createPortal(
>     children,
>     $el,
>   )
> );
> 
> <DatePicker
>   {...anotherConfigs}
>   popperContainer={PopperContainer}
> />
> ```

What I was searching for.

Issue #1060, opened 2017-10-08
Doesn't working on React 16


IftekherSunny:
hey, I can't use your package on React 16. Do you guys have any plan for supporting React 16????

mmcshinsky:
I have a working version of React 16 on a personal project with TimeSelect working as selected. Do you have example code that you can provide that we can see?

acierto:
Same here, when I select a date, I have such error

<img width="720" alt="screen shot 2017-10-12 at 11 01 27" src="https://user-images.githubusercontent.com/2265094/31488029-bd639e84-af3c-11e7-81d8-06a5e4b6e464.png">

josegers:
We have the same problem. Is there an update planned to support React 16?

antonigiske:
It works if you update yarn. I had the same problem with older version of yarn.

aij:
Duplicate of #972

Issue #972, opened 2017-07-27
Support React 16


yamat124:
Datepicker is not working with React 16.0.0 Beta 1. I was getting errors on PropTypes in the react-datepicker.min.js file.

martijnrusschen:
Can you be more specific and post the errors here?

yamat124:
Figure out it was a webpack include issue with an older version.

alex-shamshurin:
We must update proptypes package:

```
import React from 'react'
import PropTypes from 'prop-types'
```

rafeememon:
I think this is going to be trickier than just updating the proptypes package.  React is declared as a peer dependency supporting two different major versions (0.x and 15.x).  Ideally, the package versions of `react` and `prop-types` should match.  I'm not exactly sure if there's some dependency combination that ensures that.  I'm reopening this issue so that we can figure out a solution.

aij:
FWIW, We're currently on React 15.1.0, which doesn't use the `prop-types` package. The only harm in using components that use the `prop-types` package seems to be that we get deprecation warnings about how calling prop types directly is not supported. (They're somewhat misleading in that they blame the render method of a component rather than blaming React itself for the deprecated invocation.)

I do not know how it interacts with 0.14.

Do we actually support 0.14? The README would appear to indicate 15.5 is required for newer versions of `react-datepicker`, though `package.json` still says `^0.14.0 || ^15.0.0`.

dimitropoulos:
for notoriety of the impact of this issue: +1 if this is preventing you from using React 16 (now that it's released)

Baldrs:
Any workaround to make it work with react 16? This is the only incompatible library that I have in my dependencies

bjankord:
There is an open PR to add compatibility with react v16 here: https://github.com/Hacker0x01/react-datepicker/pull/1047 @martijnrusschen @rafeememon would you have some time to take a look?

jeffijoe:
I'm also having issues with React 16:

![image](https://user-images.githubusercontent.com/2505474/31382963-9bc0a404-adb9-11e7-809b-e8c03eb1407f.png)

rufman:
Looks like https://github.com/souporserious/react-popper/pull/59 is merged, so updating to the latest version will resolve the issue.

pawarvijay:
### Still Its giving error with react 16 ,  am i missing anything

```
toms-Macbook:~ tom$ cd fireworks/
[...]

PR #1080, opened 2017-10-20
package.json: Allow React 16 as a peer dependency

rufman:


aij:
Is that all that's really needed?

From some comments on #972 and #1060 it sounds like it might be... I suppose it would at least prevent people from mixing different versions of React, which seems to never go well.

rufman:
Yeah I tested it on my setup and everything works, except for `yarn check` which throws a peer dep error. `react-popper` will need to be updated too once https://github.com/souporserious/react-popper/pull/59 merges.

PR #1087, opened 2017-10-24
package.json: update `react-onclickoutside` to work with React 16

rufman:


rafeememon:
We downgraded the dependency (https://github.com/Hacker0x01/react-datepicker/pull/1083) because of compatibility issues, so I'm skeptical of blindly making this change.

rufman:
@rafeememon When I tried the local build (yarn run start), everything seemed to be working fine.

PR #1090, opened 2017-10-25
Extract font color of month and time headers to vars

firec0der:
This tiny patch will give us the opportunity to set font color of these headers.

image


firec0der:
Btw, I mentioned that there is little mess with these variables.
There is no way to customize only the color of "prev" and "next" arrows. There is `$datepicker__muted-color` variable for it, but it's also used for disabled dates.

I can separate it if you want.

PR #1091, opened 2017-10-27
Update react-onclickoutside dep

Andarist:


zslabs:
You'll probably want to run `yarn` in your PR branch so the `yarn.lock` file can be updated as well.

Andarist:
ah yeah, done the changes from the github UI, havent noticed this project has a lockfile

fixed :)

william57m:
Perfect, this need to be merged, I have some warnings because of that since I updated my app to React 16, thanks.

SpencerTorres:
I hope it gets merged sometime today. My production app is using a terrible hack that increased my app bundle from about 60 to 600kb. 

Thanks for getting this PR submitted!

zslabs:
If you need it “now now” then you could always just fork the repo with the change and add that as your dep instead - then check back in a few days.

Andarist:
For the time being if you use npm5 you can specify package.json entry as:
```
"react-datepicker": "Andarist/react-datepicker#fix/deps"
```

It will build this project correctly inside of your `node_modules` using its dev deps and all.

rafeememon:
Let's get this in, thanks for your work @Andarist

PR #1094, opened 2017-10-28
Add prepare npm script

Andarist:
This allows npm5 to prepare the package when installed from the git.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "test": "cross-env NODE_ENV=test grunt travis",
     "eslint": "eslint **/*.jsx",
     "prepublish": "npm run build",
+    "prepare": "npm run build",
     "codecov": "cat coverage/*/lcov.info | ./node_modules/codecov.io/bin/codecov.io.js"
   }
 }


package.json
```diff
@@ -104,6 +104,7 @@
     "test": "cross-env NODE_ENV=test grunt travis",
     "eslint": "eslint **/*.jsx",
     "prepublish": "npm run build",
+    "prepare": "npm run build",
```
rafeememon: I'm not sure why this is needed -- isn't the use case covered by `prepublish`? https://docs.npmjs.com/misc/scripts

Andarist: see https://docs.npmjs.com/misc/scripts#deprecation-note

+ https://github.com/npm/npm/releases/tag/v5.0.0
> Git dependencies with prepare scripts will have their devDependencies installed, and npm install run in their directory before being packed.

martijnrusschen: So should we remove the prepublish command and move to prepare completely?

Andarist: truth to be told it is a little messy - in general the answer is **yes**, if you have certainty that you won't use by accident npm older than v4 when publishing, because then without a `prepublish` script you might publish broken package, because it wont recognize `prepare` script, it has happened to me (other collaborator on the package I maintain had npm3 and he had published by accident broken package), that's why I have added simple guard in my projects for the time being:
```
"prepublish": "test $(npm -v | tr . '\\n' | head -n 1) -ge '4' || exit 1",
```

npm still warns me that `prepublish` is deprecated and I should migrate to `prepare`, but at least it will prevent broken publications

if you are the only one working on a project, u might not need to do that

Andarist:
I forgot I've done this PR and I've added `prepare` script in my recent 'modern-build' PR. Closing this as it has already found its way to `master`

PR #1115, opened 2017-11-03
Added module entry point - refactored build process, removed grunt in…

Andarist:
… favor of npm scripts


.babelrc.js
```diff
@@ -0,0 +1,29 @@
+var NODE_ENV = process.env.NODE_ENV
+var MODULES = process.env.MODULES
+
+var modules = MODULES === 'false' || NODE_ENV === 'test' ? 'commonjs' : false
```
Andarist: this `MODULES` env variable should be removed when you upgrade your webpack (should be fairly easy, just following migration guides, but I didnt have time to do that)

martijnrusschen: That's a good one. I actually tried upgrading a while ago but got stuck on a error I couldn't explain. See #735.

martijnrusschen:
Thanks, giving this a go.

martijnrusschen:
It loos like these scripts broke the re-generation of the documentation site stored in /docs-site. I just pushed some changed to there and don't see these reflected when running the build scripts.

Andarist:
I'm not sure if I understood correctly. Were you able to fix the issue or is there an issue with the changes and the fix is needed?

martijnrusschen:
Yeah sorry, still having issues. Are you able to help me out here?

Andarist:
Sure, could you describe the issue in more details? That would get me starting sooner.

martijnrusschen:
To publish new changes to the `gh-pages` branch I'm using a script that's stored here: https://github.com/Hacker0x01/react-datepicker/blob/master/docs-site/publish.sh. This script also uses the `npm build` scripts, but I'm not sure whether this is actually updating the bundle.js in the docs-site folder. After pushing some changes to the docs-site and running the release script, I didn't see those changes committed to the gh-pages branch, so I thought the new scrips might not take this case into account yet.

Andarist:
That's probably the case, gonna look into it soon.

PR #1121, opened 2017-11-06
Update README.md

honzajerabek:
update css files dist paths for 0.60.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,10 +25,10 @@ import React from 'react';
 import DatePicker from 'react-datepicker';
 import moment from 'moment';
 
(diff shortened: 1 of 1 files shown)


martijnrusschen:
Hmm, I would prefer changing back the compiled CSS files to the original names to avoid breaking behaviour for existing installations.

honzajerabek:
Agree! I saw you were doing some building refactors, so I changed the names, but previous state would be better for most users.. I can send a PR after lunch ^^

martijnrusschen:
Thanks, yes that would be appreciated!

honzajerabek:
pushed a new PR, closing this one

PR #1123, opened 2017-11-06
New functionality to adjust day when changing week, month or year

forum-is:
We propose the following added functionality to improve the user experience for the datepicker component.

If the new property **adjustDateOnChange** is set to **true**, the date will automatically change when the current month is altered via the arrow controls or a different year is selected via the year dropdown. Also, all changes through keyboard shortcuts are supported. If **adjustDateOnChange** is undefined or false, the behaviour is unchanged from the current release versions.

The following example illustrates the new behaviour:

- current date is 2017-10-31 (October 31st, 2017)
- the month is changed to November by pressing the right arrow symbol inside the datepicker
- the date will automatically adjust to 2017-11-30 (nearest possible date in the following month)

The same methodology applies if the year is changed.


martijnrusschen:
Can you add some tests to cover your new behaviour? We want to avoid breaking this behaviour next time someone touches this code.

PR #1155, opened 2017-11-29
Update moment package to fix vulnerability

lymancheung:


rafeememon:
I don't think this change is necessary -- the new version already satisfies the caret dependency.

aij:
@rafeememon `yarn.lock` will keep us pinned at the previous version. I think this change is actually insufficient.

@lymancheung Please use `yarn add` instead of editing the files by hand, and commit both files.

martijnrusschen:
Upgraded in 14f886b283e73bb8a4b0fb26e27c0cce4892136a

Issue #1077, opened 2017-10-18
Problem with two datepickers for date range on Safari


InekeScheffers:
![react-datepicker-safari](https://user-images.githubusercontent.com/12294859/31717261-775abd86-b40b-11e7-8385-0b0c4b9965de.gif)
On Safari when you select a date on one of the datepickers and change to the other datepicker too quickly, both open the calendar and the focus on the input fields keeps switching.

cliener:
I've seen similar oddities in iOS Safari. Looks like events are firing out of order potentially related to #597

echarso:
Hi I am facing the same issue here

pmacmillan:
The place where the click happens seems to matter.

![datepickerissue](https://user-images.githubusercontent.com/3280199/32070499-903c6926-ba5a-11e7-9a31-3f926eeea9f5.gif)

If I click past the end of input, everything works fine. If I click in the middle of the input, the issue crops up.

pmacmillan:
What I am seeing is:  if I have one open and...

- click after the end of the other input, then `handleCalendarClickOutside` gets called
- click within the value of the other input, `handleCalendarClickOutside` is never called

In both cases, `onBlur` is called, but in the second instance the value of `this.state.open` is true and the component uses `deferFocusInput` to regain focus.

(all within `src/datepicker.jsx`).

In Chrome and Firefox, `handleCalendarClickOutside` is always called.

pmacmillan:
Looks like Safari is event drunk.

![safari-events](https://user-images.githubusercontent.com/3280199/32111503-ad8dc728-bb08-11e7-9455-181b261b1614.gif)

vs

![chrome-events](https://user-images.githubusercontent.com/3280199/32111506-b324af3a-bb08-11e7-92f4-420fc87fe541.gif)

from: https://www.quirksmode.org/dom/events/tests/blurfocus.html

Basically, if you click within the content of the text input with another input element focused, you get something like:

<pre>
blur on textarea
focus on text
blur on text
focus on textarea
blur on textarea
focus on text
</pre>

This would explain the observed behaviour.

I suppose an approach to fix this might be to debounce the focus/blur events, reacting only to the last one... but... yikes.

pmacmillan:
Interesting: https://github.com/facebook/react/issues/10871 Related?

pmacmillan:
At issue are the spurious events that safari seems to emit like crazy.

[...]

PR #1157, opened 2017-12-01
Added `react-input-fixer` to fix Safari 10/11 MacOS phantom events fr…

cliener:
…om firing. Addressing the bug raised in #1077 

Fixed `setFocus()` to handle instances where the input isn't correctly set.
Tidied up a couple of inline `if` statements.
Rolled `extract-text-webpack-plugin` back to 1.0.1 for compatibility with `webpack` 1.13.0 which is currently used in the project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "eslint-plugin-react": "^7.4.0",
     "eslint-plugin-standard": "^3.0.1",
     "express": "^4.16.2",
-    "extract-text-webpack-plugin": "^3.0.2",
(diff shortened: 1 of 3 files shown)


martijnrusschen:
Seems worth it, can you make sure the build's succeeding?

cliener:
I've looked into the four failing tests and can't see what might be done about them. When testing each manually in-browser, the tests pass.
The fix I've implemented deliberately suppresses fast blur/focus events which is probably what's causing the failures.
Any suggestions about what can be done to pass the tests?

PR #1166, opened 2017-12-08
Allow passing ref via custom props

lewisdiamond:
When using react-datepicker with material-ui's `TextField`, you get an error because `TextField` is a stateless fn component and you can't pass `ref` to it. It however accepts `inputRef`. Allow users to define how the ref is passed to the custom input.

### DIFF
--- a/docs/datepicker.md
+++ b/docs/datepicker.md
@@ -12,6 +12,7 @@ General datepicker component.
 |`children`|`node`|||
 |`className`|`string`|||
 |`customInput`|`element`|||
+|`customInputRef`|`string`|`'ref'`|The property used to pass the ref callback|
 |`dateFormat`|`union(string\|array)`|`'L'`||
 |`dateFormatCalendar`|`string`|`'MMMM YYYY'`||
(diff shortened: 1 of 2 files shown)


aij:
I think it would make more sense to wrap `TextField` in a component that follows the standard API.

lewisdiamond:
@aij I disagree. Adding `customInputRef` just adds more flexibility, you can even wrap your custom input in a component (e.g. that does styling) without having to worry about the input API every time.

aij:
It doesn't *just* add more flexibility. It also adds more complexity.

What's the reasoning for not also adding `customInputValue', 'customInputOnBlur`, `customInputId` `customInputClassName` etc?

lewisdiamond:
`customInput` already exists except it's limited to only class components because of the `ref` prop. `ref` isn't accessible from within the component, it's removed by React. `ref` is the only property like that.

jwm0:
@lewisdiamond Can I see an example of how it's supposed to be done using material-ui TextField?

PR #419, opened 2016-03-19
Move moment to peerDependencies

martijnrusschen:
Fixes #122

PR #1168, opened 2017-12-11
Move moment to peerDependencies

### LINKED ISSUE #122: Moment.js as a peer dependency
Hi,

Moment.js is a big dependency that I don't want to add twice to my bundle. Could it be moved to the peerDependencies in `package.json` like react?.

Thanks for this component, works great.

Cheers,
Javi

Kerumen:
Supersedes #419 

Fixes #122 

It's better to move `moment` to `peerDependencies` because it prevents duplication of moment instances which can leads to bugs (when loading locals for instance).
In our project, we have a fixed dependency of `moment@2.18`. `react-datepicker` installs `moment@2.19` so 2 instances of moment are loaded and `locals` doesn't work for the moment of `react-datepicker`.

If you look at `react-dates`, this is how they require `moment`: https://github.com/airbnb/react-dates/blob/master/package.json#L126

I also removed `package-lock.json` as it's not used in the project anymore. Tell me if I should remove this from the PR as it's not quite related (but I didn't know which lockfile to update so I thought it should be better to delete it).


martijnrusschen:
I agree this is probably better, however there's some discussion going on here #122.

martijnrusschen:
Not sure if we can make this change without breaking everyone's datepickers. What do you think?

Kerumen:
Some discussion which ended 1 year ago. This should be reconsidered.

I think that everyone who use the datepicker are already using `moment` but if we want to be _truely_ semver compliant, this can be released as a breaking change.

aij:
@Kerumen I think the plan is to bump the major only when we think the API is stable. Keeping it at 0 is also truly semver compliant since semver says

> 4. Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

Kerumen:
@aij Yes, seems legit too. npm won't auto-upgrade people and if they manually do it, they will know why their code break (if it's break).

martijnrusschen:
Okay lets do this. We use yarn.lock instead of the package-lock.json, so lets make sure that's updated.

martijnrusschen:
@Kerumen Let me know if you want to proceed with this, happy to merge now we're reaching consensus.

Kerumen:
@martijnrusschen Conflicts resolved. I set `^2.0.0` as semver hence people won't have warnings if they install a previous version than `2.19.4`. I suppose all moment `>=2` works.

PR #1181, opened 2017-12-18
Added preventOpenOnFocus prop

### LINKED ISSUE #1177: Optionally prevent the DatePicker from opening onKeyDown
Hello,
Thank you for the DatePicker, it looks great so far.

I’m trying to prevent the DatePicker from opening from anything but clicks.
The goal is to allow the user to either navigate with the _Tab_ key in the form and type the date manually, or optionally open the DatePicker on click. I'm using the disabled keyboard navigation.

I found an `onPreventeFocus` state I can turn on that successfully prevents the DatePicker from opening when I gain focus from _Tab_ key.
However, since the `onKeyDown` prop calls `onInputClick()` line 336 of index.jsx, it opens the DatePicker every time the user tries to type something.

If I could add` !this.state.preventFocus` inside the `if(…)` line 336 of _index.jsx_, I should be able to achieve the behavior I’m looking for.

Preferably, I could make a new prop or state to achieve the same result (depending on if you want to expose it publicly).
It would allow to leave `this.state.preventFocus` alone, since it seems to have been created for a different purpose.

I’m ready to create a pull request depending on the solution you prefer.

I tried my best to solve my issue on my own, please tell me if I missed anything or if you need more information.

Jmenache:
Resolves #1177

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -93,6 +93,7 @@ export default class DatePicker extends React.Component {
     popperClassName: PropTypes.string, // <PopperComponent/> props
     popperModifiers: PropTypes.object, // <PopperComponent/> props
     popperPlacement: PropTypes.oneOf(popperPlacementPositions), // <PopperComponent/> props
+    preventOpenOnFocus: PropTypes.bool,
     readOnly: PropTypes.bool,
     required: PropTypes.bool,
     scrollableYearDropdown: PropTypes.bool,
@@ -138,6 +139,7 @@ export default class DatePicker extends React.Component {
       onSelect () {},
       onClickOutside () {},
       onMonthChange () {},
+      preventOpenOnFocus: false,
       monthsShown: 1,
       withPortal: false,
       shouldCloseOnSelect: true,
@@ -213,7 +215,9 @@ export default class DatePicker extends React.Component {
   handleFocus = (event) => {
     if (!this.state.preventFocus) {
       this.props.onFocus(event)
-      this.setOpen(true)
+      if (!this.props.preventOpenOnFocus) {
+        this.setOpen(true)
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


martijnrusschen:
Can you add some tests?

Jmenache:
Yes, I'll try to.

martijnrusschen:
It looks there's a merge conflict, can you make sure that's resolved so that I can merge this when it's ready?

Jmenache:
It is resolved.

msueping:
> @Jmenache I just came across the `preventOpenOnFocus` prop which it looks like you introduced in this PR. My assumption of how it would behave is that when a DatePicker instance is focused (via keyboard tab), the calendar popover would not automatically open. To open it once in focus, you'd have to hit ENTER or DOWN. This is not working, however, and it appears that with `preventOpenOnFocus` is `true`, there is no way to open the calendar popover via keyboard. Please advise.

Did you ever get an answer to this? I can't believe that there is no way to open calendar via the keyboard when the preventOpenOnFocus prop is true. WOW.

Jmenache:
@mellis481 I am sorry, I realize I missed your question last year.
Your assumption of how it would behave seems good to me (however, to open it once in focus, you can also click).

I don't remember much about this PR and I do not use the `preventOpenOnFocus` option anymore.
However the issue I opened at the time #1177 gives a bit of extra information. Please note that I was a beginner at the time.

The code of the PR is very short.
I only added a check in `handleFocus`, which seems reasonable, and one in `onInputKeyDown` which is causing the problem.
I am not sure why I added the later, but **it does look like a mistake**.

I guess I should have added two different props, `preventOpenOnFocus` and `preventOpenOnKeyDown`. That's actually the original name of my issue #1177.
Probably creating a new pull request to add a `preventOpenOnKeyDown` would be a solution.

Hope this helps.

PR #1189, opened 2017-12-25
Use prettier for code styling

martijnrusschen:
Don't think about style anymore, prettier will automatically format your code when committing. Added a pre-commit hook to make sure prettier runs before pushing code, and removed all ESlint style rules.

Expect happier developers that don't have to worry about failing builds because of wrong styles anymore.


aij:
Prettier isn't really a replacement for eslint.

I think it makes sense to use prettier for code formatting while continuing to use eslint for linting. There is a tiny bit of overlap (eg: semicolon enforcement), so they would need to be configured to not contradict each other.

I also generally prefer keeping the actual change in a separate commit from the automated reformatting if possible. It makes it much easier to rebase any unmerged commits.

martijnrusschen:
Okay, brought back ESlint for the non-style rules. I didn't look into the separate commit yet, but I don't think that's a blocker. I know a lot of folks are also using prettier plugins in their editors, so I guess that's less important to have all prettier changes in a separate commit. We can look into this later on though. If you have something, feel free to whip up a PR.

Issue #1142, opened 2017-11-14
Input is not refreshed when selected prop changed to null or undefined


marekadamek:
Version: 0.60.2

Steps to reproduce:
1. (important) Paste valid input value using ctrl-c ctrl-v
2. Manipulate selected prop from outside datepicked component and set it to null or undefined

Result:
onChange callback is executed with new value (null/undefined) but label shows old value

Note: 
Field is properly cleared when using build-in clear button or when the value is selected by click or enter

keawade:
This occurs for me when pasting input or manually typing input.

bikerp:
I'm facing the same issue as keawade. Thanks for the fix.

kblodget:
Could you post the fix to this please?

chrisjpatty:
Also facing this issue. Is there a function to manually clear the input in the meantime?

chrisjpatty:
I answered my own question, and it may help someone else. After looking through the code it looks like you can manually call the clear function that gets called when you click on the clear button. 

First, add a reference to the `DatePicker` component
```jsx
<DatePicker ref={datePicker => this.datePicker = datePicker} />
```

Then you can access the function by calling:
```jsx
someMemberFunction = (e) => {
    this.datePicker.onClearClick(e)
}
```
This should clear the input. 
*Note:* You do have to pass an event object to this function or it will crash while trying to call `e.preventDefault()`. If you don't have an event object I suppose you could synthesize one by passing an object like so: `this.datePicker.onClearClick({preventDefault: () => {}})`

Hopefully that helps in the meantime while the maintainers get around to looking at this and related issues.

keawade:
@chrisjpatty's work around does not resolve the issue for me. If I manually enter a date and then clear it by setting the value of the moment variable to `null`, the value is cleared but the text stays in the input field.

The workaround just changes the steps taken to set the value to null. Calling the clear function just passes `null` to the component's update function which calls my update function instead of just calling my update function directly.

Issue #888, opened 2017-06-14
Resetting state on parent component doesn't reset input


dsvictorh:
When inputting the date from the input typing rather than choosing with the datepicker controls if I reset the state of the parent component that passes the date value onto a child component that inside has the Datepicker which is turn uses the property of the middle component to set it's own value we can see the choice on the calendar resetting, but the input stays unchanged.

If for example I have a reset button in this scenario that resets to today's date after choosing the date 05/05/2017 15:00 and resetting the calendar and the input reset back to say 06/14/2017. If however I type 05/05/2017 15:00 entirely from the input or even choose the date and type in a small change, say the time itself, at the point where I reset you see the calendar choice back at 06/14/2017....yet the input is still set to what I previously typed disregarding the current value of the calendar.

![before-reset-choice](https://user-images.githubusercontent.com/4581131/27150683-65723af4-5105-11e7-9808-a8daf4eaba06.png)
![after-reset-choice](https://user-images.githubusercontent.com/4581131/27150682-6571ce84-5105-11e7-9220-8ceb0569e89e.png)
![before-reset-typed](https://user-images.githubusercontent.com/4581131/27150681-656eea20-5105-11e7-915f-05ba2a5a86dd.png)
![after-reset-typed](https://user-images.githubusercontent.com/4581131/27150684-6573f8ee-5105-11e7-8612-69b44a68384c.png)

aij:
If you want to reset the value of the input string as the user is typing, I think you can still do that via the `value` prop.

The problem with resetting the input every time `selected` changes, is that it's common to update `selected` from the `onChange` function, which results in issues like #560, where the input was getting reset while the user is typing. (If they happened to type an intermediate value that was parseable as a date.)

dsvictorh:
It was definitely a bit hectic but playing with a separate inner state just to define the value of the input I can force an empty string or today's date at reset! Thanks @aij !!!

alex-shamshurin:
I think react-datepicker should get rid of internal state. It must be maintained somewhere else.

martijnrusschen:
Anything actionable what we can do today? Removing internal state seems more like a long term solution.

BillOTei:
related to #1025 any news?

BillOTei:
issue seems to come from `_this.setSelected(date,` event, true);` keepInput boolean should be `false` when called in `_this.handleChange`

aij:
@BillOTei No, that would result in resetting the input while the user is typing.

[...]

Issue #754, opened 2017-03-01
There is no way to clear date filters other than isClearable prop.


ghost:
I have successfully integrated date picker in my app.
I have used it as for date to and date from it is working good. However I have one CTA on click of that CTA I want to clear both the date fields, but there no ways to do this. If anyone can suggest how to clear input fields then it will be helpful for us.

victordelarocha-gpsw:
Same issue here. How can I clear the date picker without explicitly using isClearable?

aij:
Have you tried passing `selected={null}`?

jcrben:
Passing null does seem to work, but it also seems to not too hard to end up with an 'Invalid date' - which is what I got when I pulled down https://github.com/Hacker0x01/react-datepicker/commit/40c7ee79754c979e97511c1c107749df5ad7e619, ran `yarn start`, and then tried clearing the Hero Example and hitting enter, I get NaN. Probably because `moment(null)` returns a moment with `_i: NaN` - see https://github.com/Hacker0x01/react-datepicker/blob/b1decffb4bbeabed8b0313050f0f57acc49e6208/src/datepicker.jsx#L130 and https://github.com/Hacker0x01/react-datepicker/blob/b1decffb4bbeabed8b0313050f0f57acc49e6208/src/datepicker.jsx#L275

nosykretts:
I think you should return `undefined` like this

```javascript
<DatePicker
  selected={props.value? moment(props.value) : undefined}
  onChange={date => props.onChange(date ? date.toISOString() : undefined)}
/>
```

Aprillion:
while passing `selected={null}` kind of works for me, when I press ENTER then some random date will replace whatever is in the input `value` (previously selected value / today by default)

even with `disabledKeyboardNavigation` :(

Aprillion:
1.) not possible to overwrite with a `customInput={<input onKeyDown=...}` because it's cloned in https://github.com/Hacker0x01/react-datepicker/blob/3ce7f9b1e237efb87ef8ce658db3945ec938b6c2/src/datepicker.jsx#L407  so the library `onKeyDown` takes precedence over custom props :(

2.) and this is the place that inserts "today" on ENTER - https://github.com/Hacker0x01/react-datepicker/blob/3ce7f9b1e237efb87ef8ce658db3945ec938b6c2/src/datepicker.jsx#L289

I will create a PR to check for `disabledKeyboardNavigation` to solve 2., ignoring 1. for now

[...]

PR #1198, opened 2018-01-06
Added public function for programatically clearing the input

chrisjpatty:
First of all, thank you so much for all the hard work that's gone into `react-datepicker`. It's been invaluable to our company to have such a well-made react date picker available. 

I ran into a small issue where after a date is selected, passing `null` to the `selected` prop wouldn't actually clear out the input. After looking through the code I realized I could call `onClearClick` manually to clear the input without needing to display the clear button. This works in my case but it does currently require that an event object be passed to the `onClearClick` function or it crashes. This pull request makes a small change to not require that event object. It also adds a new `clear` function that could be included in the docs as part of the public API. 

If this all looks acceptable I would be happy to make another pull request to include this method in the docs. 

Thanks again for this excellent package! It has seriously saved me so much time!

_Edit:_ I should mention, this PR at least partially addresses several open issues including: 
#1142 
#888 
#754

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -444,10 +444,18 @@ export default class DatePicker extends React.Component {
   };
 
   onClearClick = event => {
-    event.preventDefault();
-    this.props.onChange(null, event);
-    this.setState({ inputValue: null });
+  	if (event) {
+  		if (event.preventDefault) {
+  			event.preventDefault()
+  		}
+  	}
+  	this.props.onChange(null, event)
+  	this.setState({ inputValue: null })
   };
+
+  clear = () => {
+    this.onClearClick()
+  }
 
   renderCalendar = () => {
     if (!this.props.inline && (!this.state.open || this.props.disabled)) {
--- a/test/datepicker_test.js
+++ b/test/datepicker_test.js
@@ -895,4 +895,11 @@ describe("DatePicker", () => {
     );
     expect(datePicker.state.open).to.be.false;
   });
+  it("should clear the input when clear() member function is called", () => {
+    const datePicker = TestUtils.renderIntoDocument(
+      <DatePicker selected={utils.newDate("2015-12-15")} />
+    );
+    datePicker.clear();
+    expect(datePicker.state.inputValue).to.be.null;
+  });
 });


src/index.jsx
```diff
@@ -444,11 +444,19 @@ export default class DatePicker extends React.Component {
   };
 
   onClearClick = event => {
-    event.preventDefault();
-    this.props.onChange(null, event);
-    this.setState({ inputValue: null });
+  	if (event) {
+  		if (event.preventDefault) {
```
martijnrusschen: Isn't this if always true when there's an event available?

src/index.jsx
```diff
@@ -444,11 +444,19 @@ export default class DatePicker extends React.Component {
...
+  			event.preventDefault()
+  		}
+  	}
+  	this.props.onChange(null, event)
+  	this.setState({ inputValue: null })
   };
 
+  clear = () => {
```
martijnrusschen: Is this method called externally? How's this being used in this component?

src/index.jsx
chrisjpatty: Sort of, there's always an edge case where a truthy value was passed in as a parameter but it's not actually a valid event object. This just covers that case by first checking if the parameter was defined, then if it has the function you're attempting to call before calling it. Could probably be accomplished in a different way if you have an idea to revise it.

src/index.jsx
chrisjpatty: Yeah so the way I'm using it is by setting a ref on the Datepicker component like so:
```jsx
render(){
    return(
        <DatePicker ref={datePicker => {this.datePicker = datePicker}} />
    )
} 
```
and then I'm calling the clear function in a member method:
```jsx
someMemberMethod = () => {
    this.datePicker.clear()
}
```
This can currently be accomplished by setting the ref and calling `this.onClearClick()` but a method just called `clear()` seemed more appropriate for a method intended to be part of the public API.

chrisjpatty:
Also I just added a test for the `clear()` function. To be honest, writing tests isn't my strong point at the moment so hopefully I did it right. ;)

martijnrusschen:
Thanks!

PR #1199, opened 2018-01-09
Change close icon link href from # to javascript:void(0)

nassimbenkirane:
Hello ReactDatepicker maintainers,
Kudos for this library,

I use the `Datepicker` component with `isCleareable={true}`, but whenever I try to clear the datepicker, it messes with my urls, and the browser tries to go to "#" and navigates me to the home page of my app

I couldn't do anything with event.preventDefault() but changing to `javascript:void(0)` in my sources solves it. 

Would you consider merging this PR ?
Thanks,

Nassim

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -557,7 +557,7 @@ export default class DatePicker extends React.Component {
       return (
         <a
           className="react-datepicker__close-icon"
-          href="#"
+          href="javascript:void(0)"
           onClick={this.onClearClick}/>
       );
     } else {


aij:
Should it just be a `span` instead?

aij:
As is, it ends up in the tab order, with no visual indication when it has the focus, and no label nor description for screen readers.

nassimbenkirane:
Hi @aij, thanks for the answer 👍

Using a span would work indeed, but it would remove the ability to clear the input with the keyboard 🤔  Meaning it would be a regression...

Alternatively, a `<button type='button' />` also works, do you think it's better than the `<a href />` ?

Also, I like the fact that we are able to style the close button on focus (`.react-datepicker__close-icon:focus`) 

What do you think ?

aij:
Backspace/delete still work, don't they? As a regular keyboard user (though not exclusively), I would typically use those keys to clear a field (or Ctrl-A + delete), not Tab + Enter.

Is the clear button supposed to be for consistency with Edge? (I seem to remember seeing clear buttons on inputs in Windows, though I don't have a VM handy right now.) If so, do those close buttons appear in the tab order, or does tab go to the next actual field?

Anyway, I suppose we can discuss that separately... Actually going to # on clear certainly seems buggy.

nassimbenkirane:
Thanks a lot for your reactivity !

PR #1205, opened 2018-01-11
issue-1198 replaced the date-picker navigation links with buttons

jontrainor:
#1198 - replaced navigation <a> links with <button>


jontrainor:
closed this pr because it referred to the wrong issue and opened a new one

PR #1206, opened 2018-01-12
issue-1204 replaced the datepicker navigation links with buttons

### LINKED ISSUE #1204: react-datepicker__navigation--next and react-datepicker__navigation--previous are anchor tags without an href and are interfering with keen.io's event reporting
The navigation buttons are anchor tags without an href like `<a class="react-datepicker__navigation react-datepicker__navigation--next"></a>` and the event capturing platform we are using intercepts clicks on anchor tags, records them and then sends the browser to whatever is in the href attribute. 
```
function deferClickEvent(evt, anchor, callback){
  var timeout = 500,
      targetAttr,
      cbResponse;

  // Get 'target' attribute from anchor
  if (anchor.getAttribute !== void 0) {
    targetAttr = anchor.getAttribute("target");
  } else if (anchor.target) {
    targetAttr = anchor.target;
  }

  // Fire listener and catch possible response (return false)
  cbResponse = callback(evt);

  // If prevented within callback, bail:
  if (('boolean' === typeof cbResponse && cbResponse === false) || evt.defaultPrevented || evt.returnValue === false) {
    if (evt.preventDefault) {
      evt.preventDefault();
    }
    evt.returnValue = false;
    return false;
  }
  // Else if anchor doesn't kick off a new window or tab.. defer and replay the event:
  else if (targetAttr !== '_blank' && targetAttr !== 'blank' && !evt.metaKey) {
    if (evt.preventDefault) {
      evt.preventDefault();
    }
    evt.returnValue = false;
    setTimeout(function(){
      window.location = anchor.href;
    }, timeout);
  }

  return false;
}
```

So every time the navigation button in this date picker is clicked the page reloads. I've already reached out to keen.io to see if there is a way to avoid this issue from their side but I thought it was worth asking if that has to be an <a> tag? I would be happy to make a pr and change them to buttons. Thanks for the great UI element!

jontrainor:
fixes #1204

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -272,7 +272,7 @@ export default class Calendar extends React.Component {
       return;
     }
     return (
-      <a
+      <button
         className="react-datepicker__navigation react-datepicker__navigation--previous"
         onClick={this.decreaseMonth}/>
     );
@@ -297,7 +297,7 @@ export default class Calendar extends React.Component {
       classes.push("react-datepicker__navigation--next--with-today-button");
     }
 
-    return <a className={classes.join(" ")} onClick={this.increaseMonth} />;
+    return <button className={classes.join(" ")} onClick={this.increaseMonth} />;
   };
 
   renderCurrentMonth = (date = this.state.date) => {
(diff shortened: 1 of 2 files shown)


martijnrusschen:
Thanks!

jontrainor:
is there any chance this can be released in a new tag soon? thanks!

PR #1207, opened 2018-01-12
Add support for combined month-year dropdown

mnikkane:
Allow selecting month and year from one dropdown. This is convenient if date selection often goes over year boundary.


martijnrusschen:
Sweet, thanks!

PR #1211, opened 2018-01-15
Add option to allow showing month navigation as disabled,…

mnikkane:
… when no next/previous months are available.

When showDisabledMonthNavigation is set true, the month navigation buttons are always shown, but buttons disabled (non-clickable) when there is no previous/next month available. Also --disabled classname is added to buttons in disabled state allowing them to be styled.

This is bit different from forceShowMonthNavigation, which does not disable month changing.


martijnrusschen:
Great work!

PR #1216, opened 2018-01-19
Add an includeTimes prop

jennyscript-ridecell:
This option is the same idea as "includeDates" but applies to the time select. Only times that are included in the "includeTimes" list and also match up with settings like "timeIntervals" will be included. All other times are disabled.

This uses the same helper function as "excludeTimes" for implementation.


jennyscript-ridecell:
Is there any additional work needed here before it can be merged? I see a few PRs have been merged by @mnikkane since I submitted this one, so please provide feedback if you'd like to see anything else here.

I'd really like to use the datepicker on my company's application and we need this feature. Also, I've found some other bugs I'd be happy to submit PRs for after this one.

rafeememon:
Looks good, thanks!

PR #1217, opened 2018-01-19
Updated calendar features with header splitting and arrow replacement

ivstefano:
- Added features to the calendar to split the header so the user can select the month and year separately. 

- Allowed replacement of the arrows in the header with text


alex-shamshurin:
Could you provide an example - what all this is about?

martijnrusschen:
@ivstefano See question above.

martijnrusschen:
Closing due inactivity.

Issue #1248, opened 2018-02-01
Calendar position is not updated after month change


vtaits:
If calendar opened with `top` position, there is a bug after change to month with other weeks number (gifs attached below). It happens because `Calendar` stores current month in own state, and `Popper` don't update position.

### Expected behavior

![reactdatepicker-fixed](https://user-images.githubusercontent.com/4801414/35690164-3ee83946-0786-11e8-904b-39d6ab6c4d44.gif)

### Actual behavior

![reactdatepicker](https://user-images.githubusercontent.com/4801414/35690152-38621394-0786-11e8-96b4-d2eea47f30f2.gif)

`Popper` can provide `scheduleUpdate` function for manual trigger of position recount.
https://github.com/souporserious/react-popper/blob/c48416f9664a1a3ca40ad38241972b6b950f0a19/src/Popper.jsx#L199

martijnrusschen:
Should be fixed with the next release.

PR #1253, opened 2018-02-02
#1248 - update calendar position on month and year change

vtaits:
Fixed updating calendar position (#1248)

- Upgrade `react-popper` to `0.8.0`.
- Provided `scheduleUpdate` function from `react-popper` to child node of `PopperComponent`.
- Call `scheduleUpdate` from `Calendar` on every change of year or month.


AgentChris:
it almost work, the popper position is still a bit off, but great fix

PR #1273, opened 2018-02-14
added box-sizing to time-list

ivdma:
Enforce the time-list to have a default box-sizing, which is content-box.


ivdma:
screen shot 2018-02-14 at 12 03 34

Issue #1271, opened 2018-02-13
how can i change the direction of the days from right to left?


avivdevalore:
![image](https://user-images.githubusercontent.com/34057014/36158554-47a6b414-10e5-11e8-8cd1-5ee9f0a676e3.png)

I want to know if there is a way to make the days of the week and the date to start from the right side.
In Hebrew/Arabic you start from the right side. 
I thought the locale will fix it but it doesn't.
I've tried looking in the props but didn't see any relavent prop.

Thanks

avivdevalore:
Thanks @sarahmarciano.
did you changed the direction to RTL?

if i add to the calendar 'direction: RTL' then it start from the right but then if i will change the locale back to English it will stay RTL and that is no good.

sarahmarciano:
yes, indeed I've changed the direction to RTL,
what I can suggest you if you need your app to be flexible is to have two divs one with direction of rtl and one with direction of ltr and render the div depending on the locale direction
your code will look something like this: (I did not test this snippet of code but you got the idea...)

```
const rtlLocales = [
    'ae',	/* Avestan */
    'ar',   /* 'العربية', Arabic */
    'arc',  /* Aramaic */
    'bcc',  /* 'بلوچی مکرانی', Southern Balochi */
    'bqi',  /* 'بختياري', Bakthiari */
    'ckb',  /* 'Soranî / کوردی', Sorani */
    'dv',   /* Dhivehi */
    'fa',   /* 'فارسی', Persian */
    'glk',  /* 'گیلکی', Gilaki */
    'he',   /* 'עברית', Hebrew */
    'ku',   /* 'Kurdî / كوردی', Kurdish */
    'mzn',  /* 'مازِرونی', Mazanderani */
    'nqo',  /* N'Ko */
    'pnb',  /* 'پنجابی', Western Punjabi */
    'ps',   /* 'پښتو', Pashto, */
    'sd',   /* 'سنڌي', Sindhi */
    'ug',   /* 'Uyghurche / ئۇيغۇرچە', Uyghur */
    'ur',   /* 'اردو', Urdu */
    'yi'    /* 'ייִדיש', Yiddish */
];

function isLocaleRtl(locale) {
    return rtlLocales.indexOf(locale.split('-')[0]) !== -1;
}

render() {
        const { locale } = this.props;
        if (isLocaleRtl(locale)) {
            return (
                <div dir='rtl'>
                    <DatePicker
                        selected={date}
[...]

PR #1274, opened 2018-02-14
Fixing the arrows in rtl languages

sarahmarciano:
When using this datepicker component with rtl languages the two arrows for moving from month to month are inverted
so I've just added a little check to check if the locale is an rtl language (hope I've included all of them feel free to add more if you know more rtl languages) and then decide if the arrow shall increase the month or decrease it.
I've also changed the functions names since each function is rendering the right/left arrow and not the prev/next month.
see my comment on this issue: https://github.com/Hacker0x01/react-datepicker/issues/1271

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -32,6 +32,8 @@ import {
   getEffectiveMaxDate
 } from "./date_utils";
 
+import { isLocaleRtl } from "./locale_utils";
+
 const DROPDOWN_FOCUS_CLASSNAMES = [
   "react-datepicker__year-select",
   "react-datepicker__month-select",
@@ -304,7 +306,7 @@ export default class Calendar extends React.Component {
     ];
 
     let clickHandler;
-	if(this.props.locale === 'he' || this.props.locale === 'ar') {
+	if(isLocaleRtl(this.props.locale)) {
 		clickHandler = this.increaseMonth;
 	} else {
 		clickHandler = this.decreaseMonth;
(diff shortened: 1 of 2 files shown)


src/calendar.jsx
```diff
@@ -303,7 +303,12 @@ export default class Calendar extends React.Component {
       "react-datepicker__navigation--previous"
     ];
 
-    let clickHandler = this.decreaseMonth;
+    let clickHandler;
+	if(this.props.locale === 'he' || this.props.locale === 'ar') {
```
rafeememon: Instead of hardcoding these I think we should do some kind of detection.  Does the ltr/rtl flag exist in moment somewhere?

sarahmarciano: I also tried to find in moment.js a function that detect if a locale is rtl or ltr but I didn't find
instead you can maybe use this library: https://github.com/shadiabuhilal/rtl-detect
or build your own util function and have something like this: https://stackoverflow.com/questions/4192382/is-there-any-way-to-detect-an-rtl-language-in-java

rafeememon: Maybe we can steal the list of locales from https://github.com/shadiabuhilal/rtl-detect/blob/master/lib/rtl-detect.js#L137.  I'd also pull that into a util file and encapsulate it as `isLocaleRtl`

martijnrusschen:
Build seems to be failing. Also, any way to add a test for this behaviour?

martijnrusschen:
Closing due inactvity

PR #1294, opened 2018-03-05
Add formatWeekDay prop for weekdays custom formatting

andrewdormi:
Sorry for large diff, i think this is prettier changes

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -23,6 +23,7 @@ import {
   isBefore,
   isAfter,
   getLocaleData,
+  getFormattedWeekdayInLocale,
   getWeekdayShortInLocale,
   getWeekdayMinInLocale,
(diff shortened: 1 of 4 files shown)


martijnrusschen:
Thanks for submitting this, can you expand a bit on what you're changing here and why? No worries about the large diff, it's doable. :)

andrewdormi:
Here is a new formatWeekDay prop for function that format weekday.

For example if you need to show only fist symbol of weekday:
```javascript
<ReactDatePicker formatWeekDay={day => day[0]} />
```

Sometimes i need to change case of weekday. Now i can do it without css changes
```javascript
<ReactDatePicker formatWeekDay={day => day.toLowerCase()} />
```

martijnrusschen:
Thanks, for the info!

PR #1300, opened 2018-03-06
Manual editing / zone related bugfixes

### LINKED ISSUE #1295: After setting time or manual input can no longer inject new moment
bug

### Expected behavior
Should be able to manually edit time and then inject a new moment.

### Actual behavior
Editing the time manually or through the UI doesn't let you change the input through props.

### Steps to reproduce
```
import React, { Component } from 'react';
import logo from './logo.svg';
import './App.css';
import 'react-datepicker/dist/react-datepicker.css';
import moment from 'moment';

class App extends Component {

  constructor(props){
    super(props);
    this.state = {
      selected: moment().subtract(1,'months')
    };
    this.format = "LLL";
  }

  render() {
    return (
      <div className="App">
        <header className="App-header">
          <img src={logo} className="App-logo" alt="logo" />
          <h1 className="App-title">Welcome to React</h1>
        </header>
        <p className="App-intro">
          To get started, edit <code>src/App.js</code> and save to reload.
        </p>
        <DatePicker
            dateFormat='MMMM DD, YYYY h:mm A'
            selected={this.state.selected}
            allowSameDay
            onChange={this.onChange}
            showTimeSelect
            timeIntervals={15}
            disabledKeyboardNavigation
        />
        <button type="button" onClick={this.onClick}>Change time to now</button>
      </div>
    );
  }

  onClick = () => {
    this.setState({selected: moment()});
  }

  onChange = (date) => {
    console.log(date.format(this.format));
    this.setState({selected: date});
  }
}

export default App;
```

### LINKED ISSUE #1298: UTC offset lost on manual editing
### Expected behavior
Inputing a selected moment with some utc offset without zone formatting should work

### Actual behavior
It will instead parse it as a local time.

### LINKED ISSUE #1299: manual editing with zones in date format does not work
Use a format with the "Z" character which shows UTC offset, upon manually editing of the string it will fail to parse and think it's an invalid moment.

tomtwttr:
Fixes #1299
Fixes #1298
Fixes #1295

datepicker_test.js has a test where it was inputing in one format and expecting it to be formatted in a different way. I don't know how this was working at all, but with my changes I couldn't find a good way to support it. I don't think the use-case is a good one though, I don't think the datepicker should support this IMO.

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -67,8 +67,14 @@ export function cloneDate(date) {
   return date.clone();
 }
 
-export function parseDate(value, { dateFormat, locale }) {
-  const m = moment(value, dateFormat, locale || moment.locale(), true);
+export function parseDate(value, { dateFormat, locale, selected }) {
+  let m = null;
+  if (dateFormat.indexOf("Z") !== -1)
+    m = moment.parseZone(value, dateFormat, locale || moment.locale(), true);
+  else {
+    m = moment(value, dateFormat, locale || moment.locale(), true);
+    if (selected) m.utcOffset(selected.utcOffset(), true);
+  }
   return m.isValid() ? m : null;
 }
 
@@ -246,17 +252,17 @@ export function equals(date1, date2) {
 
 export function isSameYear(date1, date2) {
   if (date1 && date2) {
-      return date1.isSame(date2, "year");
+    return date1.isSame(date2, "year");
   } else {
-      return !date1 && !date2;
+    return !date1 && !date2;
   }
 }
 
 export function isSameMonth(date1, date2) {
   if (date1 && date2) {
-      return date1.isSame(date2, "month");
+    return date1.isSame(date2, "month");
   } else {
-      return !date1 && !date2;
+    return !date1 && !date2;
   }
 }
 
--- a/test/datepicker_test.js
+++ b/test/datepicker_test.js
@@ -743,14 +743,11 @@ describe("DatePicker", () => {
   it("should preserve user input as they are typing", () => {
     const onChange = date => datePicker.setProps({ selected: date });
     const datePicker = mount(
-      <DatePicker
-        dateFormat={["YYYY-MM-DD", "MM/DD/YYYY", "MM/DD/YY"]}
-        onChange={onChange}
-      />
+      <DatePicker dateFormat={"YYYY-MM-DD"} onChange={onChange} />
     );
     expect(datePicker.find("input").prop("value")).to.equal("");
 
-    const str = "12/30/1982";
+    const str = "1982-12-30";
     str.split("").forEach((c, i) => {
       datePicker.find("input").simulate("change", {
         target: { value: datePicker.find("input").prop("value") + c }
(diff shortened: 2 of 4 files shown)


src/date_utils.js
```diff
@@ -67,8 +67,14 @@ export function cloneDate(date) {
   return date.clone();
 }
 
-export function parseDate(value, { dateFormat, locale }) {
-  const m = moment(value, dateFormat, locale || moment.locale(), true);
+export function parseDate(value, { dateFormat, locale, selected }) {
+  let m = null;
```
martijnrusschen: I think we should change the `m` variable to a more descriptive name. Any suggestions?

test/datepicker_test.js
```diff
@@ -743,14 +743,11 @@ describe("DatePicker", () => {
   it("should preserve user input as they are typing", () => {
     const onChange = date => datePicker.setProps({ selected: date });
     const datePicker = mount(
-      <DatePicker
-        dateFormat={["YYYY-MM-DD", "MM/DD/YYYY", "MM/DD/YY"]}
```
aij: This test was meant to avoid regressions of https://github.com/Hacker0x01/react-datepicker/issues/560#issuecomment-248445794

(example in https://github.com/Hacker0x01/react-datepicker/issues/560#issuecomment-272533692)

test/datepicker_test.js
```diff
@@ -743,14 +743,11 @@ describe("DatePicker", () => {
...
-        onChange={onChange}
-      />
+      <DatePicker dateFormat={"YYYY-MM-DD"} onChange={onChange} />
     );
     expect(datePicker.find("input").prop("value")).to.equal("");
 
-    const str = "12/30/1982";
+    const str = "1982-12-30";
```
aij: Note that no prefix of this new string is an ambiguous format for the new date format above.

aij:
Is this

> datepicker_test.js has a test where it was inputing in one format and expecting it to be formatted in a different way.

supposed to be in reference to

     expect(
       utils.formatDate(datePicker.prop("selected"), "YYYY-MM-DD")
     ).to.equal("1982-12-30");
?

That's not expecting the *input* to be formatted in a different way. It's expecting the resulting *selected value* to have the same date as the date that was entered.

Ie, if "12/30/1982" is entered in a datepicker with `dateFormat` set to parse any of `["YYYY-MM-DD", "MM/DD/YYYY", "MM/DD/YY"]`, then that input should be interpreted as the 30th day of December in the year 1982. (Which means reformatting it as an ISO date should result in "1982-12-30".)

tomtwttr:
Thank you very much for the feedback @aij and @martijnrusschen . For my use case I've been able to remove the need to use moment utc related info. I'm basically wrapping the DatePicker component with my own component and using a unix() based interface, therefore ignoring all of the weird moment zone related cases. In general it feels moment itself handles this a bit unexpectedly with the separate parsing functions. I will be closing this in favor for a small PR without any zone related changes.

tomtwttr:
Opened https://github.com/Hacker0x01/react-datepicker/pull/1303

PR #1303, opened 2018-03-07
Fix for time selection not allowing other changes

### LINKED ISSUE #1295: After setting time or manual input can no longer inject new moment

tomtwttr:
Fixes #1295

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "author": "HackerOne",
   "name": "react-datepicker",
   "description": "A simple and reusable datepicker component for React",
-  "version": "1.2.4",
+  "version": "1.2.2",
   "license": "MIT",
   "homepage": "https://github.com/Hacker0x01/react-datepicker",
   "main": "lib",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "author": "HackerOne",
   "name": "react-datepicker",
   "description": "A simple and reusable datepicker component for React",
-  "version": "1.2.2",
+  "version": "1.2.4",
   "license": "MIT",
   "homepage": "https://github.com/Hacker0x01/react-datepicker",
   "main": "lib",
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -173,6 +173,8 @@ export default class DatePicker extends React.Component {
         highlightDates: getHightLightDaysMap(nextProps.highlightDates)
       });
     }
+    if(!this.state.focused)
+      this.setState({ inputValue: null });
   }
 
   componentWillUnmount() {
@@ -206,7 +208,8 @@ export default class DatePicker extends React.Component {
         : boundedPreSelection,
       // transforming highlighted days (perhaps nested array)
       // to flat Map for faster access in day.jsx
-      highlightDates: getHightLightDaysMap(this.props.highlightDates)
+      highlightDates: getHightLightDaysMap(this.props.highlightDates),
+      focused: false
     };
   };
 
@@ -239,6 +242,7 @@ export default class DatePicker extends React.Component {
         this.setOpen(true);
       }
     }
+    this.setState({focused: true});
   };
 
   cancelFocusInput = () => {
@@ -261,6 +265,7 @@ export default class DatePicker extends React.Component {
     } else {
       this.props.onBlur(event);
     }
+    this.setState({focused: false});
   };
 
   handleCalendarClickOutside = event => {
@@ -372,7 +377,7 @@ export default class DatePicker extends React.Component {
 
     this.props.onChange(changedDate);
     this.setOpen(false);
-    this.setState({ inputValue: changedDate.format(this.props.dateFormat) });
+    this.setState({ inputValue: null });
   };
 
   onInputClick = () => {
--- a/test/timepicker_test.js
+++ b/test/timepicker_test.js
@@ -7,20 +7,22 @@ import Time from "../src/time";
 
 describe("TimePicker", () => {
   let datePicker;
+  let div;
+  let onChangeMoment;
+
+  beforeEach(() => {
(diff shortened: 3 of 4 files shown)


src/index.jsx
```diff
@@ -173,6 +173,7 @@ export default class DatePicker extends React.Component {
         highlightDates: getHightLightDaysMap(nextProps.highlightDates)
       });
     }
+    this.setState({ inputValue: null });
```
aij: I don't think it's a good idea to discard user input on *every* prop change.

As you can see from the failed test, this could result in discarding user input as they are typing.

test/timepicker_test.js
```diff
@@ -7,20 +7,23 @@ import Time from "../src/time";
...
-    let onChangeMoment = null;
-    const onChange = m => {
-      onChangeMoment = m;
-    };
-    makeDatePicker(onChange);
+    renderDatePicker("February 28, 2018 4:43 PM");
     expect(getInputString()).to.equal("February 28, 2018 4:43 PM");
     setManually("February 28, 2018 4:45 PM");
-    expect(onChangeMoment.format("LLL")).to.equal("February 28, 2018 4:45 PM");
+    expect(getInputString()).to.equal("February 28, 2018 4:45 PM");
```
aij: This test used to check the moment passed to `onChange`. Why shouldn't it do that any more?

test/timepicker_test.js
```diff
@@ -7,20 +7,23 @@ import Time from "../src/time";
 
 describe("TimePicker", () => {
   let datePicker;
+  let div;
+  let format;
```
aij: Why should this be a variable? Isn't it always going to be "LLL"?

src/index.jsx
```diff
@@ -372,7 +373,7 @@ export default class DatePicker extends React.Component {
 
     this.props.onChange(changedDate);
     this.setOpen(false);
-    this.setState({ inputValue: changedDate.format(this.props.dateFormat) });
+    this.setState({ inputValue: null });
```
aij: This does seem like a good idea. :)

package.json
```diff
@@ -2,7 +2,7 @@
   "author": "HackerOne",
   "name": "react-datepicker",
   "description": "A simple and reusable datepicker component for React",
-  "version": "1.2.2",
+  "version": "1.2.3",
```
aij: We usually bump the version when doing a release, not with every PR.

tomtwttr:
Hey @aij, thanks for the feedback. I've updated the PR and I think this solution works nicely.

src/index.jsx
tomtwttr: To be clear, I think this is correct. This was actually my addition in https://github.com/Hacker0x01/react-datepicker/commit/a72bc43be63a00f8719361256c18466df7c2fb58#diff-73642d587163f2eb4d72f0fa6ebfba02R375.

Before that change the time would not change at all. The reason is that the input state side effect ignores changes. Clearing the input state makes sense if you're selecting a time.

test/timepicker_test.js
```diff
@@ -29,23 +31,42 @@ describe("TimePicker", () => {
...
   }
 
-  function makeDatePicker(onChange) {
-    datePicker = TestUtils.renderIntoDocument(
+  function renderDatePicker(string) {
+    return renderDatePickerFor(moment(string, "LLL", true));
+  }
+
+  function renderDatePickerFor(selected) {
+    datePicker = ReactDOM.render(
```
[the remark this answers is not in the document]
tomtwttr: The is needed such that the same component will be used AKA componentWillReceiveProps. A TestUtils.renderIntoDocument will not work for that kind of test.

package.json
```diff
@@ -2,7 +2,7 @@
   "author": "HackerOne",
   "name": "react-datepicker",
   "description": "A simple and reusable datepicker component for React",
-  "version": "1.2.2",
+  "version": "1.2.4",
```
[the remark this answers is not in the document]
tomtwttr: Is it possible to leave this and let this bump to 1.2.5? Internally here I published this version a private registry, and I don't know if I can overwrite this version with the one you guys do.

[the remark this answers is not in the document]
rafeememon: No, sorry; please leave the versioning to us and our build process.

tomtwttr:
Reverted the version change as requested by @rafeememon

rafeememon:
Thanks @tomtwttr

tomtwttr:
You're welcome

PR #1307, opened 2018-03-08
Convert week_number.jsx to flow (as a simple example)

aij:
This is a pretty minimal conversion. We could make `onClick` and `handleClick` be functions that take a `SyntheticEvent` for stricter checking.

### DIFF
--- a/src/week_number.jsx
+++ b/src/week_number.jsx
@@ -1,14 +1,18 @@
+// @flow
 import React from "react";
 import PropTypes from "prop-types";
 import classnames from "classnames";
 
-export default class WeekNumber extends React.Component {
+export default class WeekNumber extends React.Component<{
+  weekNumber: number,
+  onClick?: Function
+}> {
   static propTypes = {
     weekNumber: PropTypes.number.isRequired,
     onClick: PropTypes.func
   };
 
-  handleClick = event => {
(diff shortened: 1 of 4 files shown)


src/week_number.jsx
```diff
@@ -1,14 +1,18 @@
...
 
-export default class WeekNumber extends React.Component {
+export default class WeekNumber extends React.Component<{
+  weekNumber: number,
+  onClick?: Function
+}> {
   static propTypes = {
     weekNumber: PropTypes.number.isRequired,
     onClick: PropTypes.func
   };
```
aij: If we were to use babel-plugin-flow-react-proptypes, these `propTypes` could be removed as they would be automatically generated from the types above.

martijnrusschen:
Nice work 👍

martijnrusschen:
Looks like the CI integration works, shall we go ahead and merge this?

aij:
@martijnrusschen Sure, this should be fine to merge.

I was toying around with webpack integration, but unsure where it would be best. `server.js` and `webpack.docs.config.js` are two options. Is `webpack.config.js` actually used anywhere?

martijnrusschen:
Good question, need to dive into that. I was planning on upgrading webpack as well, but haven't got the time to make this happen.

PR #1313, opened 2018-03-14
added renderChildrenForDate prop

aybb:
Add prop `renderChildrenForDate` which will allow to render any component in addition to day of month.


martijnrusschen:
Can you expand a bit on the use case here?

aybb:
@martijnrusschen I just need to add text for date. See screenshot.

klis87:
Very nice idea, I also need this to show some tooltips for given dates, with customer renderer this would be trivial.

martijnrusschen:
Can we get the merge conflicts resolved so I can merge this?

jRichardeau:
I also need this to customize cell content !
But why a render prop that only adds children to date ? I think a better idea would be to have a render prop for all date cell, isn't it ?

PR #1317, opened 2018-03-16
Adds title attribute to clear button

nateha1984:
To resolve an accessibility issue, a title attribute was added to the clear button that will read and display as a tooltip "Clear date"

### DIFF
--- a/docs/datepicker.md
+++ b/docs/datepicker.md
@@ -10,6 +10,7 @@ General datepicker component.
 | `calendarClassName`          | `string`                       |                 |                                            |
 | `children`                   | `node`                         |                 |                                            |
 | `className`                  | `string`                       |                 |                                            |
+| `clearButtonTitle`           | `string`                       |                 |                                            |
 | `customInput`                | `element`                      |                 |                                            |
 | `customInputRef`             | `string`                       | `'ref'`         | The property used to pass the ref callback |
 | `dateFormat`                 | `union(string\|array)`         | `'L'`           |                                            |
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -128,7 +128,8 @@ export default class DatePicker extends React.Component {
     minTime: PropTypes.object,
     maxTime: PropTypes.object,
     excludeTimes: PropTypes.array,
-    useShortMonthInDropdown: PropTypes.bool
+    useShortMonthInDropdown: PropTypes.bool,
+    clearButtonTitle: PropTypes.string
   };
 
   static get defaultProps() {
@@ -588,7 +589,7 @@ export default class DatePicker extends React.Component {
           className="react-datepicker__close-icon"
           href="javascript:void(0)"
           onClick={this.onClearClick}
-          title="Clear date"/>
+          title={this.props.clearButtonTitle}/>
       );
     } else {
       return null;
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -587,7 +587,8 @@ export default class DatePicker extends React.Component {
         <a
           className="react-datepicker__close-icon"
           href="javascript:void(0)"
-          onClick={this.onClearClick}/>
+          onClick={this.onClearClick}
+          title="Clear date"/>
       );
     } else {
       return null;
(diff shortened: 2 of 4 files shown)


src/index.jsx
```diff
@@ -587,7 +587,8 @@ export default class DatePicker extends React.Component {
         <a
           className="react-datepicker__close-icon"
           href="javascript:void(0)"
-          onClick={this.onClearClick}/>
+          onClick={this.onClearClick}
+          title="Clear date"/>
```
martijnrusschen: If you want to continue with adding a text to the button, I'd recommend you let the user configure the text. Also, adding some tests would be appreciated :-)

nateha1984: Made the button title configurable and added a test

nateha1984:
@aij I updated the text to be user configurable so a dev can set any text they deem appropriate as the title, or no text at all and the attribute will not show up.

As for who will benefit, the change is to allow screen readers such as JAWS to pick and read the intention of the button back to a visually-disabled user. The button can be accessed by tabbing through the control (e.g., put the cursor in the date field, hit tab once, then hit enter and the field will clear) and without any sort of title or aria-label attribute a screen reader won't indicate the purpose of the button to the user.

nateha1984:
Looks like I still need to increase test coverage a tad, will work on that...

aij:
@nateha1984 Oh, that button didn't used to appear in the tab order, but an href was added recently.

It seems like it would be more accessibility friendly not to burden keyboard users and screen reader users with extra complexity  which is only really useful to mouse and touchscreen users.

Even with a title the clear button still has accessibility issues for sighted keyboard users. It interferes with the tab order and it gives no visual indication that the close button is where the focus is.

nateha1984:
@aij I'm not sure I understand how there is extra complexity added to have the purpose of a button read out by a screen reader or to show a tool tip on hover. In fact, it's necessary for users with assistive devices to understand what is going on on the screen. Users who do not need extra assistance won't even know there's any change unless they hover on the button.

I do agree that there does need to be some visual styling added if the button is going to be accessible from the tab order. I did go ahead and remove the button from the tab order as you are correct in that it shouldn't be part of the tabbing since it's part of the control; tab should take the user to the next component/control/whatever. I also changed the element to be a button type instead of an anchor.

martijnrusschen:
@aij Let me know once this is ready to merge.

docs/datepicker.md
```diff
@@ -10,6 +10,7 @@ General datepicker component.
 | `calendarClassName`          | `string`                       |                 |                                            |
 | `children`                   | `node`                         |                 |                                            |
 | `className`                  | `string`                       |                 |                                            |
+| `clearButtonTitle`           | `string`                       |                 |                                            |
```
aij: This should get regenerated during the release process

test/week_number_test.js
```diff
@@ -1,20 +1,31 @@
...
 import WeekNumber from "../src/week_number";
 import { shallow } from "enzyme";
+import sinon from "sinon";
 
 function renderWeekNumber(weekNumber, props = {}) {
   return shallow(<WeekNumber weekNumber={weekNumber} {...props} />);
 }
 
 describe("WeekNumber", () => {
+  let shallowWeekNumber;
```
aij: Minor nitpick: Sharing this across the two tests and making it mutable seems a little weird, especially given that both tests will just reassign it first thing.

test/week_number_test.js
```diff
@@ -1,20 +1,31 @@
...
+      shallowWeekNumber = renderWeekNumber(weekNumber);
       expect(
         shallowWeekNumber.hasClass("react-datepicker__week-number")
       ).to.equal(true);
       expect(shallowWeekNumber.text()).to.equal(weekNumber + "");
     });
+
+    it("should call the onClick function if it is defined", () => {
+      const onClick = sinon.spy();
+      shallowWeekNumber = shallow(
```
aij: This could just be a separate local variable.

jshih7:
Sorry to piggyback off of this issue, but I have a question about this change.

My understanding is that the original intention was to add a title to the button so that screen readers are able to read it.

However, if we are removing the button from the tab order, wouldn't this lead to an inconsistency where screen readers are now reading this button's title, but it is not tab accessible, making it more confusing for the visually-disabled?

Let me know if this is concern, and I'll create a new issue.

PR #1325, opened 2018-03-21
fix: Updated react-popper version

greg-a-smith:
Updating react-popper to 0.8.3.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "classnames": "^2.2.5",
     "prop-types": "^15.6.0",
     "react-onclickoutside": "^6.7.1",
-    "react-popper": "^0.8.2"
+    "react-popper": "^0.8.3"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -107,7 +107,8 @@
       "cat coverage/*/lcov.info | ./node_modules/codecov.io/bin/codecov.io.js",
     "prebuild": "rimraf es lib dist",
     "build": "cross-env NODE_ENV=production run-p build:** && run-p css:**",
-    "build-dev": "cross-env NODE_ENV=development run-p build:** && run-p css:**",
```
martijnrusschen: This change seems not intentional

greg-a-smith: Agreed -- not sure where that came from.

greg-a-smith: Okay, what's going on here?  I went back to explicitly revert this and it showed it as reverted, but after committing and pushing it, it keeps changing it to this format.

martijnrusschen: Looks like both are the same now. I'm fine with this.

Issue #1283, opened 2018-02-27
Suggestion: support Luxon


trevordmiller:
[Luxon](https://github.com/moment/luxon) is the new, leaner version of Moment.js. Have you considered supporting it so moment.js isn't required?

Architektor:
Looking forward to see that too

aij:
This seems like a special-case of #382 .

@prayerslayer has already moved all the moment code to `src/date_utils.js` in #1046 and #1052, so it should now be a fairly simple*** matter of implementing that API on top of Luxon.

I expect PRs would be welcome if anyone cares to support Luxon.

***: No guarantees as to how simple it will actually be.

prayerslayer:
Yeah, there's also a half-baked POC in #1059 to support any date library as backend. However straight-up replacing `moment` is simpler than adding `luxon` support, because we don't need to care about changing the build process and how to distribute `react-datepicker` with exchangeable date backends.

We could replace `moment`, release a new major version, then check if `date-fns` support is still desirable.

trevordmiller:
Ah yeah 382 looks great. Closing this in favor of that!

PR #1327, opened 2018-03-22
WIP: Replace moment with luxon

prayerslayer:
Hi, this is work-in-progress. I just wanted to get the ball rolling on #1283 / #382 and prevent someone else from doing duplicate work. Is there a `dev` branch or similar? I wouldn't want to merge this in master.

Plan: 

- [x] Add luxon
- [x] Refactor `date_utils` to work with that and make its tests pass
- [x] Figure out localization & formatting
- [x] Update examples and docs (<strike>do separately</strike> I actually need this to test code correctness)
- [ ] Revert cosmetic changes in examples, apparently my `prettier` config is different, oops
- [ ] Remove `cloneDate`
- [ ] ?????
- [ ] Profit


aij:
Shouldn't most of these changes be in `date_utils.js`, or rather a new,  luxon-based version of that file?

For example, rather than removing `cloneDate`, the luxon version of that would just be the identity function.

Or is the plan to drop support for moment?

aij:
Fine by me. :smiley: It will be a major breaking change, but at CCAP we're handling *most* conversions to/from moment in a single place, which should be relatively easy to switch to luxon. I don't know how common that is though. 

How long would we expect to maintain a 1.x branch while users transition?

prayerslayer:
> How long would we expect to maintain a 1.x branch while users transition?

I don't have experience running a popular library, so I don't really know what a time frame would be. 3–6 months of backporting bugfixes, but not adding features feels alright to me. Maybe we could release a codemod to make the transition easier? The necessary changes are mainly mechanic, subtle differences exist though.

martijnrusschen:
> If you add luxon support alongside moment, yes, but when I replace moment I need to change all the examples, which contain userland code :)

I have no clear picture of the usage of Luxon vs Moment currently but would it be an idea to implement it in a way that we can run Luxon next to moment? Giving the users the flexibility to use whatever library they want to use (Luxon, Moment, Pure JS dates, etc.) Another route would be to remove moment altogether (something like #1059).

prayerslayer:
> I have no clear picture of the usage of Luxon vs Moment currently

If NPM download stats are worth anything, moment is ~240x as popular 🙃 

> would it be an idea to implement it in a way that we can run Luxon next to moment? Giving the users the flexibility to use whatever library they want to use (Luxon, Moment, Pure JS dates, etc.)

I mean yeah, that technically works and sounds nice, but there are a couple of things to consider:

* Build & distribution: I guess you could add a property to `react-datepicker` which specifies the date backend, and then import that async as needed.
* Differences in implementations: JS Dates and `date-fns` don't support timezones, `moment` has a mutable API (vs immutable of `luxon` and `date-fns`), `weekday` depends on locale in `moment` but doesn't in `luxon`, only `moment` has custom locales etc. For me an outcome of #1059 was that either the lowest common denominator of all possible date backends is implemented, or only very similar date backends are supported. Otherwise this grows too complex to maintain *and* use.
* Documentation: Basically there'll be one set of documentation per date backend, definitely more effort to maintain than now. Also educating developers about the new possibility/complexity will take more time than just saying "use luxon from <some point in the future> on."
* Bugfixes: You now need to test them with all the date backends.

Personally I think that's too much, but it's up to you to decide I guess. I'm fine with whatever decision you make :)

acontreras89:
My two cents (with barely 5 minutes of library usage):

* :+1: to removing `moment` altogether
* :heart: raw js dates, `date-fns` for :art:

I'd leave out everything else for users to implement. Although I haven't worked with `luxon`, no idea how that is

SampsonCrowley:
Id definitely be on the Raw JS dates bandwagon. I hate including extra libraries where I van avoid it. and ES6/7 makes working with regular dates pretty simple Id be happy to help out if someone has something specific that needs work

martijnrusschen:
Moment was removed in #1527.

Issue #1328, opened 2018-03-26
Feature Proposal: Ability to inject specific times outside of interval


jimmynguyc:
Hi first of all thanks for the great library :)

Right now the times shown are only controlled via the `timeIntervals` option, which makes a lot of sense. But sometimes I want to show my users a special time (e.g. `23:59 PM` for end of day). 

I would like to propose a new feature whereby specific times can be injected into the list of time via a new property `injectTimes`. 

Usage :- 

```javascript
<DatePicker
    selected={this.state.startDate}
    onChange={this.handleChange}
    showTimeSelect
    timeFormat="HH:mm"
    injectTimes={[moment().hours(23).minutes(59)]}
    dateFormat="LLL"
/>
```

Resulting in something like this 

<img width="295" alt="screen shot 2018-03-26 at 22 48 56" src="https://user-images.githubusercontent.com/377774/37913638-f387f31a-3147-11e8-81f0-0ed3fc24e337.png">

I'll submit a PR with this feature.

martijnrusschen:
Merged the PR, closing this

TEMHITHORPHE:
> Is there a way to only show the specific times I inject, and no other times? Like if i only want 11:59pm, and 12:05am, and thats all? Like exclude all other times other than the injected ones?

Please were you able to resolve this?

coltanium13:
> 
> > Is there a way to only show the specific times I inject, and no other times? Like if i only want 11:59pm, and 12:05am, and thats all? Like exclude all other times other than the injected ones?
> 
> Please were you able to resolve this?

I think I used CSS to hide specific times I did not want to see, and the interval property (I think). It's been a while and I don't have that code base to refer to anymore.

PR #1329, opened 2018-03-26
Ability to inject times outside of default time interval

jimmynguyc:
Proposed feature according to https://github.com/Hacker0x01/react-datepicker/issues/1328


martijnrusschen:
Nice work!

PR #1330, opened 2018-03-27
Improvements to injectTimes

jimmynguyc:
Following up on https://github.com/Hacker0x01/react-datepicker/pull/1329
Feature proposed at https://github.com/Hacker0x01/react-datepicker/issues/1328

Added ability to :-
- Insert multiple times in between each interval. Previously only one injected time allowed per interval.
screen shot 2018-03-27 at 13 46 25

- Auto sort injected times so that they appear chronologically
screen shot 2018-03-27 at 13 49 02


martijnrusschen:
Again, nice work!

PR #1331, opened 2018-03-27
Calculate initial scroll of TimeSelect

honzajerabek:
Hi, I've come up to a problem with Time component when using complete custom stylesheet - the selected time wasn't centered properly because of some hardcoded height values, so I made some tweaks to calculate the scroll needed in order to have selected time scrolled in the middle of the list.

basically it works like this pseudocode:
`listScrollNeeded = centerLi.offsetTop - (list.height/2 - centerLi.height/2)`
 where `list` is previously added ref to `<ul>` and `centerLi` is newly added ref to `<li>`, which matches hours and minutes of the `selectedDate` prop, or fallbacks to at least `currH` (either `selectedDate` or `new Date()`) - just like before.

Was also wondering about adding tests, but supposedly, enzyme doesn't support DOM properties (like `height`, or `scrollTop`), as it doesn't really "render" anything.

Maybe you have some ideas? :)

ps. prettier made pretty it ugly 😆

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -130,7 +130,7 @@ export default class Calendar extends React.Component {
     this.state = {
       date: this.localizeDate(this.getDateInView()),
       selectingDate: null,
-      monthContainer: this.monthContainer
+      monthContainer: null
     };
   }
 
@@ -552,7 +552,10 @@ export default class Calendar extends React.Component {
   };
 
   renderTimeSection = () => {
-    if (this.props.showTimeSelect) {
+    if (
+      this.props.showTimeSelect &&
+      (this.state.monthContainer || this.props.showTimeSelectOnly)
+    ) {
       return (
         <Time
           selected={this.props.selected}
--- a/src/time.jsx
+++ b/src/time.jsx
@@ -38,13 +38,20 @@ export default class Time extends React.Component {
     };
   }
 
+  static calcCenterPosition = (listHeight, centerLiRef) => {
+    return (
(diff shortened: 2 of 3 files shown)


honzajerabek:
is this project abandoned?

martijnrusschen:
It's not, but it's driven by volunteers who all have a full time job as well. I'm sorry for the delay. Can you perhaps add a test for your fix, so that we won't break this again in the future?

honzajerabek:
I know @martijnrusschen, I was just wondering that there isn't any reply on any of open PRs for a month. As I said, I was thinking about adding tests, but without any success.. I've been testing this fork on production in the meantime.

weyert:
I noticed a new release was created, any reason why this PR didn't make the cut?

martijnrusschen:
No reason, but happy to cut another release. It looks like some checks are failing on code coverage. Any chance we can get that fixed?

honzajerabek:
Hi, sorry for delay, I've finally got back to this and made some tests, even with live code and I've found out that there's another issue with `Time` component height - there's a `height` property passed from `monthRef` with some hardcoded value, which is imo unnecessary and can be solved with CSS. I'll have some time during weekend and have a deeper look at it, maybe it could be simplified. I'll finish the tests later and push them, but that height issue will be a blocker for now..

honzajerabek:
@martijnrusschen hi, I've done tests and removed another hardcoded height value, being now used as computed property from the node. I'm still seeing many hardcoded height px values in the stylesheet, I might have a look at it and post another PR later, I'm using custom stylesheet, made with flexbox :)

PR #1346, opened 2018-04-04
Bk fix year pre selection bug, fixes #1345

### LINKED ISSUE #1345: Changes on the Year do not affect the the Calendar selection
In Inline Mode, change the year of a date

### Expected behavior

The calendar switches to the correct Month+Year

### Actual behavior

The calendar selection does not update

### Steps to reprodcue

Select a Date in the Datepicker, change the year in some other input.

The problem is, that only a month-change will trigger the selection update.

https://github.com/Hacker0x01/react-datepicker/blob/master/src/index.jsx#L169

webdeb:
fixes #1345

### DIFF
--- a/docs-site/src/examples/inline.jsx
+++ b/docs-site/src/examples/inline.jsx
@@ -35,7 +35,8 @@ export default class Inline extends React.Component {
           <DatePicker
             inline
             selected={this.state.startDate}
-            onChange={this.handleChange}/>
+            onChange={this.handleChange}
+          />
         </div>
       </div>
     );
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -15,7 +15,6 @@ import {
   getSecond,
   getMinute,
   getHour,
-  getMonth,
   addDays,
   addMonths,
   addWeeks,
@@ -31,12 +30,25 @@ import {
   getEffectiveMaxDate,
   parseDate,
   safeDateFormat,
-  getHightLightDaysMap
+  getHightLightDaysMap,
+  getYear,
+  getMonth
 } from "./date_utils";
 import onClickOutside from "react-onclickoutside";
 
 const outsideClickIgnoreClass = "react-datepicker-ignore-onclickoutside";
 const WrappedCalendar = onClickOutside(Calendar);
+
+// Compares dates year+month combinations
+function hasPreSelectionChanged(date1, date2) {
+  if (date1 && date2) {
+    return (
+      getMonth(date1) !== getMonth(date2) || getYear(date1) !== getYear(date2)
+    );
+  }
+
+  return date1 !== date2;
+}
 
 /**
  * General datepicker component.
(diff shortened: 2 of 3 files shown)


webdeb:
@martijnrusschen if it looks good to you, I'll squash it urgently.

We are facing this bug currently with a combination of a custom input field and this bug is an actual show-stopper.

martijnrusschen:
Thanks for taking a look at this! Can you get the test fixed?

src/date_utils.js
```diff
@@ -149,6 +149,11 @@ export function getYear(date) {
   return get(date, "year");
 }
 
+// Returns a string, which is unique to each year + month combination
+export function getView(date) {
```
rafeememon: Can we get a better name for this method?  Or better yet, change it to something like `areMonths Equal`

aij: This might make more sense to define elsewhere, or just inline. It's not really moment-specific, and wouldn't make sense to redefine for every date library we might want to support.

webdeb: I agree, that its not moment specific, I think its better to place it inside the component itself, because its the only place its used.

I'll fill define a custom function inside index.js to make the componentReceiveProps callback a bit leaner.

Naming is hard, but something like this should be clear enough..

`hasPreSelectionChanged = (date1, date2) => bool`

docs-site/src/examples/inline.jsx
```diff
@@ -32,10 +32,27 @@ export default class Inline extends React.Component {
           </code>
         </pre>
         <div className="column">
+          <div>
+            <input
```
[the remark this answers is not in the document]
webdeb: I used onBlur to update the date, to work on the issue precisely.

webdeb:
@rafeememon the codecov thing is failing because there are no tests for that `hasPreSelectedChanged` function. This was the main reason to put it into utils, so that it could be tested. What do you think, can we merge it anyway?

webdeb:
@aij added a test as suggested, thank you for looking into it

docs-site/src/examples/inline.jsx
```diff
@@ -35,7 +35,8 @@ export default class Inline extends React.Component {
           <DatePicker
             inline
             selected={this.state.startDate}
-            onChange={this.handleChange}/>
+            onChange={this.handleChange}
+          />
```
webdeb: cannot get rid of this changes, because of the prettier precommit-hook :D

webdeb:
@rafeememon can you please remove you "change request" since it refers to an old state?

src/index.jsx
```diff
@@ -31,13 +30,27 @@ import {
...
+  getYear,
+  getMonth
 } from "./date_utils";
 import onClickOutside from "react-onclickoutside";
 
 const outsideClickIgnoreClass = "react-datepicker-ignore-onclickoutside";
 const WrappedCalendar = onClickOutside(Calendar);
 
+// Compares dates year+month combinations
+const hasPreSelectionChanged = (date1, date2) => {
```
rafeememon: I think this should be tightened up a bit:

```
function hasPreSelectionChanged(date1, date2) {
  if (date1 && date2) {
    return getYear(date1) !== getYear(date2) || getMonth(date1) !== getMonth(date2);
  } else {
    return date1 !== date2;
  }
}
```

webdeb: Do you agree, that this is really primarily opinion based?
For me both ways are just fine, no personal preference here.

Lets get a third voice here, @aij / @martijnrusschen ?

rafeememon: Sure it's opinion, but
* Using `function` instead of `const` is _much_ more idiomatic
* The change to the body that I suggested is more direct in what is being compared rather than roundabout

webdeb: Ok, however, its fine for me. I changed it to your suggestion. Hope thats good to go now. Thanks

rafeememon:
@martijnrusschen can you please cut a patch release?

martijnrusschen:
Done!

webdeb:
Awesome, thank you guys!

PR #1356, opened 2018-04-17
Added renderDayContents prop for custom day contents/day formatting

DriesH:
Following prop will allow you to pass in a function that receives the date and will return whatever the user specifies.

It will allow you to do the following:

```js
<Datepicker renderDayContents={day => <div>{day}</div>} />
```

The reason I added this props is because we currently lack the possibility to customise the internals of the day component. It would be nice if we can pass in some extra stuff like components and other text for each day component.


karlkras:
So this functionality was never pulled in and there isn't any support for customizing the Day strings in the calendar header?

DriesH:
I think so yea, but it's been a long time since I last used this component, so I can't say for certain if this functionality was added in newer versions

jRichardeau:
Hi,
Is it planned to merge this PR soon ?
It would be very helpfull to fully customize this calendar !
Thank you

martijnrusschen:
There are some merge conflicts we need to fix before merging.

jRichardeau:
Yep saw it, do you have any idea of a deliver date ?

DriesH:
Fixed the merge conflicts

martijnrusschen:
Thanks, merged!

mikeyyyzhao:
Similar to renderDayContents, is there a prop for customizing time for the time dropdown?

naitschi:
This prop is missing in the datepicker.md. Can someone add it?

PR #1376, opened 2018-05-03
Adding babel polyfill to docs-site for IE compatibility

### LINKED ISSUE #1004: Demo Page Not working in IE 11
I tried to open https://hacker0x01.github.io/react-datepicker/ in IE 11 and the page is unable to display.

jeffredodd:
This PR resolves #1004 by fixing support for IE.  This fix works by introducing `babel-polyfill` as a dev dependency for the docs site. It corrects issues with IE and `Array.includes` and `Array.find`. This change is simply adding a dev dependency and including it in our webpack. This change _ONLY_ affects the doc site and not the datepicker directly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "babel-plugin-react-transform": "^3.0.0",
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.10",
+    "babel-polyfill": "^6.26.0",
     "babel-preset-airbnb": "^2.4.0",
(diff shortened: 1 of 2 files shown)


rafeememon:
Hey @jeffredodd, thanks for the fix.  I'm more than happy to get this PR merged, but a solution I would much prefer would be to remove the uses of `Array.includes` and `Array.find` in the codebase so that the datepicker is compatible with IE for our consumers without a polyfill.  I believe both of those operations can be trivially replaced.

martijnrusschen:
Going to merge this as a band-aid.

Issue #1384, opened 2018-05-16
How to render children at the top of the popup?


whatisaphone:
I see that it's [possible](https://reactdatepicker.com/#example-43) to render extra content inside the popup, but it always appears at the end. Is there any supported way to include content *before* the calendar, something like this?

![](https://i.imgur.com/oOMml7J.png)

PR #1387, opened 2018-05-21
Add calendarContainer prop

johnsoft:
This patch lets you specify a container for the calendar, solving #1384.

Example usage:

```js
<DatePicker
  onChange={this.handleChange}
  calendarContainer={MyFancyBox}
/>
```

Implementation notes:

* I modeled the feature after the existing `popperContainer` prop.
* The approach was stolen from [react-day-picker](http://react-day-picker.js.org/examples/input-custom-overlay/). Their approach is good because it is composable, letting you add elements to the top or bottom, inside or outside the container, and in extreme cases even manipulating or replacing the children.
* This patch obviates the need for supporting `children` as a prop, however I left that feature untouched.
* The original container is exported from the entrypoint as a component so it can be reused.
* It is tested
* I added a docs page example

Example

### DIFF
--- a/docs-site/src/example_components.jsx
+++ b/docs-site/src/example_components.jsx
@@ -39,6 +39,7 @@ import MultiMonth from "./examples/multi_month";
 import MultiMonthDrp from "./examples/multi_month_drp";
 import MultiMonthInline from "./examples/multi_month_inline";
 import Children from "./examples/children";
+import CalendarContainer from "./examples/calendar_container";
 import Portal from "./examples/portal";
 import InlinePortal from "./examples/inline_portal";
 import RawChange from "./examples/raw_change";
@@ -239,6 +240,10 @@ export default class exampleComponents extends React.Component {
     {
       title: "Children",
       component: <Children />
+    },
+    {
+      title: "Calendar container",
+      component: <CalendarContainer />
     },
     {
       title: "Get raw input value on change",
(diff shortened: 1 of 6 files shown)


johnsoft:
The coverage went up. What is codecov complaining about?

aij:
Please don't commit build artifacts (`docs-site/bundle.js`). They will be updated during the release process. Other than that, this looks great. Thanks. 

Codecov is complaining about the diff coverage, but it looks like (mostly?) false positives. [Eg.](https://codecov.io/gh/Hacker0x01/react-datepicker/compare/e052b4ad349c92670b63c0a19d69a6a5f1fb6396...b8d50ef1ec2c6fa34d4a27251c379e4d54415fad/diff#D5-14) I think we can safely ignore it.

johnsoft:
I was wondering why the bundle was not gitignored... I was just trying to go with the flow 😃 

Makes sense. I cleaned up the branch.

aij:
I think it's not gitignored because of the release script... Though, getting the release script to work with .gitignore is probably a better approach in the long run.

Welcome to Contributors! :smiley:

johnsoft:
All right! 🎉 

Do you happen to know when the next release is going to be? We're not sure if we should spend the time to publish our fork temporarily while we wait.

WeijianXu:
Hello，I just noticed this commit. I have the same demand and wait the publish. when is the next release going to be?

Issue #1334, opened 2018-03-27
Demo page rapidly flashes


JacquiManzi:
### Expected behavior
Demo page does not rapidly flash

### Actual behavior
Clicking on the datepicker input area causes the page to rapidly flash.

### Steps to reprodcue
Go to https://reactdatepicker.com/ and click on the date picker input, rapid flashing occurs.

Browser Chrome Version 64.0.3282.186 (Official Build) (64-bit)

martijnrusschen:
Just tested this on the latest Chrome and seems to work fine for me. Can you try this with a different browser and see if it might be related to your Chrome?

martijnrusschen:
I'm using `Version 65.0.3325.181 (Official Build) (64-bit)`

JacquiManzi:
Just had a co-worker re-produce this on Chrome, it didn't happen for him at first, but once he refreshed the page and tried again he got it to flash. I'll try Firefox.

martijnrusschen:
Does it happen on any specific example, or isn't that relevant to reproduce the issue?

mmcc:
^^ I am the above coworker, using Firefox 59.0.1. It seems to only happen on the _first_ click to the example in the hero. If I click elsewhere I can get the flashing to stop, and other examples (including that one) seem to work fine. If I refresh the page again and click the hero example, it triggers it again. Here's a quick recording: http://take.ms/Yj3sj

martijnrusschen:
Hmm, interesting issue. Have you tried this with the latest Chrome as well? I haven't been able to reproduce the issue. I'm not sure if you have time to take a look, but if you can reproduce this on the local build, perhaps we can find out what's going wrong. 

(We (and many others) use this date picker in our production apps and it doesn't seem to be causing any issues for us)

mmcc:
Yep, [also happens in the latest Chrome (65.0.3325.181)](http://take.ms/enEfl). My guess is the problem is just with the demo, not the library itself since it's really only one scenario that causes it. Unfortunately that scenario is also the most prominent example on the demo site, but I don't think this is a mission critical issue with the library itself.

mmcc:
[...]

PR #1390, opened 2018-05-30
Remove second autoFocus example

### LINKED ISSUE #1212: Flickering issue when focusing on input at the top of the docs site
### Expected behavior
Component becomes focused and accepts input

### Actual behavior
Screen flickers and scroll position changes rapidly. See video.

datepicker-bug

### Steps to reproduce
Load demo page https://reactdatepicker.com/ in Chrome (Tested on OSX Chrome 63.0.3239.132)
Before scrolling or clicking anywhere, click inside the input at the top of the page

Issue also exists on Safari

aij:
hero_example already has autoFocus, and having two inputs
focused by default dosn't make sense.

Fixes #1212 #1334


martijnrusschen:
Can you check the failing build?

aij:
I'm seeing the same test failures on master and v1.5.0, but I see they passed on v1.5.0 before. Any idea what may have changed?

Anyway, I'll look into it...

aij:
Oh, it's another transient failure. :disappointed: 

What changed is today's date, so today is now rendered on both the last week of May and the first week of June (before the 1st). I expect it will start passing again next week, but it would be good to make the test more reliable.

aij:
I think 639fddd4aaa0def6a44995f29e82ffacd51cf430 could be reverted now. It was just to scroll back to the top after auto focusing the bottom example, wasn't it?

PR #1392, opened 2018-06-01
Fixed test which fails when today is at the start of the month

mpress:
The test 'should update the preSelection state when a day is selected with mouse click' was failing because today is at the start of the month. Hence today is rendered multiple times and findRenderedDOMComponentWithClass was throwing an error.

By switching to scryRenderedDOMComponentsWithClass and taking the first element the test passes when multiple months are rendered.


aij:
Thanks for the contribution! Is that still a problem? I thought I fixed it yesterday in #1390 .

mpress:
My fix was post-#1390.

PR #1396, opened 2018-06-07
scroll the selected time into view

S1rFrancis:
Scroll the defaultValue or selected value into view when the time picker opens.

### DIFF
--- a/src/time.jsx
+++ b/src/time.jsx
@@ -40,10 +40,12 @@ export default class Time extends React.Component {
 
   componentDidMount() {
     // code to ensure selected time will always be in focus within time window when it first appears
-    var list_elms = this.list.children;
-    for (var i = 0; i < list_elms.length; i++) {
-      if (list_elms[i].innerText === this.props.selected._i) {
-        list_elms[i].scrollIntoView();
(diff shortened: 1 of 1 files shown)


kdrich:
Ran into the same issue. The scrollTop height is calculated at 30px: `this.list.scrollTop = 30 * (multiplier * currH);`, but the element is 31px. I set the height of the element to 30px by overriding the CSS:

```
	.react-datepicker__time-container .react-datepicker__time .react-datepicker__time-box ul.react-datepicker__time-list li.react-datepicker__time-list-item {
		height: 30px;
	}
```

apolzon:
Another artifact of this issue is that the width of the time container is hard-coded, so larger font-sizes are causing each time interval to render onto two lines.

I have set the time-box and time-container elements to auto-width so that each li can render on a single line. 

The fix can then either use scrollIntoView, or ask the browser for the li height when calculating scrollTop as font-sizes, line-height, and padding may be different depending on the user's stylesheet.

As a final tweak, I would probably scroll one element past selected so that the selected item is not at the very top of the list to make it easier for the user to select the previous time interval without scrolling.

PR #1405, opened 2018-06-16
Move '}' for js code markdown.

kaekasui:
It's fine, but I want to fix it.

before the change:
```js
 <DatePicker selected={this.state.date}
  onSelect={this.handleSelect //when day is clicked}
  onChange={this.handleChange //only when value has changed}
 />
```

after the change:
```js
 <DatePicker selected={this.state.date}
  onSelect={this.handleSelect} //when day is clicked
  onChange={this.handleChange} //only when value has changed
 />
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,8 @@ You can use `onSelect` event handler which fires each time some calendar date ha
 
 ```js
 <DatePicker selected={this.state.date}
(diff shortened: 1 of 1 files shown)


kaekasui:
thanks! I'll delete this branch.

PR #1407, opened 2018-06-19
Preserve the preselect date when the user is typing

johnsoft:
This is a bugfix for a weird keyboard interaction.

The problem: A user is scrolling through dates with the arrow keys. Then they accidenatlly press <kbd>Ctrl</kbd>, an innocuous keypress which should not affect the UI state at all. But instead, the preselect date is cleared, and if they try to keep scrolling, all their progress is reset and the datepicker will take them back to today's date.

The solution: Only set the preselect date when the keypress is a navigation key.

There was a slight conflict with #946. To solve this, I tried to keep track of the most recent user intent in React state. If the user is typing, chances are they care most about the input text. Likewise, if they are navigating with arrows keys, the preselect should likely take priority.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -248,7 +248,8 @@ export default class DatePicker extends React.Component {
       preSelection:
         open && this.state.open
           ? this.state.preSelection
-          : this.calcInitialState().preSelection
+          : this.calcInitialState().preSelection,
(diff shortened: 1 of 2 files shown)


martijnrusschen:
Thanks!

PR #1408, opened 2018-06-20
Improving accessibility

aldavigdis:
* Improving keyboard navigation. Now works with time picker too.
* Side-effect of keyboard navigation is now shift key needs to be held down for calendar navigation.
* Shorthand weekday names are now displayed as `abbr` tags, with full names as title.
* Using a colour with higher contrast for month navigation.
* Adding a WAI-ARIA compatible description.
* The text field now has the `aria-live="polite"` attribute, which should update screenreader users as soon as the value changes.
* The date picker popper now has the `aria-hidden` tag, making it invisible to screenreaders and assistive technologies, in order to make navigation more straightforward.
* The aria-describedby tag has been added as well as a description of the keyboard navigation etc.

### DIFF
--- a/docs-site/bundle.js
+++ b/docs-site/bundle.js
@@ -78,11 +78,11 @@
 
       var _react2 = _interopRequireDefault(_react);
 
-      var _reactDom = __webpack_require__(8);
+      var _reactDom = __webpack_require__(7);
 
       var _reactDom2 = _interopRequireDefault(_reactDom);
 
-      var _root = __webpack_require__(16);
+      var _root = __webpack_require__(17);
 
       var _root2 = _interopRequireDefault(_root);
 
@@ -111,7 +111,7 @@
     },
     /* 3 */
     /***/ function(module, exports, __webpack_require__) {
-      /** @license React v16.3.2
+      /** @license React v16.1.0
        * react.production.min.js
        *
        * Copyright (c) 2013-present, Facebook, Inc.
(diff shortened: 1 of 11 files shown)


aldavigdis:
Two things — I see the CI is broken and I will do my best to push the required fix.

Seccondly, this PR is based on a Wordpress Gutenberg ticket. For context, see https://github.com/WordPress/gutenberg/issues/1311

martijnrusschen:
Hi, thanks for making these changes. Seeing the size of the PR, I'll let @aij take a pass on this too.

aldavigdis:
I can split it up into chunks as well.

martijnrusschen:
No worries, just wanted an extra pair of eyes on this PR. No need to split up if it's up to me.

johnsoft:
This comment is about a small part of the PR—the addition of the shift key. A big part of why we chose this datepicker was that picking a date with the arrow keys is convenient and "just works". I wouldn't want to start needing the shift key for this.

I see that the rationale is to preserve cursor movement in the textbox. But on many platforms, shift+left **selects** a character to the left of the cursor, so this change would still be hijacking native behavior. On Windows, ctrl+left moves a word to the left, and you can combine the two, etc. I don't think there's any modifier that wouldn't conflict with standard input behavior on some OS.

In my mind, standard input box behavior isn't really a focus of this component. There are others that do that well (e.g. [react-day-picker](http://react-day-picker.js.org)). It's kind of a UX decision that different datepickers have made in different ways. I also don't think the choice necessarily harms accessibility—as long as dates are being announced to the screen reader it should be clear that the arrow keys are the "primary" interface.

Hope that made sense.

aldavigdis:
@johnsoft What I've been seeing on my side is complaints from users who claim the keyboard navigation is hijacking expected UI behaviour. Freeing direct use of the left and right arrow keys makes it possible to move the input cursor inside the associated input element and edit the value that way, without the user ever having to depend on the date picker.

primetwig:
You could add a prop (e.g. `keyboardNavigationTarget=input/date/time` with `date` as default) for that.

wiscow:
@aldavigdis I would have to agree with @johnsoft on his assessment of the shift key.  I was just about to start working on a PR for a slightly less invasive change which would only allow the left and right arrow key to change the date if the calendar is open.  This would allow a user to navigate the raw input without inadvertently opening up the calendar (my biggest gripe with the current state of accessibility).  The problem with the shift key is that it will have to be explained, as it will not be intuitive to a user.   Bottom line is that if the calendar is open the keyboard navigation on the calendar makes sense.  If it is not, it should be opened by a keystroke and then gain focus.   

Regardless, The improvements you have made here will keep us using this control as they are more restrictive than the current functionality.

josecoelho:
My suggestion would be to split this PR, moving the `shift+arrow` key navigation to another PR but other improvements.

I quite like @wiscow idea of using the arrow keys to navigate only when the date picker is open.

Great contribution @aldavigdis! :)

a2441918:
@aldavigdis Could you fix the conflicting files please?

ben-caplan:
Requiring that the shift key be held while using up/down/left/right arrows to navigate within the calendar dialogue/drop down goes against the W3C A11y specification as seen here: https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html#kbd_label_5

PR #1419, opened 2018-07-06
Hide calendar overlay when readOnly property set to true

### LINKED ISSUE #1418: readOnly property should hide calendar overlay
### Expected behavior
Calendar overlay should be hidden on focus when `readOnly` property is set to `true`. 

### Actual behavior
PopperComponent opens on focus.

screen shot 2018-07-06 at 1 15 56 pm

### Steps to reprodcue
Create Datepicker example with `readOnly={true}`, focus on input.

jbadan:
This PR resolves #1418 by hiding the calendar overlay (PopperComponent) when `readOnly={true}`. An example using the readOnly property has been added to the docs page and an additional test added.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -175,6 +175,7 @@ export default class DatePicker extends React.Component {
       preventOpenOnFocus: false,
       onYearChange() {},
       monthsShown: 1,
+      readOnly: false,
       withPortal: false,
       shouldCloseOnSelect: true,
       showTimeSelect: false,
@@ -270,7 +271,7 @@ export default class DatePicker extends React.Component {
   handleFocus = event => {
     if (!this.state.preventFocus) {
       this.props.onFocus(event);
-      if (!this.props.preventOpenOnFocus) {
+      if (!this.props.preventOpenOnFocus && !this.props.readOnly) {
         this.setOpen(true);
       }
     }
@@ -420,7 +421,7 @@ export default class DatePicker extends React.Component {
   };
 
   onInputClick = () => {
-    if (!this.props.disabled) {
+    if (!this.props.disabled && !this.props.readOnly) {
       this.setOpen(true);
     }
   };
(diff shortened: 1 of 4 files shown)


pniebrzydowski:
This needs to be reverted, I don't understand why it was accepted in the first place. The readOnly property worked perfectly earlier, but now the calendar does not open at all if readOnly is set to true. If you want a true read only date field, then it should just be a standard input field. There needs to be a way to disable user input while still using the datepicker component, and the most logical way to do this is with the readOnly property.

mjlangan:
How does this change make `readOnly` different from `disabled`?

peeyushsrivastava:
Exactly this change actually makes readOnly senseless, most of the chrome updated version not supporting this fix, which is hampering usability of React-Datepicker. This needst to be reverted

src/index.jsx
```diff
@@ -270,7 +271,7 @@ export default class DatePicker extends React.Component {
   handleFocus = event => {
     if (!this.state.preventFocus) {
       this.props.onFocus(event);
-      if (!this.props.preventOpenOnFocus) {
+      if (!this.props.preventOpenOnFocus && !this.props.readOnly) {
```
peeyushsrivastava: This condition will never let calendar once readOnly is set, we can apply this situation for disabled but not for readOnly

Areaxe:
It is recommended to distinguish disabled from readonly.

Kerumen:
We currently have the same issue as @pniebrzydowski. We need the `readOnly` prop to hide the keyboard.

Perhaps it was not the best naming for the API (and it's now the same as `disabled`) but you should name it something like `disableNativeKeyboard` which toggle the `readOnly` attribute on the native input.

This way, we will have a very clear difference and get back the previous behaviour.

Ellis-Dema:
I am also encountering this issue, and cannot set a _proper_ `readOnly` anywhere - on Version 5 - as it messes up my DatePicker no matter what I choose. What can we do? Why is this closed while it's still an issue?

PR #1421, opened 2018-07-09
Adding a label to the month navigation buttons

aldavigdis:
Improves accessibility and facilitates custom localisation by having a default value for the month navigation.

### DIFF
--- a/docs-site/bundle.js
+++ b/docs-site/bundle.js
@@ -111,7 +111,7 @@
     },
     /* 3 */
     /***/ function(module, exports, __webpack_require__) {
-      /** @license React v16.3.2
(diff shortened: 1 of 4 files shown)


martijnrusschen:
Thanks, looks good!

kochis:
Just a heads up, due to the default text on this prop, we started seeing the following after the upgrade to 1.6:
image

Setting `previousMonthButtonLabel` and `nextMonthButtonLabel` to be empty strings will solve it, but wondering if this should be an optional prop.

Huanzhang89:
Yeah I was confused when I saw the labels, came here to find out why XD

PR #1449, opened 2018-08-08
Fix unsafe lifecycle implementation

val1984:
Fix unsafe lifecycle initial implementation changed updating state from props before component is rendered to after component is rendered. This commit uses getDerivedStateFromProps to restore updating state from props back to before render.


val1984:
`yarn test` passes on both Windows and macOS but not Travis-CI, any idea how to get same results locally?

PR #1456, opened 2018-08-20
Fix duplicated defaultProps in Calendar

yhatt:
I found out a duplicated default prop. There are two `previousMonthButtonLabel` definitions in `Calendar` component. This PR will fix one of the default props into `nextMonthButtonLabel`.


martijnrusschen:
Thanks, nice catch!

PR #1483, opened 2018-09-07
Upgrade react-popper

prayerslayer:
There's `1.x` now.

Looks same as on master to me:

image

image

image


martijnrusschen:
Looks good!

prayerslayer:
Thanks a ton for the quick merge :) Would mind releasing a new version also? Then I can remove the dependency on a master commit.

PR #1491, opened 2018-09-18
remove extra div

bertho-zero:


prayerslayer:
I know, but it's there for a reason :grin: (As you can tell by the breaking tests.)

bertho-zero:
There is no valid reason why an extra div should be added, it is a resolvable technical limitation. https://github.com/FezVrasta/react-popper/issues/79

prayerslayer:
Yeah, I wasn't referring to any technical reason. Looking forward to see it fixed :)

bertho-zero:
I guess it's not going to be easy, I'll try. :grin:

martijnrusschen:
How're things going?

bertho-zero:
@prayerslayer @martijnrusschen It works well with the same test using enzyme.

Issue #1489, opened 2018-09-18
Popper not displaying correctly


Gwith:
When using Datepicker with time only (showTimeSelect, showTimeSelectOnly) I'm getting unexpected results:

                <DatePicker
                  selected={this.props.startDate}
                  onChange={this.props.handleChange}
                  withPortal
                  showTimeSelect
                  showTimeSelectOnly
                  timeIntervals={15}
                  dateFormat="LT"
                  timeCaption="Time"
                />

### Expected behavior
[Here](https://i.imgur.com/1C5sk0e.png) is a screenshot of what the Component should look like when rendered to the DOM. 

### Actual behavior
[This](https://i.imgur.com/lydgvp3.png) is the result when used. As you can see the popper is being cutoff some how.

### Steps to reproduce
If it matters I have this component embedded into a table, an example is shown below:

       <table>
      <thead>
        <tr>
          <td>header</td>
        </tr>
      </thead>
      <tbody>
        <tr>
          <td>
            <DatePicker
              selected={this.props.startDate}
              onChange={this.props.handleChange}
              showTimeSelect
              showTimeSelectOnly
              timeIntervals={15}
              dateFormat="LT"
              timeCaption="Time"
            />
          </td>
        </tr>
      </tbody>
    </table>

What might be the issue that is causing the popper window to not display correctly?

JacoGasp:
A workaround is to add to `index.css` the following
```
.react-datepicker__time-container .react-datepicker__time .react-datepicker__time-box ul.react-datepicker__time-list {
    padding: 0px
}
```

PR #1495, opened 2018-09-20
Fix the spacing on the time picker

benlloydjones:
An extra 70px of padding is turning up on the time picker in the latest versions of Firefox, Chrome and Safari, this is forcing hte time select to be clipped and not displayed properly. This is related to issue #1489 https://github.com/Hacker0x01/react-datepicker/issues/1489

### DIFF
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -230,7 +230,8 @@
         margin: 0;
         height: calc(195px + (#{$datepicker__item-size} / 2));
         overflow-y: scroll;
-        padding-right: 30px;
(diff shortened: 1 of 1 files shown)


src/stylesheets/datepicker.scss
```diff
@@ -230,7 +230,8 @@
         margin: 0;
         height: calc(195px + (#{$datepicker__item-size} / 2));
         overflow-y: scroll;
-        padding-right: 30px;
+        padding-right: 0px;
+        padding-left: 0px;
```
karlkras: or simply:
`padding: 0;`  ?

PR #1505, opened 2018-09-26
Scroll the selected time into view

### LINKED ISSUE #1297: Selected time doesn't scroll into view when popper is opened
### Expected behavior
When opening calendar with time list **selected** time should be scrolled into view.

### Actual behavior
Selected time is not visible in the list view port and user has to scroll the list to find it.

### Steps to reprodcue
Having **selected** property set to particular date/time along with **showTimeSelect** property set to true and **timeIntervals** set to 1 (or 2, or 5 etc.) click on DatePicker to open the popper.

idmontie:
Works by taking the rendered height of the first child element
in the time list and using that as the multiplier.

Close #1297

### DIFF
--- a/src/time.jsx
+++ b/src/time.jsx
@@ -44,7 +44,14 @@ export default class Time extends React.Component {
     const currH = this.props.selected
       ? getHour(this.props.selected)
       : getHour(newDate());
-    this.list.scrollTop = 30 * (multiplier * currH);
+
+    let itemHeight = 30;
+
+    if (this.list.children && this.list.children.length > 0) {
+      itemHeight = this.list.children[0].offsetHeight;
+    }
+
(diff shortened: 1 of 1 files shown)


martijnrusschen:
Great work! Can you add a test for this functionality?

idmontie:
@martijnrusschen How would you suggest I go about covering this behavior? I don't see any tests that currently cover the componentDidMount lifecycle.

idmontie:
@martijnrusschen pinging you again in case you didn't see my [previous comment](https://github.com/Hacker0x01/react-datepicker/pull/1505#issuecomment-438444433)

maniator:
I am not sure how this solves it though

Doesn't this still assume that the height of each item is 30px?

PR #1523, opened 2018-10-19
More convenient examples navigation

ingver:
Using `position: sticky` if supported for more always visible navbar


martijnrusschen:
Let's try :)

PR #1524, opened 2018-10-23
Set the height of the list explicitly

umanghome:
The height currently doesn't take up the entire available space until the user hovers on the component. This fixes it.


charlesmarino1:
Wooooo!

PR #1526, opened 2018-10-25
Allow passing popperProps. Fixes #1525

### LINKED ISSUE #1525: It should be possible to pass relevant props to the `Popper`
When updating this in our project we discovered that the popper is added to different positions of the document.
In previous version (we used 0.28.2) it was added closer to the `body`, but now it is added near the reference element.
This caused a regression for us that caused the popper to be hidden (when parent has `overflow: hidden`).
I noticed that [Popper](https://github.com/FezVrasta/react-popper) allows passing a prop called `positionFixed` which addresses exactly such an issue.
The same prop would also help solve #1246 

### Expected behavior
Being able to pass props to Popper

### Actual behavior
Only some props are exposed

### Steps to reproduce
Example: https://codesandbox.io/s/o7mqjjww29
Notice that the popper is hidden because the containing parent has `overflow: hidden`

idanen:
Adds the ability to send props to the `Popper` component

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -1,8 +1,8 @@
-import Calendar from './calendar';
-import React from 'react';
-import PropTypes from 'prop-types';
-import PopperComponent, { popperPlacementPositions } from './popper_component';
-import classnames from 'classnames';
+import Calendar from "./calendar";
+import React from "react";
+import PropTypes from "prop-types";
+import PopperComponent, { popperPlacementPositions } from "./popper_component";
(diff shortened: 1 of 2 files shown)


martijnrusschen:
Nice work!

idanen:
Thanks.
Wow that was quick 😊
What's the plan for publishing a new version to npm?

martijnrusschen:
Will try to get a new version out today

Issue #1349, opened 2018-04-06
Any thoughts on migrating away from moment to a more modularized date library like date-fns?


dchhetri:
moment.js is huge library and isn't modularized. There is a [issue ](https://github.com/moment/moment/issues/2373)created for moment modularization but to be frank I don't think it will be resolved anytime soon. I'm wondering if there are any thoughts on moving away from moment into a modular date library like [date-fns](https://date-fns.org/)?

martijnrusschen:
Good question. We have multiple discussion about moving away from moment to libraries like Luxon, or making moment optional so that you can use any library for your date picker.

Ideally, I think it would be great if the date picker could handle multiple date engines. This allows customers that use the library they prefer (for example, at HackerOne we use moment at many places in the app and that's why we choose this library for the date picker) but someone else could use date-fns and might not want to include moment just for this date picker.

dchhetri:
That would be ideal, do you think we can use contextAPI to help assist that somehow?

nateabele:
Just ran into this. We're trying to get off of Moment as well, but recently introduced a dependency on `react-datepicker`, which is the last transitive dependency on Moment at this point.

> In my opinion, libraries such as this should be as lean as possible. [...]

Seconded. After reviewing the codebase, it doesn't seem like abstracting for different date libraries would be worth the conceptual overhead or the code weight.

@martijnrusschen I'd be happy to submit a PR based on @jrop's 3 bullet points. I _think_ I can even do it in such a way that it's not a BC break for existing Moment + react-datepicker users. Thoughts?

treyhoover:
@nateabele Let me know if I can help. This would remove our last moment.js dependency as well.

nateabele:
@treyhoover I kinda don't even want to start on it unless I know it'd be accepted in principle... `react-datepicker` seems to be the only reasonable date picker out there for React, _but_, there seems to be enough momentum that maybe a fork is worthwhile if not.

nateabele:
@martijnrusschen Ping! Just wanted to check in one more time. Any thoughts?

jrop:
@nateabele I took a whack at creating a PR for this, but quickly abandoned it as I realized I did not have enough `time + project_familiarity` to finish it :frowning_face:

grantglidewell:
Anything new on this? Im working on bringing down the bundle size of my app and moment is a huge chunk of it currently.

martijnrusschen:
[...]

PR #1527, opened 2018-10-25
Moment removal

Darrken:
This PR removes moment as a dependency and uses the native javascript `Date` object paired with date-fns for all date functionality within the datepicker.
Addresses #1349 (among other similar requests)

Benefits:
- much smaller dependency footprint
- no more clone() everywhere; date-fns functions return a new date object
- downstream code doesn't need any specific library to handle dates

Removed:
- moment.js
- `utcOffset` as a prop - if anyone has any suggestions how this might be implemented with the native `Date` object, I'd love to hear ideas. Otherwise, I felt leaving this out was likely an acceptable trade-off for now. (date-fns does have a UTC module in development.)
- Timezone date demo

Added:
- some public helper functions to make up for the localization support moment provided: `registerLocale, setDefaultLocale, getDefaultLocale` (docs updated to reflect these changes)

This is obviously a pretty big PR, so thorough review is needed and welcome.


README.md
```diff
@@ -20,23 +20,22 @@ The package can be installed via NPM:
 npm install react-datepicker --save
 ```
 
-You’ll need to install React, PropTypes, and Moment.js separately since those dependencies aren’t included in the package. Below is a simple example of how to use the Datepicker in a React view. You will also need to require the CSS file from this package (or provide your own). The example below shows how to include the CSS from this package if your build system supports requiring CSS files (Webpack is one that does).
+You’ll need to install React, PropTypes, and date-fns (for localization support if needed) separately since those dependencies aren’t included in the package. Below is a simple example of how to use the Datepicker in a React view. You will also need to require the CSS file from this package (or provide your own). The example below shows how to include the CSS from this package if your build system supports requiring CSS files (Webpack is one that does).
```
aabenoja: I'm slightly confused by the start of this line. We have `date-fns` set as a dependency, but we're telling users to install it like we would a peer dependency?

Darrken: You're right, I was thinking they'd need it in their downstream project but they shouldn't. I'll make that more clear.

jRichardeau:
Do you know when this PR could be available ? I want to use this picker but moment is a too big dependency.
Thank you

Darrken:
@aij Date_utils still serves as an abstraction, at least the same way it did for all the moment functions.  I don't think I could have done this without the previous work to move the moment logic into date_utils! I tried to keep any calls to `new Date()` or date-fns functions behind the date_utils functions so there's a single source. If you have a suggestion on how that could better be separated or abstracted (maybe I'm misunderstanding), let me know.

4ware:
Please merge this into the master if no conflicts exist!!

martijnrusschen:
Thanks, I merged it. Going to release a new major version today to reflect the breaking changes.

martijnrusschen:
V2.0.0 is now live!

PR #1549, opened 2018-11-21
Allow option in timepicker to default to minTime

hamza:
Currently the timepicker will scroll the list to the current time when it is first opened. This change allows you to supply an option (`defaultToMinTime`) that alters this behavior, instead using the value of the `minTime` prop to select a list item to scroll in view.


hamza:
Rebasing and opening a version without conflicts momentarily

PR #1553, opened 2018-11-21
For timepicker - allow defaulting the popup to the min time

hamza:
Currently the timepicker will scroll the list to the current time when it is first opened.

This change allows you to supply an option (`defaultToMinTime`) that alters this behavior, instead using the value of the `minTime` prop to select a list item to scroll in view.

This will only work when `minTime` itself is specified, otherwise there will be no effect.

Let me know if there are any modifications to the approach here that you'd like, happy to iterate!

Note: There are also some format changes from the pre-commit hook.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -78,6 +78,7 @@ export default class Calendar extends React.Component {
     onWeekSelect: PropTypes.func,
     showTimeSelect: PropTypes.bool,
     showTimeSelectOnly: PropTypes.bool,
+    defaultToMinTime: PropTypes.bool,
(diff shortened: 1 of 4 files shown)


sweetroll:
This would be great feature. Is there a timeframe for when this could be merged? Cheers.

hamza:
To anyone monitoring this, I could use some help interpreting the code coverage check here.

The above Codecov report seems to indicate from the deltas that coverage went up on all the affected files. But the "patch" metric is too low somehow? I don't understand.

Would love to remedy this and do what it takes to get this merged. Thanks!

Issue #1102, opened 2017-10-31
onclickoutside issue on IE


tony1Centre:
There is an issue (onclickoutside) in IE11,  and I've tried your page which is broke in IE as well.
https://hacker0x01.github.io/react-datepicker/

I've tried to down the version of the package to 56, 54, not working at all.

Thanks,
Tony

bhldev:
The onclickoutside depends on Edge or other browsers and doesn't support IE https://github.com/Pomax/react-onclickoutside

The only workaround is to not use the onClickOutside handler and use say, blur and disregard the edge cases onClickOutside deals with... or use IE specific events https://stackoverflow.com/questions/3422754/problem-with-onblur-event-in-ie

Good luck

~B

PR #1554, opened 2018-11-26
IE11 fix 

S-t-e-p-s:
Fix for IE issue: #1102
Disable clickOutside behavior if month or year dropdown are shown. And re-enable it when dropdown are hidden.

Description:
In IE if you have date picker rendered in `scroll` view for month and year selectors, when user open drop down and click outside - js error happens.
To fix this: I've introduced two additional methods (props) `beforeShow` and `afterShow` which allow me to get the moment of opening and closing those popups. At that moment I disable / enable outside click library for the main calendar control so when user clicks outside opened year drop down it will close only year drop down. So user can return to calendar view.


martijnrusschen:
You are not going to continue this?

Issue #1563, opened 2018-11-28
Uncaught RangeError: Invalid interval


willbee28:
### Expected behavior
Nothing. I should be able to error handle on my own. 
### Actual behavior
When a start date is chosen after the end date, an error pops up and my application crashes.
### Steps to reproduce
Use the date range version of react-datepicker and choose an endDate that is before the startDate. In the demo on your site, I am not allowed to select an End Date before the Start Date, like physically cannot. Wonder why my result is different.

jRichardeau:
I think the component doesn't do it natively and lets you implement your own behavior.
You have to manage this in the _onChange_ callbacks : for instance block the date change or change the start/end depending of what is the value.
For my part I made the start and end have the same value if the period is invalid and I think in the demo it is the first option that is implemented.

willbee28:
Does it show how that is implemented? Also not to be an asset, but I think
that's kinda dumb, that should be on the user of the library. I like that
there is validation but I think it should be a choice and not in the form
of a application-crashing bug.

jkeam:
Opened a PR against this (https://github.com/Hacker0x01/react-datepicker/pull/1577).  Comments welcome!

willbee28:
Also, this is most likely beyond this ticket, but I submitted another
addressing that if you begin to type 20 for the year or 19 for the 20th
century, the other date updates automatically to 2020 of 2019 for 20th
century. Weird that it doesn't wait for the user to finish typing I think.

sugarfree1:
It is fixed now. Could you release it, please?

PR #1577, opened 2018-12-13
Fix range error

jkeam:
Fix for https://github.com/Hacker0x01/react-datepicker/issues/1563.

Instead of an uncaught exception, the error is handled simply as a `false` when considering if the dates are in range.

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -224,7 +224,13 @@ export function isSameDay(date1, date2) {
 }
 
 export function isDayInRange(day, startDate, endDate) {
-  return isWithinInterval(day, { start: startDate, end: endDate });
+  let valid;
+  try {
+    valid = isWithinInterval(day, { start: startDate, end: endDate });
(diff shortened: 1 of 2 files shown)


jkeam:
Looks like `prettier` clobbered the quotes (single/double).  I see the command run is `prettier --write`, but with no `--single-quote` option.

Looks like `date_utils.js` was last touched Nov 17th.  Prettier, in particular the command `prettier --write` in `package.json` that overwrites the file in place on commit, was added Nov 25.

What is the correct formatting?  Is there a `.prettierrc` that should be checked in or do you want to go with the standard defaults for prettier?  If the latter we are good to go :)

shansmith01:
Is this going to be merged? Just dealing with this issue now

martijnrusschen:
If you can rebase this PR, I'm happy to merge it.

jkeam:
Just rebased!

PR #1583, opened 2018-12-14
Allow React elements as button labels

jennyscript:
This updates the prop type and docs for `previousMonthButtonLabel` and `nextMonthButtonLabel` to allow any renderable content to be passed in. For my use case, I'm passing in some custom svgs for the icons. The code already supports this use, it just needs the updated prop type definitions

### DIFF
--- a/docs-site/bundle.js
+++ b/docs-site/bundle.js
@@ -67,7 +67,7 @@
     /* 0 */
     /***/ function(module, exports, __webpack_require__) {
       __webpack_require__(1);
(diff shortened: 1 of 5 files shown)


jennyscript:
I needed to add this test fix from another PR because there's a test failing on master https://github.com/Hacker0x01/react-datepicker/pull/1577/files#diff-4a52023af2254c0d28ddc8e88036ff61R318

martijnrusschen:
Looks like you have a merge conflict. Can you fix that?

PR #1602, opened 2019-01-16
Update to react 16 createRef API

jeremyk:
I want to preface this by saying that I don't expect this to be merged right away. This is just what I needed to change to get this working for us. Would like to know if you are interested at all in this as a pull request and if so I can do some more work if needed to get it merged.

I am not sure how customInputRef is meant to work as any value but ref seems to result in a console warning.

What I want to do is pass in my own reference to the input so I can have access to the dom node to check its size. The createRef API works great but react-datepicker has not been updated and overwrites the ref I put on the custom input. I can work around this by passing a dummy value to customInputRef but that seems kind of silly and puts a warning in the console.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -80,7 +80,7 @@ export default class DatePicker extends React.Component {
     children: PropTypes.node,
     className: PropTypes.string,
     customInput: PropTypes.element,
-    customInputRef: PropTypes.string,
+    inputRef: PropTypes.any,
     // eslint-disable-next-line react/no-unused-prop-types
     dateFormat: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
     dateFormatCalendar: PropTypes.string,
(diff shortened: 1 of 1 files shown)


jeremyk:
Hmm. I guess I could have used datepicker.input. I may switch to that while I wait to see if you want to upgrade react.

martijnrusschen:
Thanks for bringing this up. I think it would be good to move to the newer APIs as we only support React 16 and higher anyways.

src/index.jsx
```diff
@@ -80,7 +80,7 @@ export default class DatePicker extends React.Component {
     children: PropTypes.node,
     className: PropTypes.string,
     customInput: PropTypes.element,
-    customInputRef: PropTypes.string,
+    inputRef: PropTypes.any,
```
lightbringer1991: I think we can use `inputRef: PropTypes.shape({ current: PropTypes.instanceOf(Element) })`.
IMO `PropTypes.any` seems a bit too much.
https://stackoverflow.com/a/51127130

ericreis:
Any news regarding this?

Issue #1619, opened 2019-01-26
ReferenceError: window is not defined


novenix:
### Expected behavior
it suppose to render ther calendar in server side,  but when i save appear this error

### Actual behavior
Client pings, but there's no entry for page: /_error
ReferenceError: window is not defined
    at getLocaleObject (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\react-datepicker\lib\index.js:6099:3)
    at formatDate (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\react-datepicker\lib\index.js:5979:19)
    at safeDateFormat (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\react-datepicker\lib\index.js:5996:18)
    at DatePicker._this.renderDateInput (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\react-datepicker\lib\index.js:8806:154)
    at DatePicker.render (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\react-datepicker\lib\index.js:8881:14)
    at processChild (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\react-dom\cjs\react-dom-server.node.development.js:2863:18)
    at resolve (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\react-dom\cjs\react-dom-server.node.development.js:2716:5)
    at ReactDOMServerRenderer.render (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\react-dom\cjs\react-dom-server.node.development.js:3100:22)
    at ReactDOMServerRenderer.read (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\react-dom\cjs\react-dom-server.node.development.js:3059:29)
    at renderToString (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\react-dom\cjs\react-dom-server.node.development.js:3526:27)
    at renderPage (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\next\dist\server\render.js:319:26)
    at Function.getInitialProps (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\next\dist\server\document.js:65:25)
    at _callee$ (C:\Users\user\Desktop\proyectos\portfolio-nicolas\node_modules\next\dist\lib\utils.js:86:30)
[...]

PR #1634, opened 2019-02-08
Fix for window undefined error in serverside rendering

Darrken:
Addresses #1619.

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -49,6 +49,10 @@ import isWithinInterval from "date-fns/isWithinInterval";
 import toDate from "date-fns/toDate";
 import parse from "date-fns/parse";
 
+if (typeof window === 'undefined') {
+  global.window = {}
+}
+
 // ** Date Constructors **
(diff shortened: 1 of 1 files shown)


julien1619:
Hi! When do you plan to merge this PR? It would be really useful to fix the SSR case with translations enabled. Thank!

src/date_utils.js
```diff
@@ -49,6 +49,10 @@ import isWithinInterval from "date-fns/isWithinInterval";
 import toDate from "date-fns/toDate";
 import parse from "date-fns/parse";
 
+if (typeof window === 'undefined') {
+  global.window = {}
```
rafeememon: I think it would be much better to avoid modifying the global environment, and instead guard the problematic access to `window` with an undefined check.  Adding a unit test for this would be great as well.

macrozone: yes, adding stuff to global is a no-go for libraries like that. ;-)

Issue #1615, opened 2019-01-24
Update renderDayContents function


hamilton-b:
Can you please adjust the following function from:
`renderDayContents(getDate(this.props.day))`
to
`renderDayContents(getDate(this.props.day),this.props.day)`
It would be useful to have the option to use the actual date to customise the day contents

PR #1637, opened 2019-02-14
fix/render-custom-day

mkosir:
This PR fixes the issue with function property `renderDayContents` which currently receives an argument `day` as a number. The issue is that day number is returned more then once for selected month (e.g. 30th of this and previous month) from which user can't know which day is actually manipulating.
This PR changes function `renderDayContents` to receive Date object as an argument instead of day as number.
[Documentation ](https://github.com/Hacker0x01/react-datepicker/blob/master/docs/index.md) already states correct argument is being passed as `date` but `day` is actually implemented.
PR also adds new example on how to use custom day rendering to [official page](https://reactdatepicker.com/).
There is already opened issue from another user - https://github.com/Hacker0x01/react-datepicker/issues/1615


markokosir:
@martijnrusschen I tried to figure out why codecov/project fails and how to fix it, but unable to do so, not much info is available in the fail report. Can you help me out with this?

src/day.jsx
```diff
@@ -203,7 +203,7 @@ export default class Day extends React.Component {
         role="option"
       >
         {this.props.renderDayContents
-          ? this.props.renderDayContents(getDate(this.props.day))
+          ? this.props.renderDayContents(this.props.day)
```
streletss: Correct me if I'm wrong, but you could actually avoid breaking compatibility by doing 
`this.props.renderDayContents(getDate(this.props.day), this.props.day)`

markokosir: @streletss thanks for your proposition, I modified the code so it's now backward compatible with an older projects.

mkosir:
@martijnrusschen can anybody can take a look at this PR, so we could merged it? It's opened for quite some time now (1 month).

martijnrusschen:
Done!

mkosir:
Thanks.

Issue #1643, opened 2019-02-19
react-datepicker does not  'registerLocale'.


shmulikamin:
I'm trying to do an IMPORT to registerLocale  like this. 

import { registerLocale } from 'react-datepicker';
And that's the result:
'react-datepicker' does not contain an export named 'registerLocale'.

evaldosantos:
I have faced the same problem @shmulikamin. I figure out that it's a named export.

import DatePicker, { registerLocale } from "react-datepicker";

You can check it here: https://github.com/Hacker0x01/react-datepicker/blob/master/docs-site/src/examples/locale.jsx

shmulikamin:
Thanks
But I tried it already, and I still get the same message

evaldosantos:
ah sorry, my bad, I thought you were trying to import it directly like:
import registerLocale  from "react-datepicker";

AbhishekGowda28:
Check the version you are using, if you using previous versions update to the latest one.

I faced the same problem with older versions, updating to newer version fixed the issue.

Rutulpatel7077:
I have same issue. How do I solve it ?

`import DatePicker, { registerLocale } from 'react-datepicker';`

![image](https://user-images.githubusercontent.com/22271244/58727906-a01ab400-83b3-11e9-805b-15c6d04af821.png)

jkepps:
I am also seeing this issue in my production build, but not in my dev environment... my production build is using the minified build and giving me this error. any suggestions?

simonatoluca:
> I am also seeing this issue in my production build, but not in my dev environment... my production build is using the minified build and giving me this error. any suggestions?
> 
> version 2.9.6

Hi guys. I found this solution that works for me

`import DatePicker, { registerLocale } from "react-datepicker/dist/es"`

PR #1646, opened 2019-02-25
(docs) improve locale section

evaldosantos:
Just a small addition to improve the docs.

I started to use react-datepicker these days. I tried to reproduced the code of the locale example, and I didn't have success. Seems like other people faced the same problem (https://github.com/Hacker0x01/react-datepicker/issues/1643). I think this can improve the docs.


martijnrusschen:
Thanks!

PR #1656, opened 2019-03-04
Make date parsing conform strictly to the date format.

### LINKED ISSUE #1655: Date parsing does not conform strictly to the provided date format
### Expected behavior
The date value entered in the input should be parsed strictly based on the provided `dateFormat`. For example, the provided `dateFormat` is `MM/dd/yyyy`. When changing the date value in the input from `03/04/2019` to `03/04/20`, one would expect the date picker to reject `03/04/20` as a valid date value since it does not conform to the date format. However, the date in the calendar is updated to March 4th, 2020 as it accepts `20` as a valid year even when the format has a 4-digit year.

### Actual behavior
The date picker accepts date values that do not conform to the provided `dateFormat`. This causes the calendar to accept a date that does not follow the format and sets the unintended date.

react-datepicker uses the [parse function](https://github.com/Hacker0x01/react-datepicker/blob/master/src/date_utils.js#L71) from date-fns so the cause of the issue is actually in date-fns. See https://github.com/date-fns/date-fns/issues/942. One way for react-datepicker to address this issue is to add a check to format parsedDate using the provided format and compare that with the input value.

`return isValid(parsedDate) && value === format(parsedDate, dateFormat) ? parsedDate : null;
`

### Steps to reproduce
1. Go to the first Default example [here](https://reactdatepicker.com/#example-0).
2. Today's date is dafaulted. (i.e. `03/04/2019`)
3. Delete the last two digit in the year. That is, change from `03/04/2019` to `03/04/20`.
4. Notice that the date in the calendar is updated to March 4th 2020. 
5. Click outside of the input.
6. The date is updated to `03/04/2020`

Example 1:
date-parsing

Example 2: dateFormat is `yyyy/MM/dd`
date-parsing2

benbcai:
Fixes #1655 

The date value entered in the input should be parsed strictly based on the provided `dateFormat`. For example, the provided `dateFormat` is `MM/dd/yyyy`. When changing the date value in the input from `03/04/2019` to `03/04/20`, one would expect the date picker to reject `03/04/20` as a valid date value since it does not conform to the date format. However, the date in the calendar is updated to March 4th, 2020 as it accepts `20` as a valid year even when the format has a 4-digit year.

This change is to make date parse conform strictly to the provided date format.

### DIFF
--- a/docs-site/src/example_components.jsx
+++ b/docs-site/src/example_components.jsx
@@ -51,6 +51,7 @@ import IncludeTimes from "./examples/include_times";
 import InjectTimes from "./examples/inject_times";
 import DontCloseOnSelect from "./examples/dont_close_onSelect";
 import RenderCustomHeader from "./examples/render_custom_header";
+import StrictParsing from "./examples/strict_parsing";
 
 import "react-datepicker/dist/react-datepicker.css";
 import "./style.scss";
@@ -256,6 +257,10 @@ export default class exampleComponents extends React.Component {
     {
       title: "Custom header",
       component: <RenderCustomHeader />
+    },
+    {
+      title: "Strict parsing",
+      component: <StrictParsing />
     }
   ];
 
--- a/docs-site/src/examples/strict_parsing.jsx
+++ b/docs-site/src/examples/strict_parsing.jsx
@@ -1,0 +1,42 @@
+import React from "react";
+import DatePicker from "react-datepicker";
+
+export default class StrictParsing extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      startDate: new Date()
+    };
+  }
+
+  handleChange = date => {
+    this.setState({
+      startDate: date
+    });
+  };
+
+  render() {
+    return (
+      <div className="row">
+        <pre className="column example__code">
+          <code className="jsx">
+            {`
+<DatePicker
+    selected={this.state.startDate}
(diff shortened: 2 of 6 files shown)


benbcai:
@martijnrusschen Can you let me know your thought on this issue and pull request? This breaking change in V2 is a blocker for my team to upgrade to react-datepicker to V2.

benbcai:
@martijnrusschen The CI builds passed. Can you take a look at this? My team is in urgent need of this issue being fixed. Thanks!

rafeememon:
I believe this is a breaking change, so it shouldn't be enabled by default, but maybe with a prop?

benbcai:
@rafeememon react-datepicker V1.8.0 was still doing strictly parsing. This parsing/formatting behavior got changed starting in react-datepicker V2.0.0. This pull request basically changes it back to how it used to work before V2.

If you are concerned that this pull request would break existing react-datepicker V2.0.0 consumers, I'm fine adding a new prop to make this change truly passive.

benbcai:
@rafeememon and @martijnrusschen
A prop named `strictParsing` has been added to this change. You may continue with this pull request review.

benbcai:
@rafeememon, @martijnrusschen Any additional feedback on this PR? This PR is blocking my team's project and would like to get this merged ASAP. Thanks!

martijnrusschen:
Looks good, merging it and will release this later this week.

PR #1660, opened 2019-03-11
Locales without globals

### LINKED ISSUE #1658: Adding a locale modifies the `window` object
We're using react-datepicker in an app where the global context should not be modified. It looks like in v2, non-English locales won't work unless the locales are registered ahead of time on the global `window` object:

https://github.com/Hacker0x01/react-datepicker/blob/ad499a27c895491dcd052bde934fc216df59397d/src/date_utils.js#L244-L249

This does not align with how date-fns itself handles locales. In their [docs](https://date-fns.org/v2.0.0-alpha.27/docs/I18n), they have this example, where the locale object is passed directly as an argument, without going through an indirection layer:

```js
var distanceInWords = require('date-fns/distance_in_words')
// Require Esperanto locale
var eoLocale = require('date-fns/locale/eo')

var result = distanceInWords(
  new Date(2016, 7, 1),
  new Date(2015, 0, 1),
  {locale: eoLocale} // Pass the locale as an option
)
//=> 'pli ol 1 jaro'
```

The react-datepicker equivalent would look something like this:

```js
import eoLocale from 'date-fns/locale/eo';

const MyComponent = () => (
  <DatePicker
    locale={eoLocale}
  />
);
```

Allowing that usage would let you manage locales with all the flexibility that date-fns itself allows. Supporting it in a backwards-compatible way would (at first glance) require only one change, [here](https://github.com/Hacker0x01/react-datepicker/blob/ad499a27c895491dcd052bde934fc216df59397d/src/date_utils.js#L259), to something like this:

```js
export function getLocaleObject(localeSpec) {
  if (typeof localeSpec === 'string')
    // Treat it as a locale name registered by registerLocale
    return window.__localeData__ ? window.__localeData__[localeSpec] : null;
  else
    // Treat it as a raw date-fns locale object
    return localeSpec;
}
```

Can you think of any reason not to support this?

whatisaphone:
This lets you set a datepicker's locale without registering it in a global variable first. Instead, you can now just pass the locale object directly. This matches the way all date-fns functions accept locales. https://date-fns.org/v2.0.0-alpha.27/docs/I18n

It's backwards-compatible, tested, and comes with an example. Closes #1658.

### DIFF
--- a/docs-site/src/example_components.jsx
+++ b/docs-site/src/example_components.jsx
@@ -10,6 +10,7 @@ import CustomDayClassNames from "./examples/custom_day_class_names";
 import PlaceholderText from "./examples/placeholder_text";
 import SpecificDateRange from "./examples/specific_date_range";
 import Locale from "./examples/locale";
+import LocaleWithoutGlobalVariable from "./examples/locale_without_global_variable";
 import ExcludeDates from "./examples/exclude_dates";
 import HighlightDates from "./examples/highlight_dates";
 import HighlightDatesRanges from "./examples/highlight_dates_with_ranges";
@@ -124,6 +125,10 @@ export default class exampleComponents extends React.Component {
     {
       title: "Locale",
       component: <Locale />
+    },
+    {
+      title: "Locale without global variables",
+      component: <LocaleWithoutGlobalVariable />
     },
     {
       title: "Exclude dates",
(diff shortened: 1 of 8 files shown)


whatisaphone:
Thanks for the quick merge!

PR #1683, opened 2019-03-26
fix(minDate,maxDate): Fixing minDate/maxDate to allow one or the other

jgodi:
Fix: Allow the user to specify `minDate` or `maxDate`. 
Use Case: A user wants to allow the input of dates before or after a specific date and not just in-between two dates

If the user specifies `minDate` only then it will make sure the date `isAfter`. 
If the user specifies `maxDate` only then it will make sure the date `isBefore`.
If the user specifies `minDate` and `maxDate` it will make sure the date is within the range.

XR7vBGiFKa


jgodi:
Any update on this one? Anything that you would like me to address or fix?

martijnrusschen:
Looks good, can you fix the merge conflicts so we can merge this?

martijnrusschen:
Or just remove the changes to:
docs-site/bundle.js
docs-site/style.css

jgodi:
On it, thanks!

jgodi:
@martijnrusschen Done!

Issue #1159, opened 2017-12-02
Year dropdown scroll defaults to max value


Bobvantpadje:
I'm using the YearDropdown function on the datepicker but when I click it, the dropdown will always start at the top(max year). Is there a way to set the default value of the dropdown to the current year? 

**How it opens now**
![image](https://user-images.githubusercontent.com/22618775/33517165-7632b8fe-d77f-11e7-87cc-7171a90e7972.png)

**How I want it to open**
![image](https://user-images.githubusercontent.com/22618775/33517184-c0a82acc-d77f-11e7-8e0e-190c61a9dbf9.png)

lmichailian:
@Bobvantpadje You resolved this ?

@Hacker0x01 any hint ?

Bobvantpadje:
@lmichailian Nope.. I'm using another datepicker now

acollier88:
@Bobvantpadje @lmichailian I used maxDate={moment()}

ketysek:
Scroll position of the year picker is really not user friendly now ... 👎

infinitbility:
Added Example of Scrollable Year dropdown

https://infinitbility.com/react-datepicker-scrollable-year-dropdown

bhumi0901:
I'm also having same issue, i have set scrollableYearDropdown={true} but now the issue is it automatically scrolls up to the max value of year, it should scroll where the year is selected.

ardibegolli:
same issue here

javlund:
I had this issue as well. Turns out that it was due to me using `value` instead of `selected` for the value. Hope that helps someone.

PR #1700, opened 2019-04-09
feat(year_dropdown): Add optional ability to scroll to selected value in year_dropdown

bliareg:
yearDropdownScrollToSelected allow to scroll to selected year value inside the year_dropdown via
refs

https://github.com/Hacker0x01/react-datepicker/issues/1159


martijnrusschen:
Can we add some tests to make sure there's decent code coverage?

Issue #1633, opened 2019-02-07
  WebpackError: ReferenceError: window is not defined


ashishlal91:
I am having irritative behaviour with my gatsby build. I can successfully run the development build using `npm start` but when I use to run it in production mode using `gatsby build` I am getting below error

```
error Building static HTML failed for path "/application/"

See our docs page on debugging HTML builds for help https://gatsby.app/debug-html

  6097 | 
  6098 | function getLocaleObject(localeName) {
> 6099 |   return window.__localeData__ ? window.__localeData__[localeName] : null;
       | ^
  6100 | }
  6101 | 
  6102 | function getFormattedWeekdayInLocale(date, formatFunc, locale) {


  WebpackError: ReferenceError: window is not defined
  
  - index.js:6099 getLocaleObject
    [lib]/[react-datepicker]/es/index.js:6099:1
  
  - index.js:5973 formatDate
    [lib]/[react-datepicker]/es/index.js:5973:1
  
  - index.js:5990 safeDateFormat
    [lib]/[react-datepicker]/es/index.js:5990:1
  
  - index.js:8822 DatePicker._this.renderDateInput
    [lib]/[react-datepicker]/es/index.js:8822:1
  
  - index.js:8900 DatePicker.render
    [lib]/[react-datepicker]/es/index.js:8900:1
  
  - bootstrap:19 c
    lib/webpack/bootstrap:19:1
  
  
  - bootstrap:25 a.render
    lib/webpack/bootstrap:25:1
  
  - bootstrap:24 a.read
    lib/webpack/bootstrap:24:1
  
  - bootstrap:36 renderToString
    lib/webpack/bootstrap:36:1
  
  - static-entry.js:194 Module../.cache/static-entry.js.__webpack_exports__.defa    ult
    lib/.cache/static-entry.js:194:18
  
  - bootstrap:24 Promise
    lib/webpack/bootstrap:24:1
  
  
  - gatsby-browser-entry.js:2 Promise._resolveFromExecutor
    lib/.cache/gatsby-browser-entry.js:2:1
  
  - bootstrap:68 new Promise
    lib/webpack/bootstrap:68:1
```

Can someone please help why I am getting this error.
Thank you!!!

abobykin:
Have such error too.

julien1619:
I'm facing the same problem. In fact this library is not compatible with SSR (Server Side Rendering).
[...]

PR #1723, opened 2019-05-01
Fix window undefined issue

CalebKAston:
These changes should prevent the package from crashing sites during Server Side Rendering and other scenarios in which `window` is undefined.

I tried to come up with a way to test the functionality here but I can find no possible way to run a test in your test structure to run a test in which window is undefined. Window is unchangeable.

Addresses Issues: #1619 and #1633.


brenoasm:
Someone could review and/or approve this?

martijnrusschen:
Thanks!

The-Code-Monkey:
can we rebuild the version on npm as this hasn't been pushed to there as far as i can see.

suprraz:
+1 for npm push

PR #1764, opened 2019-06-04
Localization integration with the correct module

osharim:
In the documentation it points to import directly from `import {registerLocale, setDefaultLocale} from DatePicker;` which generates a `Module build failed` error.

This is solved by pointing to the main module `" react-datepicker ";`. I take this opportunity to improve the doc section as well.


martijnrusschen:
Thanks for improving the docs!

PR #1776, opened 2019-06-06
Format time select based on locale

ericreis:
Up to now, the time select did not use locale for formatting, even if using `p+` formatting tokens. This PR aims to fix this bug.

Before:

In the picture below, you can see that even using `timeFormat="p"` and `locale="pt-BR"` the options in the time select are still using the default locale (`en-US`) AM/PM style.

image

After:

Now, it's respecting the format of the locale (`pt-BR`) 24-hour style.

image

### DIFF
--- a/src/time.jsx
+++ b/src/time.jsx
@@ -25,7 +25,11 @@ export default class Time extends React.Component {
     excludeTimes: PropTypes.array,
     monthRef: PropTypes.object,
     timeCaption: PropTypes.string,
-    injectTimes: PropTypes.array
+    injectTimes: PropTypes.array,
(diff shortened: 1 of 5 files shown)


src/time.jsx
```diff
@@ -45,7 +49,7 @@ export default class Time extends React.Component {
 
   state = {
     height: null
-  }
+  };
```
felippenardi:
```suggestion
  };
```

Please review project style

ericreis:
@martijnrusschen Can you take a look at this? I think this fixes an annoying bug regarding internationalization.

martijnrusschen:
Thanks!

Issue #1713, opened 2019-04-17
maxDate not working when change date manually


mxsmirnoff:
### Expected behavior
Change time to maxDate or reset time in input.

### Actual behavior
I can change time manually to whatever I want. 

### Steps to reproduce
Open [example](https://reactdatepicker.com/#example-15) with maxDate in Datepicker site and change time to something greater than maxDate.
![Apr-17-2019 11-10-18](https://user-images.githubusercontent.com/33573380/56271544-6d697480-6101-11e9-8b69-13e119c5417f.gif)

martijnrusschen:
Fixed by #1782

PR #1782, opened 2019-06-11
Fix not working max and min date when change it manually (issue 1713)

mxsmirnoff:
I thinks it logical mistake – change date to the one that was entered if it is out of bounds. And `isDayDisabled` includes check `isOutOfBounds`.
Fix for https://github.com/Hacker0x01/react-datepicker/issues/1713

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -22,7 +22,6 @@ import {
   subWeeks,
   subYears,
   isDayDisabled,
-  isOutOfBounds,
   isDayInRange,
   getEffectiveMinDate,
   getEffectiveMaxDate,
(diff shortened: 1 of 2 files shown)


martijnrusschen:
Can you please add a test to make sure it keeps working?

mxsmirnoff:
Sure.
But I found the case and I don't know if it's ok and if it's not how it should work.
If you change date to "incorrect" it will drop to the last correct. In case of maxDate it works strange because input can have correct intermediate value.
Jun-11-2019 19-32-07

mxsmirnoff:
@martijnrusschen Added test

PR #1784, opened 2019-06-13
Fix/intl date parsing p+ tokens

### LINKED ISSUE #1774: Error on typing when using dateFormat="P"
### Expected behavior

Error screen like shown below:

image

### Actual behavior

Be able to type the date according to the given locale

### Steps to reproduce

Use `dateFormat="P"` and `locale` _props_ and try typing.

ericreis:
This fixes #1774.

This PR updates [date-fns](https://github.com/date-fns/date-fns) and uses its new feature of parsing dates based on locales using `p+` tokens for the format string.

Also, I've included one new step in the parse fallback with this piece of code:

```js
dateFormat = dateFormat
  .match(longFormattingTokensRegExp)
  .map(function(substring) {
    var firstCharacter = substring[0];
    if (firstCharacter === "p" || firstCharacter === "P") {
      var longFormatter = longFormatters[firstCharacter];
      return localeObject
        ? longFormatter(substring, localeObject.formatLong)
        : firstCharacter;
    }
    return substring;
  })
  .join("");

  parsedDate = parse(value, dateFormat.slice(0, value.length), new Date());
```

What this does is before the `new Date(value)` fallback (which only works for `mm/DD/yyyy` dates) we try applying a subset of the format string to the value. If this works we use this result as the `parsedDate` and if not we the `new Date(value)` fallback.


ericreis:
@martijnrusschen Do you mind taking a look at this too? I am really needing this to launch an internationalization feature in my project.

martijnrusschen:
Looks like the build is failing, can you look into that?

ericreis:
@martijnrusschen Apparently, those tests are already failing in the `master` branch.

image

ericreis:
Nevermind, I think I got the problem. Even in the `master` branch, I was using the updated `date-fns` version. Totally my bad.

I will take a look into this.

ericreis:
@martijnrusschen I think now it's all set!

martijnrusschen:
Thanks!

ericreis:
When will a version with this change be deployed?

Issue #1791, opened 2019-06-19
MinDate, MaxDate is not working in Month Range Picker


caothuphu2013:
I have a month range component below:
<DatePicker
            className="react-datepicker-wrapper"
            selected={this.state.startDate}
            startDate={this.state.startDate}
            endDate={this.state.endDate}
            maxDate={this.state.endDate}
            showMonthYearPicker
            onChange={(date) => onChange(date, 'startDate')}
          />
I found that maxDate is not working in only month range mode.
Please to help me resolve this issue.

gautam-pahuja:
Hi @caothuphu2013 It should ideally work, there was another issue like this which was resolved:
https://github.com/Hacker0x01/react-datepicker/issues/1685

Can you please create a JSFiddle of your use case and share it with us?

Thanks

bakaa99:
I can confirm, it is still not working for month picker

gautam-pahuja:
@bakaa99 Can you please share a codepen or JSFiddle of what you are doing.

bakaa99:
I don't know how to include libraries in there. This is component i am using all across my app. It works fine as when using as a datepicker, when i use it for months , the min and max does not work
```
<DatePicker
            id={id}
            autoComplete="off"
            dateFormat={showMonthYearPicker ? "yyyy/MM" : "yyyy/MM/dd"}
            showMonthYearPicker={showMonthYearPicker || false}
            placeholderText={placeholder}
            disabled={disabled}
            className={`${error && 'is-invalid'} form-control border-right-0`}
            openToDate={openToDate}
            defaultValue={value}
            value={value}
            minDate={minDate ? minDate : new Date()}
            maxDate={maxDate ? maxDate : null}
            readOnly={readOnly}
            required={validationRules.isRequired}
            // shouldCloseOnSelect={true}
            onSelect={(value) => handleChange(value, id)}
            onChange={(value) => handleChange(value, id)} />
```

gautam-pahuja:
Can you check this:
https://codesandbox.io/embed/peaceful-water-0kzlf

gautam-pahuja:
Check this: https://codesandbox.io/s/peaceful-water-0kzlf

bakaa99:
i does work when both minDate and maxDate are defined,not when only minDate is defined.I don't want maxDate, only minDate

gautam-pahuja:
@bakaa99 I updated the same sandbox to your use case and that's working as well, can you check?

[...]

PR #1805, opened 2019-06-27
Fix adding range-start and range-end class in month range, MinDate, MaxDate is not working in Month Range Picker

caothuphu2013:
Hi @martijnrusschen ,
Can you review and build up the release version because I have a urgent project?
Thanks a lot.


caothuphu2013:
Resolve issues as https://github.com/Hacker0x01/react-datepicker/issues/1791

PR #1845, opened 2019-08-08
handling safari input bug

pmendozav:
Input time for Safari browser crashes because `input` tag with `type = "time"` is not compatible with it browser. This branch check if browser is Safari and replace `input` tag with the `react-simple-timefield`.

### DIFF
--- a/src/inputTime.jsx
+++ b/src/inputTime.jsx
@@ -38,25 +38,26 @@ export default class inputTime extends React.Component {
     const { time } = this.state;
     const { timeString } = this.props;
 
-    let input = isSafari ? (
-      <TimeField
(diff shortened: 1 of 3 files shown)


src/inputTime.jsx
```diff
@@ -27,25 +37,47 @@ export default class inputTime extends React.Component {
...
+        className="react-datepicker-time__input"
+        placeholder="Time"
+        name="time-input"
+        required
+        value={time}
+        onChange={ev => this.onTimeChange(ev.target.value || timeString)}
+      />
+    );
+
+    // let input =
```
martijnrusschen: If this isn't needed, can you remove this code?

PR #1849, opened 2019-08-14
Added a specific css class to excluded dates

Sustenance:
### Summary
Prior to this change, dates that were passed in via the excludedDates array prop used only had the same CSS class as all other disabled days, `react-datepicker__day--disabled`. 
This made it difficult to target only excluded days with custom styles. I've added an extra CSS class to these days, `react-datepicker__day--excluded`, which makes custom styling easier.

### BTW
 - One of the existing tests was already failing on a fresh checkout of `master`, `DatePicker should not manual select date if after maxDate`.
 - I didn't commit them, but `/docs-site/bundle.js` and `/docs-site/style.css` were changed a lot by the build script, I assume by `prettier`. If you want me to go ahead and commit those as well, just let me know (I'm assuming you update the docs site as part of release).

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -360,6 +360,14 @@ export function isDayDisabled(
   );
 }
 
+export function isDayExcluded(day, { excludeDates } = {}) {
+  return (
+    (excludeDates &&
+      excludeDates.some(excludeDate => isSameDay(day, excludeDate))) ||
+    false
+  );
+}
+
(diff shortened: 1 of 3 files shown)


martijnrusschen:
Looks solid, if you could add an example to the docs site in a follow that would be appreciated.

Sustenance:
Will do, thanks!

PR #1855, opened 2019-08-21
feat: updated dependencies

mxschmitt:
I will fire up one single PR for bumping most of the dependencies, instead of using Dependabot for all of them.

TODO:
- [x] `node-resolve: setting options.jsnext is deprecated, please override options.mainFields instead`
- [x] reduce bundle size of the docs
- [x] improve handling of the babel configurations


martijnrusschen:
Sweet, I just merged 2 PRs that seem to conflict. If you can update the PR, we can get these upgraded all in one go.

martijnrusschen:
>  Error: Cannot find module "react-test-renderer/package.json"

martijnrusschen:
Looks like some test are not executed somehow:
`Firefox 63.0.0 (Linux 0.0.0): Executed 242 of 406 SUCCESS (7.251 secs / 3.886 secs)`

I assume this is causing the drop in code coverage.

martijnrusschen:
Why the build is succeeding while only executing half of the tests is a mystery to me.

mxschmitt:
I also have no clue why not all tests were executed hmm

mxschmitt:
@martijnrusschen fixed 🚀 
Now the rough part is done, only everything now needs to be tested 😄

.babelrc.js
```diff
@@ -3,43 +3,28 @@ var MODULES = process.env.MODULES;
...
-        forceAllTransforms: NODE_ENV === "production"
-      }
-    ],
-    "stage-0",
-    "react"
-  ],
-  plugins: []
-};
+const presets = [
+  "airbnb",
```
mxschmitt: not sure if we need airbnb or present env or both there

test/exclude_time_period_test.js
```diff
@@ -7,7 +7,7 @@ describe("DatePicker", () => {
   let sandbox;
 
   beforeEach(() => {
-    sandbox = sinon.sandbox.create();
```
mxschmitt: was changed, because the old was was deprecated

.travis.yml
```diff
@@ -1,7 +1,9 @@
 language: node_js
 node_js:
   - "node"
-cache: yarn
+cache:
+  - npm
```
martijnrusschen: Do we use npm cache here?

.babelrc.js
martijnrusschen: I don't know anymore, but if it works I'm fine with this.

martijnrusschen: Maybe remove the commented out code?

.travis.yml
mxschmitt: that will add node_modules to the cache. (but since july 2019 it's default active)

mxschmitt:
> Do you feel comfortable merging as is?

not sure I've got one issue left, reduce bundle size of the docs js bundle.

mxschmitt:
I guess we are ready to merge. But I have not tested yet the different output files which will distributed over NPM.

martijnrusschen:
It seems the docs site is down since merging this PR - https://reactdatepicker.com/

mxschmitt:
See #1860 for the fix.

martijnrusschen:
Cool!

PR #1859, opened 2019-08-22
Fixing markdown

eneo5541:
Markdown was broken before due to the codeblock. Still looks pretty bad, but at least it is readable now.


martijnrusschen:
Problem is that the markdown gets generated by https://github.com/reactjs/react-docgen so changing it now, won't fix it when it gets regenerated.

Issue #1773, opened 2019-06-06
Quarter Picker?


lamarant:
Is there a way to select a Quarter of a year? 

If not, this would be a great addition. EG: User can select 2019 Q2, the actual date chosen for each quarter by default would be:

Q1 = Jan 1
Q2 = April 1
Q3 = July 1
Q4 = October 1

And...for financial reporting, the starting date for the year could also be customizable because a "fiscal year" could start on any date.

gautam-pahuja:
That's a very nice idea, actually. We can make it look like this:
http://jsfiddle.net/4mwk0d5L/1/

I can start working on it but would like to wait for other ideas people can suggest or other use cases.

lamarant:
Love it. If the text that appears in the popup could be customizable as well that would be great. In the link below I just changed the "monthsShort" property to display "Q1, Q2, Q3, Q4" instead of the month names.

http://jsfiddle.net/e0sa5hdc/

Thanks!

johndpope:
I need this

PR #1877, opened 2019-08-27
Quarter year picker

rogier2012:
Similar to the month year picker, I have added this variation to choose quarters within a year. The quarter translates into a date of the first day of the quarter.

This solves issue #1773


martijnrusschen:
Looking good, will release later this week.

Issue #1029, opened 2017-09-19
Add prop to show/hide triangle


somewhatabstract:
We don't always want a triangle attached to the date picker. An option to show or hide it would be fantastic. Right now, this has to be done with CSS and relies on knowing the component's internal class names.

eliucs:
Any update on this issue?

somewhatabstract:
This is a feature request; what sort of activity does this bot want? Perhaps I can work on a PR today :)

martijnrusschen:
Sorry, trying to clean up some of the older issues. If you can put up a PR, I'm happy to review it!

somewhatabstract:
PR is up! :smile: Thanks for the nudge.

somewhatabstract:
This is closed by #1892.

PR #1892, opened 2019-08-30
Add showPopperArrow prop

somewhatabstract:
Addresses #1029 

This adds a new prop called `showPopperArrow` and corresponding example to the docs.

The property defaults to `true` to ensure backwards compatibility. Setting it to false results in the popper anchor arrow not being displayed.

image

### DIFF
--- a/docs-site/src/components/Examples/index.js
+++ b/docs-site/src/components/Examples/index.js
@@ -5,6 +5,7 @@ import slugify from "slugify";
 import CodeExampleComponent from "../Example";
 
 import Default from "../../examples/default";
+import NoAnchorArrow from "../../examples/noAnchorArrow";
 import ShowTime from "../../examples/showTime";
 import ShowTimeOnly from "../../examples/showTimeOnly";
 import ExcludeTimes from "../../examples/excludeTimes";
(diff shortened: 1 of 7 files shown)


martijnrusschen:
Awesome, can you perhaps add a few tests to make sure the behavior doesn't break in the future?

somewhatabstract:
> Awesome, can you perhaps add a few tests to make sure the behavior doesn't break in the future?

Sure. I wasn't quite sure what tests to add. The ideal thing would be a snapshot test, but I don't think y'all support that right now. Do you have a suggestion? In the meantime, I'll poke at something and see where I get.

somewhatabstract:
OK, I've added a couple of test cases. Let me know if these are insufficient and what additional tests you'd like to see. 😄

martijnrusschen:
Yep, this is sufficient. Can you resolve the merge conflict, and I'll merge it when it's ready.

somewhatabstract:
OK @martijnrusschen, I have rebased this PR onto the latest `master` and addressed the merge conflict. I think this is all set. Please let me know if there's anything else I can do to help, with this PR or anything else :)

PR #1906, opened 2019-09-09
Add missing external to rollup config

CodeWitchBella:
This adds date-fns/_lib/format/longFormatters to externals array. Otherwise the built file would contain this dependency and at would fail if bundled using rollup because of `Object.defineProperty(t, { __esModule: true })` call where `t` would be `undefined`


CodeWitchBella:
I posted alternative solution here: #1907

martijnrusschen:
Do you have a preferred way of fixing this?

martijnrusschen:
Merged #1907

PR #1907, opened 2019-09-09
Use module version of longFormatters

CodeWitchBella:
This is alternative to #1906

Fixes same problem but with different approach - this keeps bundling longFormatters but avoids the `__esModule` problem


dbatistat:
When this pull request will be merge? I have a mistake in converting to Spanish and I think this corrects the problem.

dbatistat:
Thanks!

PR #1919, opened 2019-09-20
ISSUE-1918: Add the possibility to localise the prefix of the aria-la…

martLeroux:
…bel of day

### DIFF
--- a/docs-site/src/examples/localeWithoutGlobalVariable.js
+++ b/docs-site/src/examples/localeWithoutGlobalVariable.js
@@ -5,6 +5,7 @@
       selected={startDate}
       onChange={date => setStartDate(date)}
       locale={fi}
+      ariaDayPrefix="Päivä"
     />
   );
 };
--- a/docs-site/src/examples/localeWithoutGlobalVariable.js
+++ b/docs-site/src/examples/localeWithoutGlobalVariable.js
@@ -5,7 +5,7 @@
       selected={startDate}
       onChange={date => setStartDate(date)}
       locale={fi}
-      ariaDayPrefix="Päivä"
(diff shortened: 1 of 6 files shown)


docs-site/src/examples/localeWithoutGlobalVariable.js
```diff
@@ -5,6 +5,7 @@
       selected={startDate}
       onChange={date => setStartDate(date)}
       locale={fi}
+      ariaDayPrefix="Päivä"
```
martijnrusschen: Is there a way to derive this value from the locale itself? I feel hardcoding the prefix in the props isn't the right approach, but I'm not sure what the locale file contains exactly.

martLeroux: The word "Day" is unfortunately not part of the locale.
The other option would be to use the dateFormat props to format the day in the aria-label with the appropriate locale. It would make by default a aria-label like : "21/09/2019". What do you think?

martLeroux: Or a better solution would be to have a ariaLabelDateFormat props.
By default it could be : "EEEE, MMMM d, yyyy" so that the reader read : Tuesday, August 27, 2019.

That's what React-dates does. With more complexity. The reader also reads if the date is available or not.

martijnrusschen: Sounds like a plan! Let's go with that?

martLeroux: Done! But i have no idea why the tests are failling \_(ツ)_/¯

PR #1990, opened 2019-12-20
Render custom header for month/quarter pickers.

bartverbruggen:
When the renderCustomHeader prop is present, always use the custom header. All the order times, pick the right header based on month/quarter picker properties.

Also add year functions to the custom header render props.


bartverbruggen:
Thanks!

@martijnrusschen is there a certain policy for version updates? Is it possible to bump it to a newer version after this merge?

Issue #2031, opened 2020-01-27
Make use of "excludeScrollbar" parameter from react-onclickoutside


dan-diaz:
**Is your feature request related to a problem? Please describe.**
When clicking on the scrollbar to scroll a page, `onClickOutside` is triggered and closes the datepicker.

**Describe the solution you'd like**
`react-onclickoutside` offers the `excludeScrollbar` parameter to ignore scrollbar clicks.
I'm working on a Pull Request right now to enable this parameter and set as true by default.

**Describe alternatives you've considered**
none

**Additional context**
This parameter is only in effect when the user's screen has scrollbars constantly visible.
If your scrollbars hide automatically, a click in their area does not count as a scrollbar click and will still trigger `onClickOutside`. And this is because of how `react-onclickoutside` compares the click coordinates to the documentElement. And documentElement clientWidth/clientHeight are different depending on your default scrollbar settings.

PR #2032, opened 2020-01-27
excludeScrollbar prop for react-onclickoutside

dan-diaz:
#2031 
By default, react-onclickoutside will not register clicks on the scrollbar if the users scrollbars are constantly visible (an OS setting).
The property is available to disable but it makes sense to enable this by default.

I wanted to create a test case for this, but testing onClickOutside on a scrollbar in a headless browser is really REALLY difficult.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -103,7 +103,8 @@ export default class DatePicker extends React.Component {
         return date;
       },
       inlineFocusSelectedMonth: false,
-      showPopperArrow: true
+      showPopperArrow: true,
+      excludeScrollbar: true
     };
   }
 
@@ -222,7 +223,8 @@ export default class DatePicker extends React.Component {
     inlineFocusSelectedMonth: PropTypes.bool,
     onDayMouseEnter: PropTypes.func,
     onMonthMouseLeave: PropTypes.func,
-    showPopperArrow: PropTypes.bool
+    showPopperArrow: PropTypes.bool,
+    excludeScrollbar: PropTypes.bool
   };
 
   constructor(props) {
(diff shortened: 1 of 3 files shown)


src/index.jsx
```diff
@@ -103,7 +103,8 @@ export default class DatePicker extends React.Component {
         return date;
       },
       inlineFocusSelectedMonth: false,
-      showPopperArrow: true
+      showPopperArrow: true,
+      excludeScrollbar: true
```
ben-caplan: @dan-diaz do we want this to default to `true`? What was the existing behavior? It probably makes sense to have the new behavior be opt in instead of opt out. Thoughts?

dan-diaz: I can go either way on this

martijnrusschen:
2 small merge conflicts, can you get those resolved and I'll merge it.

ben-caplan:
Thanks for your review @martijnrusschen we will fix these merge conflicts asap! @dan-diaz if you add me to your repo I am happy to make the changes as well.

ben-caplan:
Thanks @dan-diaz! 

@martijnrusschen this is super weird! It looks like the same testing concerns from the prior PR are now being surfaced here? This PR does not really impact code coverage, the update is tiny. I am guessing this is some strange bug in the `codecov` tool between the last PR and this one. 

Like we were noting on the pervious PR Dan has updated the tests to remove some redundant tests and ones that are no longer applicable after the changes. He also added some new ones to cover the enhanced functionality. 

Can you advise on how you would like us to proceed here? I would suggest (if you are open to it) that we merge this PR (as it does not effect code coverage -- see diff) and then we could create a new PR to extend any coverage you feel is lacking? Thoughts?

PR #2034, opened 2020-01-28
Custom time input element with customTimeInput prop

info-bit:
Added a `customTimeInput` prop of PropTypes.element type to render a custom input in place of the default one used in the inputTime component. This enable the user to implement a custom time input control with low effort.

We already use this edited version of react-datepicker in production because our end users are used to work with a smarter time input we developed. We also have to support Safari which doesn't work with stardard date time inputs.

This doesn't affect other behaviours and should be easy to document!

Thanks!


martijnrusschen:
Thanks for fixing!

PR #2047, opened 2020-02-10
fix: docs generation on build

mxschmitt:
Seems to be, that they have changed some templating. That's why e.g. the heading is different now.


martijnrusschen:
Thanks! This is helpful!

PR #2063, opened 2020-02-21
Dynamic aria labels fix

oscarcarlstrom:
Props for controlling aria-labels/aria-label prefixes were missing in the parent components


martijnrusschen:
Thanks, I'll try to get this released later today.

oscarcarlstrom:
@martijnrusschen That would be awesome, thank you!

PR #1541, opened 2018-11-15
Solution to issue #1337

CalebKAston:
Added a prop and logic for keeping the currently displayed months the same when working with a multi-month inline calendar.

See issue: [https://github.com/Hacker0x01/react-datepicker/issues/1337](https://github.com/Hacker0x01/react-datepicker/issues/1337).

This specifically applies to multi-month inline datepickers. Originally, when a date was selected in a month other than the first month, when the months were re-rendered, the calendar "jumped" to make the selected month the first month rendered. I added logic that keeps track of which rendered month the date was selected in and then the code uses that month to render the months in the same order they were prior to a date being selected. This stops the "jumping" that was happening.

This new functionality is the default behavior. Though, should someone desire that the jumping be maintained, everything I added can be negated by setting the new prop `inlineFocusSelectedMonth` to `true`.

I did do thorough QA with these changes throughout all of the examples and it appears that I was successful in only altering what I intended to alter.

Specific changes:
- Added a prop to toggle the new functionality (defaults to using the new functionality)
- Added internal logic to track which month the date was selected in when using an inline datepicker
- Used that additional information when rendering months to stop the "jumping" from happening

PR #2128, opened 2020-05-03
Replace `inlineFocusSelectedMonth` with `focusSelectedMonth`

dizel3d:
`focusSelectedMonth` works like `inlineFocusSelectedMonth`
but doesn't depend on `inline`, nor `monthsShown`.

It  solves https://github.com/Hacker0x01/react-datepicker/pull/1541#issuecomment-569304221


martijnrusschen:
This is a breaking change, will need to do a major release to get this merged in.

dizel3d:
> This is a breaking change, will need to do a major release to get this merged in.

Yes. When are you going to merge it in?
I able to save backward compatibility if it's needed.

PR #2135, opened 2020-05-07
Fixes issues when using renderCustomHeader with alt pickers

### LINKED ISSUE #2134: Cannot read property 'monthDate' of undefined
**Describe the bug**
Exception `Cannot read property 'monthDate' of undefined` is thrown when using props `renderCustomHeader` and `showYearPicker` together. The exception is thrown when trying to destructure `headerArgs` in `renderCustomHeader`: https://github.com/Hacker0x01/react-datepicker/blob/7294cf965f6dde0fa55299c178882dec0f57f7d5/src/calendar.jsx#L625

There are two calls to `renderHeader` (which in turn calls `renderCustomHeader`). One in the `renderMonths` function (which is fine):
https://github.com/Hacker0x01/react-datepicker/blob/7294cf965f6dde0fa55299c178882dec0f57f7d5/src/calendar.jsx#L720

and one in the `renderYears` function:
https://github.com/Hacker0x01/react-datepicker/blob/7294cf965f6dde0fa55299c178882dec0f57f7d5/src/calendar.jsx#L778

I believe the `renderCustomHeader` function needs to handle no `headerArgs` being passed to it and to only render the day names if we aren't showing the yearPicker.

**To Reproduce**
Steps to reproduce the behavior: [codesandbox](https://codesandbox.io/s/upbeat-cache-q65xh?file=/src/App.js)

**Version**
`v2.15.0`

### LINKED ISSUE #2136: Day names are present when using alt pickers
**Describe the bug**
When using `renderCustomHeader` with `showMonthYearPicker`, `showYearPicker` or `showQuarterYearPicker`, day names are rendered even though the calendar isn't showing dates.

**To Reproduce**

Screenshot 2020-05-07 at 18 16 07

**Expected behavior**
Day names should not be rendered when using the above props.

**Additional context**
`v2.15.0`

mercyaj:
Fixes #2134 
Fixes #2136

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -649,6 +649,11 @@ export default class Calendar extends React.Component {
       this.props
     );
 
+    const showDayNames =
+      !this.props.showMonthYearPicker &&
+      !this.props.showQuarterYearPicker &&
+      !this.props.showYearPicker;
+
     return (
       <div
         className="react-datepicker__header react-datepicker__header--custom"
@@ -667,7 +672,7 @@ export default class Calendar extends React.Component {
           prevYearButtonDisabled,
           nextYearButtonDisabled
         })}
-        {monthDate !== undefined && (
+        {showDayNames && (
           <div className="react-datepicker__day-names">
             {this.header(monthDate)}
           </div>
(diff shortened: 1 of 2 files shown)


martijnrusschen:
Thanks for this!

Issue #2150, opened 2020-05-15
you're setting z-index, but it is dangerously low


codefist:
https://github.com/Hacker0x01/react-datepicker/blob/c691d6caf1fa744d30789c60c23d3a7a3ac1cb01/src/stylesheets/datepicker.scss#L45

since it is ALWAYS the intention of the user to be able to see the picker box you should set this value to the css maximum of 2147483647 instead.

Setting it to 1 is too low for the default material ui (react), any i'm sure many others will have the same problem.

Issue #2126, opened 2020-05-01
Z-Index issue


astritspanca:
I am having a serious issue with z-index. Date picker always gets under other divs and it is impossible to change that. I have tried this solution but without success.
```

.react-datepicker-popper {
    z-index: 3333;
 }
```

codler:
try 
```
div.react-datepicker-popper {
  z-index: 1000;
}

PR #2151, opened 2020-05-15
Set .react-datepicker-popper to maximum z-index

codefist:
There's no real use cases where the date picker should be underneath any other element, so let's just set it to the maximum value to squash any problems.

### DIFF
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -42,7 +42,7 @@
 }
 
 .react-datepicker-popper {
-  z-index: 1;
+  z-index: 2147483647;
 
   &[data-placement^="bottom"] {
     margin-top: $datepicker__triangle-size + 2px;


codefist:
https://github.com/Hacker0x01/react-datepicker/issues/2150

codefist:
https://github.com/Hacker0x01/react-datepicker/issues/2126

codefist:
#2150

FezVrasta:
That’s not how z-index works, it’s enough to have the element wrapped in a different stack context to break this workaround

mjchang:
Using portals fixes the problem you're trying to solve.

src/stylesheets/datepicker.scss
```diff
@@ -42,7 +42,7 @@
 }
 
 .react-datepicker-popper {
-  z-index: 1;
+  z-index: 2147483647;
```
jwallet, line 45: that's not how it works
I think what you want is use the prop `popperContainer` and and a component that implements `ReactDOM.createPortal` ou `<Portal />`

martijnrusschen:
Closing this as there are different ways to fix this https://github.com/Hacker0x01/react-datepicker/pull/2151#discussion_r442169067

Issue #2083, opened 2020-03-11
ui is sometimes mixed up with other components..


littlehome-eugene:
**Describe the bug**

I'm using it with react-native-web and when the ui gets drawn under the input, it looks like 

<img width="266" alt="Screen Shot 2020-03-11 at 11 28 24 PM" src="https://user-images.githubusercontent.com/24217396/76427735-08410080-63f0-11ea-8a1b-2c42e362d8e7.png">

whereas it looks fine when the ui gets drawn above the input like below

<img width="260" alt="Screen Shot 2020-03-11 at 11 28 37 PM" src="https://user-images.githubusercontent.com/24217396/76427841-37577200-63f0-11ea-8984-00f0b18c3060.png">

**To Reproduce**
```

 import DatePicker from 'react-datepicker'
 import "react-datepicker/dist/react-datepicker.css";

           <DatePicker
             selected={Date.parse(values.available_to)}
             dateFormat="yyyy-MM-dd"
             onChange={val => {
               setFieldValue('available_to', val);
             }}
           />
```
**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

- OS: mac
- Browser chrome 
- Version ^2.13.0"

**Additional context**

JofBigHealth:
@littlehome-eugene The issue is that react-native-web sets `z-index: 0` on all views in order make a new stacking context for each component. This means that react-datepicker's popup (z-index: 1) will render over the components in that view but under all subsequent components.

You have a variety of options for fixing this. For us, we use an override in our index.html file along with setting a z-index for a custom class option for the date picker:

```
<style>
    * {
      // WARNING! This breaks the stacking in RNW!!!
      z-index: auto !important;
    }
</style>
```

This is a somewhat nuclear option, but it works ok for us because z-index is banned in our codebase and we don't use any other third party libraries (at least on web).

I longer-term solution will be to use react-datepicker in inline mode and render it in a custom portal which can handle click coordinates, scroll and so-on to keep the menu in place. There's quite a few third party libraries that do that.

PR #2152, opened 2020-05-19
Render popper in a Portal by Id to avoid overflow of parent

flakronkadriu:
Using createPortal from ReactDOM the popper is added in a portal and displayed outside of Root Element in order to avoid overflow

Docs added

Unit tests are not added because when unit testing the portal is rendered inside by default

### DIFF
--- a/src/popper_component.jsx
+++ b/src/popper_component.jsx
@@ -3,6 +3,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import { Manager, Reference, Popper, placements } from "react-popper";
 import TabLoop from "./tab_loop";
+import Portal from "./portal";
 
 export const popperPlacementPositions = placements;
 
@@ -33,7 +34,8 @@ export default class PopperComponent extends React.Component {
     popperProps: PropTypes.object,
     targetComponent: PropTypes.element,
     enableTabLoop: PropTypes.bool,
-    popperOnKeyDown: PropTypes.func
+    popperOnKeyDown: PropTypes.func,
+    portalId: PropTypes.string
   };
 
   render() {
@@ -47,7 +49,8 @@ export default class PopperComponent extends React.Component {
       popperProps,
       targetComponent,
       enableTabLoop,
-      popperOnKeyDown
+      popperOnKeyDown,
+      portalId
     } = this.props;
 
     let popper;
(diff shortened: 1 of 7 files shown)


src/popper_component.jsx
```diff
@@ -80,6 +83,10 @@ export default class PopperComponent extends React.Component {
       popper = React.createElement(this.props.popperContainer, {}, popper);
     }
 
+    if (portalId && !hidePopper) {
+      popper = <Portal portalId={portalId}>{popper}</Portal>;
```
akondratsky, line 87: It would be nice to avoid calling extra "React.createElement" if portalId exists. I believe the best way to move this logic together with creating popper container into `if (!hidePopper) {` block, just for more consistency.

mjchang:
FYI his can be achieved with popperContainer already
```
popperContainer={({ children }) =>createPortal(children, document.body)}
```

If you wanted a more direct prop, I think it will be better to add a DOM node instead of an ID to allow more flexibility.

akondratsky:
@mjchang, thanks a lot, it's great note.

I've tried to find way to implement portals (to avoid "overflow:hidden" zone) too, and in this situation I was confused by withPortal attribute. Our team thought that better (and more universal way) to add type check for withPortal prop to pass DOM node (or query string). We started to think about contribution.

But it even more different in light of information of popperContainer. Don't you think it's good idea to add little more documentation since with **flakronkadriu**  we are not alone who bumped into the "issue"? I believe for large part of end users even cleanest self-documented code is less preferable than full and consistent documentation.

mjchang:
Oh absolutely. It's definitely not obvious with the current documentation. I bumped into this when I was implementing this library as well. I only saw this solution when combing through the issues.

As a side note, when trying to find that issue to link I came across this PR: https://github.com/Hacker0x01/react-datepicker/pull/1074. This makes it seem that popperContainer was created for this exact use case.

flakronkadriu:
> FYI his can be achieved with popperContainer already
> 
> ```
> popperContainer={({ children }) =>createPortal(children, document.body)}
> ```
> 
> If you wanted a more direct prop, I think it will be better to add a DOM node instead of an ID to allow more flexibility.

I tried but we have implemented a custom datepicker and when that datepicker is added in a portal it is not rendered correctly because react when mounting the component in a different location for some reason it doesn't render it correctly and for that reason i though it will be good idea to create a new implementation that will render the portal it inside the dom but in the correct position and i achieved this in this implementation but withPortal Prop it just renders it in a modal so i wasn't sure if i should replace that prop with my current implementation, for sure it may not be portalId i was just assuming that some users may want to add it in a different location and yes of course it might be a DOM Node as you preffer.

flakronkadriu:
This PR closes #2083 and related overflow issues

rmlevangelio:
Is this production ready?

Issue #2141, opened 2020-05-12
Accessibility: Dates unreachable via keyboard when `inline` set to `true`


miina:
**Describe the bug**
When the calendar is displayed inline (`inline` set to `true`), the calendar dates seem to be unreachable via keyboard. E.g. tabbing to the date and using arrow keys to navigate doesn't seem to work.

**To Reproduce**

This can also be seen on the examples website.

1. Go to https://reactdatepicker.com/#example-inline-version (Inline version example)
2. Click on one of the month arrows to visibly have the focus there
3. Click tab to go into dates
4. See the focus going out of the calendar

**Expected behavior**
Dates should be accessible via keyboard independent of the calendar having `inline` set to true.

**Screenshots**
Tabbing here:
![tabbing](https://user-images.githubusercontent.com/3294597/81727399-35aa3900-9489-11ea-9f0a-c3b044aff375.gif)

**Desktop (please complete the following information):**

- OS: iOS
- Browser: Chrome

miina:
Trying to use the calendar inline [in a plugin](https://github.com/google/web-stories-wp/pull/1743/files#diff-506f36fac1a9c8ae575f0854723a283eR44-R49), however, the keyboard accessibility doesn't seem to work in case of `inline`.
.
Assuming it's a bug since it doesn't work on the demo website either, however, it would be good to know if it needs a special configuration instead.

arsnyder16:
@miina I am seeing the same behavior with inline. For me when i navigate with arrows through the days it seems to actually be moving but the class is not getting applied to the element that will show the styling that it is keyboard selected. There is a specific check for inline that will always return false for keyboard selected 

![image](https://user-images.githubusercontent.com/6554919/82330839-530f6380-99b1-11ea-8bbf-4ce41f682c8c.png)

I am not sure of the history of why that is there.

Issue #1337, opened 2018-03-29
With 2 months shown and inline calendar, the months shift when range selected spans the 2 months


sshilal1:
I have an inline calendar with 2 months shown. Whenever a user selects a date range that spans the 2 months, if the range is not in the same month, the left month slides out of view. As seen in the gif below, In order to view the full range, a user must select the back arrow to see the range again.

![datepicker1](https://user-images.githubusercontent.com/15115953/38110725-90294006-336a-11e8-8994-8e698c95c3d8.gif)

Here is my code
```
<DatePicker
  inline
  selected={this.state.startDate}
  startDate={this.state.startDate}
  endDate={this.state.endDate}
  onChange={this.handleDateSelect}
  monthsShown={2}
  shouldCloseOnSelect={false}
/>
```

I feel as if there should be a prop to enable / disable this.

aij:
I don't think the months were intended to shift when selecting a day. It seems like a buggy interaction between `monthsShown` and `inline`. (Normally, the date picker would close after selecting a date, so re-opening to the last selected date is fine/expected, but staying open and shifting is weird.)

fooberichu150:
We're running into this issue as well.  I don't mind so much that it shifts the calendar as that when you're on a mobile device (I think it may just be iOS) the previous date placeholder on the 2nd month still shows highlighted until you "touch" somewhere else on your screen.  Getting this bug fixed would at least also prevent that associated display issue from showing up.

Before:
![screenshot_before_ios](https://user-images.githubusercontent.com/27869244/48450226-b1c3a380-e762-11e8-888f-e44458062af8.jpg)

After (December 13th being my selection):
![screenshot_after_ios](https://user-images.githubusercontent.com/27869244/48450238-c0aa5600-e762-11e8-8e22-f092b20940b4.jpg)

martijnrusschen:
Should be fixed now

smithaitufe:
Please I need to know how this was eventually resolved.

How do you know if it is the startDate that was selected or if the endDate that was selected?

smeletis:
This is not fixed, why is this issue closed?

eivindchr:
> This is not fixed, why is this issue closed?

I fixed this using shouldCloseOnSelect={false}

BeKnowDo:
This is definitely not fixed... 2021 😆

singladno:
> This is definitely not fixed... 2021 😆

Have you tried this?
`focusSelectedMonth={true}`

[...]

PR #2153, opened 2020-05-19
remove inline check for keyboard selected

arsnyder16:
When using an inline date picker, you can navigate the days with the keyboard but they do not style correctly so it doesn't appear that it is working

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -92,7 +92,6 @@ export default class Calendar extends React.Component {
     includeDates: PropTypes.array,
     includeTimes: PropTypes.array,
     injectTimes: PropTypes.array,
-    inline: PropTypes.bool,
     locale: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.shape({ locale: PropTypes.object })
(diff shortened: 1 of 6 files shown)


arsnyder16:
#2141

arsnyder16:
#2141

arsnyder16:
#1541  
#1337

arsnyder16:
@martijnrusschen Any idea why the inline check was there ?

arsnyder16:
@martijnrusschen  Any feedback would be appreciated.  Waiting for this to fix #2141

martijnrusschen:
I didn't check this yet since the build was broken. Can you check what's going wrong?

arsnyder16:
@martijnrusschen  Sure its just a test that is checking the exact condition that i am removing,but there doesn't seem to be any indication to why the check is needed. It looks more like the test was just written based on the logic not for any specific use case

arsnyder16:
@martijnrusschen I went ahead and just updated the test. All green.

arsnyder16:
@martijnrusschen Any update here?

PR #2157, opened 2020-05-22
Add two column month display layout flag for month picker

LuminousSpike:
Add a flag to allow month pickers to show in a two column month display layout


dipikas:
This is perfect, exactly what we need :)

GabrielMMSestrem:
I'd really like to have this feature!

PR #2164, opened 2020-05-29
Update onChangeRaw and onSelect - Fixes #1951

### LINKED ISSUE #1951: onSelect fires on every update, instead of every click
**Describe the bug**
From the README:
> `onSelect={this.handleSelect} //when day is clicked`

This indicates that `onSelect` will fire every time that you click on a day inside the calendar, but will _not_ fire when you are updating the text input.

However, this is not the case. `onSelect` fires alongside `onChange` with no discernible difference between the two. This is causing a problem, as I am working on something that I want to _only_ happen when a day is clicked, and is independent of the text input.

**To Reproduce**

1. Create a datepicker as follows:
```
<DatePicker
  selected={startDate}
  onChange={(date) => {
    console.log('onChange fired');
    setStartDate3(date);
  }
  onSelect={() => console.log('onSelect fired')}
/>
```

2. Click on a date in the calendar. See messages for both `onChange` and `onSelect` (this is fine)
3. Change text in input. See messages for both 'onChange' and 'onSelect' (this is NOT fine).

**Expected behavior**
`onSelect` should only fire when you click on a date in the calendar, as the readme indicates.

If this _is_ the expected behavior, then the docs should be updated, but I would then ask; why does `onSelect` exist? If it does the exact same thing as `onChange`, then what's the point of it's existence? It seems much more useful to have `onSelect` work like the docs indicate it works.

**Screenshots**
N/A

- OS: [e.g. iOS] Windows
- Browser [e.g. chrome, safari]: Chrome
- Version [e.g. 22]: 2.9.6

N/A

**Additional context**
N/A

tsrahm:
1. Updates the onChangeRaw callback such that it fires on all changes, including when a date is selected from the calendar. It would be useful to users if there was a callback that fires on every change. For example, I am performing date validation for user error messaging and it's necessary to perform that validation on every change. 

2. Updates the onSelect callback so it only fires for a calendar date selection. Currently, onSelect and onChange fire on the same events. It would be helpful to differentiate the two. This change also matches the behavior to the docs.

Fixes #1951

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -455,6 +455,9 @@ export default class DatePicker extends React.Component {
       );
       return this.preventFocusTimeout;
     });
+    // if (this.props.onChangeRaw) {
+    //   this.props.onChangeRaw(event);
+    // }
     this.setSelected(date, event, false, monthSelectedIn);
     if (!this.props.shouldCloseOnSelect || this.props.showTimeSelect) {
       this.setPreSelection(date);
@@ -501,9 +504,8 @@ export default class DatePicker extends React.Component {
       this.props.onChange(changedDate, event);
     }
 
-    this.props.onSelect(changedDate, event);
-
     if (!keepInput) {
+      this.props.onSelect(changedDate, event);
       this.setState({ inputValue: null });
     }
   };
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -455,9 +455,9 @@ export default class DatePicker extends React.Component {
       );
       return this.preventFocusTimeout;
     });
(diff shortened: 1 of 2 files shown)


martijnrusschen:
Hi thanks for this, can you add a test to make sure we keep this behavior in the future as well?

tsrahm:
Sure, no problem.

tsrahm:
My original fix for the broken maxDate test did not address the root cause. The new fix does by passing in the dateFormat used in the test since it doesn't match the default dateFormat.

tsrahm:
Thanks, happy to contribute.

Issue #2167, opened 2020-06-01
Initial value from the beginning and not from the end?


AndrewOkonar:
User wants to set/edit a new date for the deadline. And the user should choose future date while editing this field. but as you see all selected value from "showYearPicker" stared from the end (screenshot 1). is it possible make selected value always first in the container? (screenshot 2)

1)
![image](https://user-images.githubusercontent.com/23657360/83421613-0efc7400-a431-11ea-8281-6154aca09bf5.png)

2)
![image](https://user-images.githubusercontent.com/23657360/83422277-f50f6100-a431-11ea-9bba-745d29a5e1ea.png)

martijnrusschen:
Fixed by https://github.com/Hacker0x01/react-datepicker/pull/2186

Issue #2162, opened 2020-05-29
YearPicker does not highlight selected year


LuukvE:
**Describe the bug**
YearPicker does not highlight selected year

**To Reproduce**

1. Add `showYearPicker` prop to `<DatePicker />` component
2. Click on a year
4. See the lack of highlighting

**Expected behavior**
I expect the selected year to be highlighted

PR #2168, opened 2020-06-01
Add selected class for year picker - Feature #2162

3bOONAL:
Feature for [#2162](https://github.com/Hacker0x01/react-datepicker/issues/2162) I've noticed that this behavior wasn't implemented. 
Added some logic to add class.
Added class name to styles the same way as for months.
Covered with a test.

Also, I noticed this issue [#2167](https://github.com/Hacker0x01/react-datepicker/issues/2167) I think I could check this. Maybe #2167 needs to add a new prop to choose where to show selected year(beginning or end of the calendar) ?

### DIFF
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -364,7 +364,8 @@
 }
 
 .react-datepicker__month,
-.react-datepicker__quarter {
+.react-datepicker__quarter,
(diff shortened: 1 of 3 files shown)


martijnrusschen:
Looking good

PoltP:
@martijnrusschen Martijn, could you please clarify, why these changes is not available in version 3.1.3?

PR #2179, opened 2020-06-09
border bottom right time section

whatever555:
Fix for rendering issue on time section:
Bug visible on Chrome: Version 81.0.4044.138 (Official Build) (64-bit)

image


martijnrusschen:
Nice catch!

Issue #2166, opened 2020-06-01
Selected value first in the box


AndrewOkonar:
I'm using "year picker" for input where the user should select the next data for future events.

```
      {jobStartsAtFormat === 'specific_date' && <S.DatePickerOptional
        selected={job.startsAt && moment(job.startsAt).toDate()}
        onSelect={handleDate('startsAt')}
        showYearPicker
        dateFormat="yyyy"
        customInput={<InputRef
          name="startsAt"
          label="Eintrittsdatum"
          error={errors.startsAt}
          placeholder="Gibt es eine Eintrittsdatum?" /> } />}
```

![image.png](https://images.zenhubusercontent.com/596cb2bd61b4ca3b0a74a9a9/fc69cef4-364d-41f5-8257-ab30be01d632)

And as you see selected year always the last in the box. is it possible to make always first in the box and show future years?

martijnrusschen:
Fixed by https://github.com/Hacker0x01/react-datepicker/pull/2186

Issue #2154, opened 2020-05-19
Year Picker min/max dates.


essedger:
**Describe the bug**
minDate and maxDate doesn't work correctly. 

**To Reproduce**
add to <DatePicker/>
showYearPicker
dateFormat="yyyy"
minDate={startDate}
maxDate={endDate}

**Expected behavior**
For example dates:
const [startDate, setStartDate] = useState(new Date("2001/02/08"));
const [endDate, setEndDate] = useState(new Date("2015/04/08"));

Picker shows 3 pages and doesn't disable values over the period.
Also it duplicats 2001 and 2012 years. 

**Screenshots**
![image](https://user-images.githubusercontent.com/12384438/82347748-b07ed100-9a00-11ea-878e-9a2bcdfed260.png)
![image](https://user-images.githubusercontent.com/12384438/82347793-bc6a9300-9a00-11ea-9964-1269370f12c8.png)
![image](https://user-images.githubusercontent.com/12384438/82347813-c391a100-9a00-11ea-84b0-775193d5c811.png)

- OS: Windows 10
- Browser [chrome, safari, etc]
- Version [2.15.0]

Marc19:
Yeah, I am facing the same problem. Is there at least a way to give a className for out of range dates?

martijnrusschen:
Fixed by https://github.com/Hacker0x01/react-datepicker/pull/2186

Marc19:
How can I get this fix? I am trying 'npm update' but there is no update.

PR #2186, opened 2020-06-16
YearPicker upgrades

3bOONAL:
Update work with change year by next/previous buttons,
update render of years with saving position of years,
add preselected year styles,
cover by tests for next/previous buttons and showing periods of years,
add work with min/max dates.
solutions for #2167, #2166, #2154


martijnrusschen:
Thanks, this is great work!

PR #2198, opened 2020-06-27
add function of back current month in custom header

NagaiKoki:
when using react datepicker and this custom header function, you cannot be back current month easily.
For example, when you selected 09/06/1995, and want to back current year and month,  you need to chose them by using pull down of custom header or make like this code.

```js
const months = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]

const handleBackCurrentMonth = (
    changeYear,
    changeMonth
  ) => {
    const now = new Date()
    const yearValue = now.getFullYear()
    const monthValue = months[now.getMonth()]
    changeYear(yearValue)
    changeMonth(monthValue - 1)
  }
```
※ `changeYear` and `changeMonth` function is custom header fucntion

It would be great if there is default function of back current month in custom header.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -267,6 +267,12 @@ export default class Calendar extends React.Component {
     );
   };
 
+  backCurrentMonth = () => {
+    this.setState({ date: new Date() },
+      () => this.handleMonthChange(this.state.date)
+    );
+  };
+
   handleDayClick = (day, event, monthSelectedIn) =>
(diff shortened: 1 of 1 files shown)


martijnrusschen:
Can you add a test to make sure this keeps working in the future?

NagaiKoki:
> Can you add a test to make sure this keeps working in the future?

Sure 

> Can you add a test to make sure this keeps working in the future?

Sure, I'll do soon.

Issue #584, opened 2016-10-11
Date Range with single DatePicker


hoodsy:
Is this at all possible? The example for date range is exactly the functionality I would want, but it would be best if the user is selecting the range on a single calendar, not through 2 separate inputs.

rafeememon:
Yep, it's possible -- you'll need to define your own way of setting `startDate` and `endDate` based on `onChange` events.

hoodsy:
@rafeememon sweet! thanks for the quick response. How about the hover styling, to show the range. It seems like that is controlled by ```selectsStart``` and ```selectsEnd```

rafeememon:
Check out the example on the [demo page](https://hacker0x01.github.io/react-datepicker/#example-6)

rafeememon:
Actually, I don't think we currently support having `selectsStart` and `selectsEnd` on the same picker.  Not sure how the semantics of this would work, but PRs to support this are welcome.

sahanarula:
Looking forward for the best solution.

TrejGun:
would love to see this feature too

martijnrusschen:
I agree, would love to see if someone could diff up a PR.

TrejGun:
probably its better to use https://github.com/airbnb/react-dates :)

Charlie91:
Ok, it's possible through the setting startDate and endDate, but how can I display two dates at one input field? It seems that I can only display one date at one time

marinarjumina:
May be it helps someone in the future:

### Get `weekStartDay` and `weekEndDay`:

const weekStartDay = startOfWeek(currentDate, { weekStartsOn: 1 })
const weekEndDay = endOfWeek(currentDate, { weekStartsOn: 1 })

### Prepare `inputValue` range dates output:

const inputValue = format(weekStartDay, "dd/MM/yyyy") + ' - ' + format(weekEndDay, "dd/MM/yyyy")_

### Pass formatted `inputValue` to Datepicker's `value`:

<DatePicker selected={weekStartDay} highlightDates={highlightDates} onChange={setCurrentDate} locale="enGB" **value={inputValue}** />

For me it works.

vedadeepta:
Recently came across this use-case. This can be implemented as follows:

```

  handleChange = date => {
    const { format } = this.props
    if (date) {
      if (!this.state.date) {
        return this.setState({ date })
      }
      this.setState({
        endDate: date,
[...]

PR #2204, opened 2020-06-29
Range on one datepicker for #584

3bOONAL:
Tried to implement one datepicker for range. Thanks to the guys from #584 discussion, their examples are the base for this PR. I hope this will help.


martijnrusschen:
Thanks for adding support for this!

PR #2209, opened 2020-07-01
Add closeOnScroll prop

mjchang:
Add closeOnScroll prop to enable. The prop accepts either a boolean or a function. If true, the datepicker will be closed when the document is scrolled. If function, then it executes the function to see whether the datepicker should be closed or not.


martijnrusschen:
If you can help me resolving the merge conflicts, we can get this one merged.

PR #2230, opened 2020-07-09
fix(date_utils): fixing navigation on yearpicker

hillterence007:
When using the yearpicker the navigation is disappearing while clicking on next. The user can´t get to the initial state.


martijnrusschen:
Thanks, can you add a test to make sure we don't break this in the future?

hillterence007:
> Thanks, can you add a test to make sure we don't break this in the future?

Done

saswat3115:
Is this fix publidhed to npm ? I dont see this changes in version 3.1.3 @hillterence007 @martijnrusschen

PR #2255, opened 2020-07-09
fix: docs

mxschmitt:
Let's see if CI passes. This time we build the docs also on CI, should be still continue to work I think.


martijnrusschen:
I was just thinking that we should add some check to make sure the docs are working well before the build goes green, maybe  a simple test might be sufficient

Issue #2331, opened 2020-08-14
Provide a way to create DatePicker for Years with NxM grid for year values


PoltP:
I create date picker for Years and our design contains 4x4 grid for year values, for sample:
```
2015  2016  2017  2018
2019  2020  2021  2022
2023  2024  2025  2026
2027  2028  2029  2030
```
Now, there is no way to implement it.

As you know, inside '../src/date_utils.js':
```
export function getYearsPeriod(date) {
  const endPeriod = Math.ceil(getYear(date) / 12) * 12;
  const startPeriod = endPeriod - 11;
  return { startPeriod, endPeriod };
}
``` 
we see '12' as a number of years. Also, this number is used inside another related code (yearsDisabledBefore, yearsDisabledAfter etc.).

Could you implement a property to set number of viewer years or grid dimensions NxM similar to showTwoColumnMonthYearPicker?
I reviewed code and it looks not so hard as hard for us to fix it on our side.

PR #2345, opened 2020-08-20
Add property to show specific year item number for year picker - Feature #2331

PoltP:
Feature for #2331 I've noticed that this behavior wasn't implemented.
Instead of implementing NxM year values grid, I just changed magic constants for year periods to provided property. Using this property and changing CSS it is easy to show NxM year values grid.
I've added property `yearItemNumber` to show specific number of year items.
Corresponding behavior has been covered with tests.
Example provided.


shiftrtech:
You forgot to update the [types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-datepicker/index.d.ts) y have an issue with the attribute `yearItemNumber` and typescript.

Tossko:
same here

TibiTotoro:
Me too

PR #2409, opened 2020-09-25
Bugfix: Not setting focus to selected day on page load for inline calendar

bldunn-dev:
This is a bug fix.  Our calendar is displayed inline, and we need to set the selected date when our upon page load.  However, since the user at this point has not done anything, document.activeElement === document.body, and the selected day is getting focus due to handleFocusDay() in day.jsx.
I tried to make this as low-impact a fix as possible, but it still required a lot of prop drilling.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -98,6 +98,7 @@ export default class Calendar extends React.Component {
     includeDates: PropTypes.array,
     includeTimes: PropTypes.array,
     injectTimes: PropTypes.array,
+    inline: PropTypes.bool,
     locale: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.shape({ locale: PropTypes.object })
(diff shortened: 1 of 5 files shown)


martijnrusschen:
Thanks for this, can add a test to make sure this doens't break again in the future?

bldunn-dev:
I tried to add a unit test but, when I used enzyme's "mount" function inside of day_test.js -- because testing focus requires "mount" -- the line of code I changed in day.jsx did not get executed.  However, that line _did_ get executed when I ran calendar-text.js.  Would it be okay to add a unit test for this inside calendar_test.jsx, or some other test file, or must it be a part of day_test.js?

bldunn-dev:
Sorry, I have to close this PR because I made the commit using my employer-provided email address, and this is a violation of company policy. I will create a new PR using only my personal GitHub account.

Issue #2378, opened 2020-09-07
need a filterTime feature


Riba-Kit:
Hi! In my project I need to filter times in time-picker select-list. It would be nice if I can use filterTime just as filterDate.

alanunesouza:
I also have the same problem ... I need to set the time range per day and I had to look for another alternative.

Jonofat:
> I also have the same problem ... I need to set the time range per day and I had to look for another alternative.

Did you find something else? I want to be able to choose different time options for weekdays and different time options for weekends.

ArchanaSharma95:
I am having same issue, I want to disable the past time using "filterTime" prop. Please suggest if you find any other solution.

saltire:
Hey folks, there is a filterTime property now, merged from PR #2420.

https://reactdatepicker.com/#example-filter-times

ymh357:
I copied exactly the same code from the example, but it not works as expected. Date larger than today still applies the filter.

`const filterPassedTime = time => {
    const currentDate = new Date();
    const selectedDate = new Date(time);

    return currentDate.getTime() < selectedDate.getTime();
  } `

I want only the time before today to be disabled, not after today. Could you please have a look?

Riba-Kit:
> > I also have the same problem ... I need to set the time range per day and I had to look for another alternative.
> 
> Did you find something else? I want to be able to choose different time options for weekdays and different time options for weekends.

I was generated full list of times instead of filtering) but I see that something was merged

Riba-Kit:
> Hey folks, there is a filterTime property now, merged from PR #2420.
> 
> https://reactdatepicker.com/#example-filter-times

Thanks a lot! how can I close this issue?

ishvar-kanajariya:
> I copied exactly the same code from the example, but it not works as expected. Date larger than today still applies the filter.
> 
> `const filterPassedTime = time => {
> const currentDate = new Date();
> const selectedDate = new Date(time);
> 
> ```
> return currentDate.getTime() < selectedDate.getTime();
> ```
> 
> } `
> 
> I want only the time before today to be disabled, not after today. Could you please have a look?

Try this one!
  filterPassedTime = (time: any) =>{
    const currentDate = new Date(time);
    const selectedDate = this.state.selectedDate; 
    if (selectedDate.getDate() === new Date().getDate()) {
      return currentDate.getTime() > selectedDate.getTime();
    }
    return currentDate.getTime() < selectedDate.getTime();
  }

PR #2420, opened 2020-09-29
Add filterTime property

saltire:
Works similarly to filterDate. Includes tests and an example. Let me know if there are any problems!

Related issue: #2378


ymh357:
Having a problem that the time of date larger than today are filtered unexpectedly.

`
export default function DatetimeSelector(props: IProps){
  const {onCancel, onConfirm} = props
  const [startDate, setStartDate] = useState(
    setHours(setMinutes(new Date(), 0), 9)
  );
  const filterPassedTime = time => {
    const currentDate = new Date();
    const selectedDate = new Date(time);

    return currentDate.getTime() < selectedDate.getTime();
  }
  return (
    <DatePicker
      selected={startDate}
      showTimeSelect
      filterTime={filterPassedTime}
      dateFormat="MMMM d, yyyy h:mm aa"
    />
  );
)
`

PR #2432, opened 2020-10-06
fix autofocus on selected day when inline prop is true

bldunn-dev:
This fixes a big issue we were having -- the inline DatePicker at the bottom of our home page was grabbing focus upon page load.  This caused an accessibility issue where, if the user pressed the Tab key, they were unexpectedly on the footer instead of on the first clickable link.  

I made this a very narrow fix to only change the one line in day.jsx that was causing the issue, but it required a lot of prop drilling to get down to it.  I added a couple of unit tests to test when inline=false and =true.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -98,6 +98,7 @@ export default class Calendar extends React.Component {
     includeDates: PropTypes.array,
     includeTimes: PropTypes.array,
     injectTimes: PropTypes.array,
+    inline: PropTypes.bool,
     locale: PropTypes.oneOfType([
       PropTypes.string,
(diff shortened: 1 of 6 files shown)


martijnrusschen:
Thanks!

jmmccasland-hint:
This still seems to be an issue for my team in Safari on react-datepicker `v3.2.2`. We're using focus-trap-react when opening a side drawer and the current day seems to be catching focus when the component mounts rather than focusable elements further up the dom tree. 

Has this been released yet?

jmmccasland-hint:
No longer an issue in v3.3.0. Thanks for your work @martijnrusschen and @bldunn-dev!

Issue #2284, opened 2020-07-22
Inline version selects multiple dates 


MD-AZMAL:
**Describe the bug**
In Inline Version when you select some other date the initial selected date remains the same with blue background
**To Reproduce**

1. Go to 'inline version or inline portal version in docs'
2. Click on any date
4. See error when multiple dates are selected

**Expected behavior**
Only one date is supposed to be selected
**Screenshots**
![Screenshot from 2020-07-22 14-57-06](https://user-images.githubusercontent.com/33421347/88160168-ee2fee00-cc2b-11ea-90ab-671b8d20a54f.png)

**Desktop (please complete the following information):**

- OS: Ubuntu
- Browser Chrome

jxa8057:
I'm also running into this, where the current day of the month always appears selected, event when changing months.

MD-AZMAL:
@jxa8057 do let me know if you find any solution

PoltP:
It can be easily reproduced using [demo](https://reactdatepicker.com/#example-inline-version) after "Next" or "Previous" button is clicked. This bug makes component unusable in production.

MD-AZMAL:
@PoltP yes i think so, is there any work around ?

PoltP:
@MD-AZMAL I have not found any usable workaround, now I've temporary disabled keyboard navigation using `disabledKeyboardNavigation` property.

MD-AZMAL:
@PoltP okay got it! still keeping this issue please share if you find any other solution

camilaazuma:
Same issue here, anyone got updates? 👀

camilaazuma:
Hey @MD-AZMAL, made this is a workaround until the lib is properly fixed. Basically:

1. Make sure to have a `ref` to the calendar
2. Whenever a day is clicked, search for the `keyboard-selected` class and removes it from the element.
3. Then adds the same class to the current selected day

This only solves the layout issue. Focus is still on original `keyboard-selected` element.
Don't forget to declare the `selected` on your `useEffect`, so it triggers every time the date changes.

```
  useEffect(() => {
    if (inline) {
      const keyboardSelected = ref.current.calendar.componentNode.querySelector('.react-datepicker__day--keyboard-selected');
      const mouseSelected = ref.current.calendar.componentNode.querySelector('.react-datepicker__day--selected');

      if (keyboardSelected && mouseSelected) {
        keyboardSelected.classList.remove('react-datepicker__day--keyboard-selected');
[...]

PR #2436, opened 2020-10-09
Feature hide duplicate days

fredrikabsa:
This PR does two things. 

1. Solves #2284 . As described in issue, keyboard focus does not update in inline version. By updating preSelection date when changing date by click the keyboard date and selected date will be in sync.
Before:
before
After:
after

2. While solving 1, i broke a the test: "should not switch months in inline mode when a day is clicked". After looking into it i saw that the test is not actually testing if the month is switched or not. But i found the case interesting. Firstly, i mean that in default inline case, if you click on a outside-month day, it should change month (and it currently does). Secondly believe it should be the case if its multiple months shown. The current version will switch between months when clicking on an outside-month day (the inline design of two months is another issue): 
before
My take on a solution: Why show duplicate days at all when viewing multiple months?
after
This also effects non-inline version: 
Screenshot 2020-10-09 at 15 53 22

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -761,6 +761,8 @@ export default class Calendar extends React.Component {
       var monthsToAdd = i - this.props.monthSelectedIn;
       var monthDate = addMonths(fromMonthDate, monthsToAdd);
       var monthKey = `month-${i}`;
+      var monthShowsDuplicateDaysEnd = i < (this.props.monthsShown - 1);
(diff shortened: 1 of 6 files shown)


martijnrusschen:
Hi, this is great. Thanks for putting this together. I agree with your assessment.

sebastienbarre:
This didn't solve the issue with inline calendar rendering different dates twice, unfortunately. aka 2284

PR #2450, opened 2020-10-14
Bind renderTime to selected date

### LINKED ISSUE #2449: function filterTime is not binded to date
I wanted to filter time that has already passed but noticed that component Time render values based on current date.
I expected Time will render values based on selected date.

**Example filter** (Changing the date will not change anything):
```
const filterTime = (time) => {
    const currentDate = new Date();
    const selectedDate = new Date(time);

    return currentDate.getTime() < selectedDate.getTime();
};
```

mroforolhc:
fixed #2449


martijnrusschen:
Thanks

ysinchenko:
is it deployed in npm as alpha ?

Issue #1721, opened 2019-04-25
Current month is not shown in custom header with multiple months inline


Helpermedia:
### Expected behavior
Multiple months (monthsShown={3}) have a current month title in the header. 
With renderCustomHeader a custom header can be made, current month is not an option to add. I would expect it would be still visible.

### Actual behavior
Current month is not longer shown when renderCustomHeader is used.

### Steps to reproduce
Implement renderCustomHeader with multiple months inline. Current month is not shown.

brojd:
any update on this?

Zertz:
Looking at the code, it seems like [this behavior is intentional](https://github.com/Hacker0x01/react-datepicker/blob/99d45b9e884ae89a1a0f9b43f1eb3c74d62dbf7a/src/calendar.jsx#L673).

I've started working on #2520 to fix this.

PR #2520, opened 2020-11-04
fix: display default header in additional months

Zertz:
This is a quick fix to test things out manually on my end, I'll clean up related code and update automated tests shortly.

Related to #1721

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -739,7 +739,7 @@ export default class Calendar extends React.Component {
 
   renderHeader = headerArgs => {
     switch (true) {
-      case this.props.renderCustomHeader !== undefined:
+      case this.props.renderCustomHeader !== undefined && headerArgs.i === 0:
         return this.renderCustomHeader(headerArgs);
       case this.props.showMonthYearPicker ||
         this.props.showQuarterYearPicker ||


lironezra:
Hi Zertz, 
Any update on that? did you add something to achieve that functionality? 
I have just tried to display two month calendars with custom header for both and it's not seems to works...

Thanks a lot!

Zertz:
Hey, I ended up [creating a fork](https://github.com/Zertz/react-d8picker) because I stumbled upon quite a few blockers for our particular use case.

It's still a work in progress but the general idea was to dramatically reduce the number of props by switching to render props.

lironezra:
Hi,

Do you have any estimation about it? 

Thank you!

Zertz:
Probably sometimes in December but keep in mind there is no migration path. It will do a lot less out of the box and most customization will be done through render props instead.

lironezra:
Hi Zertz,

Just to make things clear,  are you planning on adding this feature / fix to the original repository as well ?

Zertz:
Yes.

Zertz:
It turned out to be trickier than I thought, so unfortunately I won't.

PR #2521, opened 2020-11-05
Added link to docs for popper component

boatcoder:
Had a hard time figuring out what the props were for the popper stuff until I discovered popper was a separate component.  So, I added a description that contains a link to the documentation for the popper.


boatcoder:
This Resolves #2522

PR #2564, opened 2020-12-01
Day: custom day rendering options

beaumontjonathan:
Hi, firstly thank you for your amazing work on this datepicker! 😄

This PR introduces an extra parameter to the day `renderDayContents` prop, which allows passing down internal information about the give day. It uses the same names (after converting to JS rather than CSS convention) as the class names.

I found CSS selectors were getting pretty convoluted when I wanted to change styling of the day contents depending on these properties, and thought it could be useful if they were simply passed down into `renderDayContents`.

- The appropriate test has been updated to verify that an object is passed as a third parameter
- The `renderDayContents` example has been updated to show this information being used, though understand this might not be the best example to show the functionality
- This should not be breaking

Please let me know what you think about this. Am willing to help do further work to get it in. Thanks!!

### DIFF
--- a/docs-site/src/examples/renderCustomDay.js
+++ b/docs-site/src/examples/renderCustomDay.js
@@ -1,7 +1,7 @@
 () => {
   const [startDate, setStartDate] = useState(new Date());
-  const renderDayContents = (day, date) => {
-    const tooltipText = `Tooltip for date: ${date}`;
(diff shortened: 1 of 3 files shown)


martijnrusschen:
This is an interesting proposal. I'd say if you want to push this in, let's make some tests that cover the functionality and add an example to the docs site so that people can discover this functionality as well.

Issue #2367, opened 2020-09-01
Accessibility issue of time picker 


devangpatel21:
**Describe the bug**
cannot choose a time with tab click I don't have tab index = 0

**To Reproduce**

1. Select date picker with time 
2. open picker through the keyboard and go to time picker to choose the time 
3. Won't be able to pick time through the keyboard

**Expected behavior**
time should also able to be picked  through the keyboard

kennygrubb:
Yes this is an issue for me as well. Thanks

lpelypenko:
Trying to fix this issue in https://github.com/Hacker0x01/react-datepicker/pull/2566. 
Please vote for my fix, hopefully, it will help to release it faster.

PR #2566, opened 2020-12-03
Fix for accessibility issue 'Elements that have scrollable content should be accessible by keyboard' for scrollable time panel 

lpelypenko:
Fixed 'Elements that have scrollable content should be accessible by keyboard' WCAG violation for the scrollable time panel.
Before the fix "axe" chrome plugin was showing the following error: 

image

After the fix "axe" chrome plugin was not showing "Elements that have scrollable content should be accessible by keyboard" and it is possible to scroll time panel
Screen Shot 2020-12-03 at 16 12 03

I have followed the recommendation from dequeuniversity: 
https://dequeuniversity.com/rules/axe/4.0/scrollable-region-focusable?application=AxeChrome

### DIFF
--- a/src/time.jsx
+++ b/src/time.jsx
@@ -164,6 +164,7 @@ export default class Time extends React.Component {
             this.centerLi = li;
           }
         }}
+        tabIndex="0"
       >
         {formatDate(time, format, this.props.locale)}
       </li>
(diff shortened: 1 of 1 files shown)


lpelypenko:
Please link with the issue: [Accessibility issue of time picker #2367](https://github.com/Hacker0x01/react-datepicker/issues/2367)

lunatik-210:
Really important thing guys!

martijnrusschen:
Thanks for fixing!

PR #2389, opened 2020-09-11
Adds keyboard navigation for month picker

camilaazuma:
This PR:
- Allows the user to pick a month using keyboard only
- Adds test coverage for recent changes
- Updates readme with new keybinds
- resolves #2078

PR #2579, opened 2020-12-08
Replace fill method with different syntax for IE compatibility

### LINKED ISSUE #2577: fill-method causes DatePicker to crash in IE
**Describe the bug**
In version 3.2.0, PR #2389 a bug was introduced by using the fill-method that is not supported in IE

**To Reproduce**
Open DatePicker in Internet  Explorer and it will crash with the error message stating that the property or method "fill" is not supported

**Expected behavior**
The DatePicker shouldn't crash in Internet Explorer as according to the docs, IE>=10 is still supported

- OS: Windows 10
- Browser Internet Explorer
- Version 11

relwiwa:
Fixed bug introduced in version 3.2.0 (PR #2389).
The fill method was used, but it is not supported in IE.
The method was removed and replaced with different syntax that gets transpiled and works in IE.

Fixes #2577

### DIFF
--- a/src/month.jsx
+++ b/src/month.jsx
@@ -62,7 +62,7 @@ export default class Month extends React.Component {
     monthShowsDuplicateDaysStart: PropTypes.bool
   };
 
-  MONTH_REFS = Array(12).fill().map(() => React.createRef());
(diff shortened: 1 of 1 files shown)


martijnrusschen:
Thanks for fixing this!

PR #2588, opened 2020-12-11
Fix keyboard navigation in inline version

### LINKED ISSUE #2578: Keyboard navigation is not working properly in inline version
**Describe the bug**
In version 3.3.0 with commit "fix autofocus on selected day when inline prop is true" the keyboard navigation in the inline version of the DatePicker became buggy.
When switching from one month to another, the focus gets lost and it is no longer possible to continue navigation with the keyboard.

**To Reproduce**
1. Enter the inline version of the DatePicker via keyboard, e.g in the [Inline Version example](https://reactdatepicker.com/#example-inline-version)
2. Use arrow keys, PageUp/PageDown, Pos1/End keys to navigate to a different month
3. The focus gets lost and it is not possible to continue navigation with the keyboard

**Expected behavior**
Keyboard navigation in the DatePicker's inline version should work

- OS: Windows 10
- Browser Chrome, FF, IE11

relwiwa:
When the autoFocus issue of the inline version was fixed in v3.3.0 ([this commit](https://github.com/Hacker0x01/react-datepicker/commit/d09a14dc3d7e61d59069cd51acc9e78bc822524e)), the keyboard navigation stopped working after switching between months with ArrowLeft/Right/Up/Down, PageUp/Down, End/Home keys.

With this PR the autoFocus issue remains fixed, but keyboard navigation got restored

fixes #2578

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -99,6 +99,7 @@ export default class Calendar extends React.Component {
     includeTimes: PropTypes.array,
     injectTimes: PropTypes.array,
     inline: PropTypes.bool,
+    shouldFocusDayInline: PropTypes.bool,
     locale: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.shape({ locale: PropTypes.object })
@@ -798,6 +799,7 @@ export default class Calendar extends React.Component {
             selectingDate={this.state.selectingDate}
             includeDates={this.props.includeDates}
             inline={this.props.inline}
+            shouldFocusDayInline={this.props.shouldFocusDayInline}
             fixedHeight={this.props.fixedHeight}
             filterDate={this.props.filterDate}
             preSelection={this.props.preSelection}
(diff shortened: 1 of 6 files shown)


relwiwa:
I wanted to add tests for keyboard navigation in inline version, but unfortunately didn't manage to get them to fail.
It seems the two tests that were added in [this commit](https://github.com/Hacker0x01/react-datepicker/commit/d09a14dc3d7e61d59069cd51acc9e78bc822524e) don't seem to fail either. @bldunn-dev Maybe you want to have another look?

relwiwa:
I added tests for the new state value to have a proper test coverage. Testing the keyboard focussed element when switching between months with keyboard would have been better, but I couldn't get it working with Enzyme/React testing tools. This would be easier to test with e2e-Tests.

martijnrusschen:
Thanks, this is great. Thanks for fixing and adding tests!

PR #2616, opened 2021-01-03
docs: order of "Getting setup" in CONTRIBUTING.md

czabaj:
The `yarn install` shall precede `yarn build`, that's all...

Oh, and using ordinals in markdown ordered list is not needed, it just needs to be number, so I changed the ordinals to `1.` everywhere, it is easier to maintain -- there is no need to reorder when changed.


martijnrusschen:
Nice catch

PR #2252, opened 2020-07-09
Upgrade popper

martijnrusschen:
Migration guide: https://popper.js.org/docs/v2/migration-guide/

ref https://github.com/popperjs/react-popper/pull/333

PR #2626, opened 2021-01-04
chore!: upgrade react-popper

czabaj:
This is based on #2252 but originates from the most recent `master`.

It is almost all the same as in #2252, aside from BREAKING CHANGE in props -- the `popperModifiers` is now an array as mentioned [in popper migration guide](https://popper.js.org/docs/v2/migration-guide/#10-update-modifiers).

Also, I changed the React version in **docs-site** to align with react-datepicker peerDependency range `^16.9.0` if you expect the react-datepicker to be compatible with react@17, I can adjust the peerDependency range to reflect that.

I had a huge hurdle with React hooks caused by _two versions of React in application_ which crashed the **docs-site** during development. Thanks to [this issue](https://github.com/facebook/react/issues/13991), I realized, one version is bundled in the **docs-site** and the second version is bundled in react-datepicker as devDependency. This issue exists even when both versions of react are same version, as react needs the same and only one _instance_ to exist in the app, and if there is one react in `docs-site/node_modules` and second react in root `node_modules` then the linked react-datepicker will use its own `react` instance and it blows up the app. This error shows up because of React hooks which are used inside `react-popper@2`. There might be several ways to fix this inconvenience:

1. remove the `react` and `react-dom` from the root package.json devDependencies, during development it will use React from **docs-site**, I assume the react dependencies are there for tests, this could be ideally solved by wrapping the tests in its own directory with its own `node_modules` or set the test environment to resolve modules from **docs-site** node_modules directory,
2. use yarn workspaces and keep all react versions in all projects at the same version, so yarn workspaces will hoist single react module into root node_modules -- there will be no two versions of it. This solution is fragile, as the maintainer must not allow different versions of react to pop into the project.


docs-site/package.json
```diff
@@ -4,9 +4,9 @@
   "private": true,
   "dependencies": {
     "highlight.js": "^10.5.0",
-    "react": "^17.0.1",
+    "react": "^16.9.0",
     "react-app-rewired": "^2.1.8",
-    "react-dom": "^17.0.1",
+    "react-dom": "^16.9.0",
```
czabaj, line 9: Downgraded to correlate with react-datepicker peerDependency range.

docs-site/src/examples/configurePopper.js
```diff
@@ -3,20 +3,22 @@
...
+        {
+          offset: {
+            enabled: true,
+            offset: "5px, 10px",
+          },
+          preventOverflow: {
+            enabled: true,
+            escapeWithReference: false,
+            boundariesElement: "viewport",
+          },
```
czabaj, line 19: Wrapped into an array

src/index.jsx
```diff
@@ -184,7 +184,7 @@ export default class DatePicker extends React.Component {
     placeholderText: PropTypes.string,
     popperContainer: PropTypes.func,
     popperClassName: PropTypes.string, // <PopperComponent/> props
-    popperModifiers: PropTypes.object, // <PopperComponent/> props
+    popperModifiers: PropTypes.arrayOf(PropTypes.object), // <PopperComponent/> props
```
czabaj, line 187: Changed prop-types of `react-datepicker`

src/popper_component.jsx
```diff
@@ -77,14 +79,14 @@ export default class PopperComponent extends React.Component {
...
-    if (portalId && !hidePopper) {
-      popper = <Portal portalId={portalId}>{popper}</Portal>;
+      if (this.props.popperContainer) {
+        popper = (
+          <this.props.popperContainer>{popper}</this.props.popperContainer>
+        );
+      }
+      if (portalId) {
+        popper = <Portal portalId={portalId}>{popper}</Portal>;
+      }
```
czabaj, line 89: I moved all `popper` related tasks into `if (!hidePopper) {` block. Previously, it was separate ifs where some did check `hidePopper` and some even did not take care.

src/popper_component.jsx
```diff
@@ -77,14 +79,14 @@ export default class PopperComponent extends React.Component {
...
-
-    if (this.props.popperContainer) {
-      popper = React.createElement(this.props.popperContainer, {}, popper);
-    }
-
-    if (portalId && !hidePopper) {
-      popper = <Portal portalId={portalId}>{popper}</Portal>;
+      if (this.props.popperContainer) {
+        popper = (
+          <this.props.popperContainer>{popper}</this.props.popperContainer>
```
czabaj, line 84: Changed the `React.createElement` low level call to JSX syntax

src/popper_component.jsx
```diff
@@ -93,7 +95,7 @@ export default class PopperComponent extends React.Component {
     );
 
     return (
-      <Manager className="react-datepicker-manager">
```
czabaj, line 96: Manager only uses `children`, passing `className` is redundant https://github.com/popperjs/react-popper/blob/master/src/Manager.js

src/popper_component.jsx
```diff
@@ -1,25 +1,34 @@
...
 import React from "react";
 import PropTypes from "prop-types";
-import { Manager, Reference, Popper, placements } from "react-popper";
+import { Manager, Reference, Popper } from "react-popper";
+import { placements } from "@popperjs/core/lib";
 import TabLoop from "./tab_loop";
 import Portal from "./portal";
 
 export const popperPlacementPositions = placements;
+const datepickerTriangleSize = 8;
```
czabaj, line 10: Popper2 warns when popper is displaced with `margin`, this constant corresponds to `$datepicker__triangle-size` property, as stated in [migration guide](https://popper.js.org/docs/v2/migration-guide/#4-remove-all-css-margins), margins shall be replaced with offset modifier.

src/popper_component.jsx
```diff
@@ -59,7 +68,22 @@ export default class PopperComponent extends React.Component {
...
+                  const distance =
+                    placement.startsWith("left") ||
+                    placement.startsWith("right")
+                      ? datepickerTriangleSize
+                      : datepickerTriangleSize + 2;
+                  return [skidding, distance];
+                },
+              },
+            },
+          ].concat(popperModifiers || [])}
```
czabaj, line 86: Setting the offset modifier here, this place is better than setting it in `defaultProps` as it will be appended at all times, even though the user could override the modifiers as stated here https://popper.js.org/docs/v2/faq/#how-do-i-set-modifier-defaults-and-allow-them-to-be-overridden

src/popper_component.jsx
```diff
@@ -59,7 +68,22 @@ export default class PopperComponent extends React.Component {
...
         <Popper
-          modifiers={popperModifiers}
+          modifiers={[
+            {
+              name: "offset",
+              options: {
+                offset: ({ placement }) => {
+                  const skidding = 0;
+                  const distance =
+                    placement.startsWith("left") ||
```
czabaj, line 78: @martijnrusschen ups, I used String#startsWith which is not compatible with IE10, do you use automatic polyfilling in the build process, or should I stick with IE10 compatible features?

src/stylesheets/datepicker.scss
```diff
@@ -42,8 +43,6 @@
   z-index: 1;
 
   &[data-placement^="bottom"] {
-    margin-top: $datepicker__triangle-size + 2px;
```
czabaj, line 45: removed the CSS margins.

czabaj:
I would like to try an [arrow modifier](https://popper.js.org/docs/v2/modifiers/arrow/), have you any experience with that? Any thoughts?

docs-site/src/examples/configurePopper.js
```diff
@@ -3,20 +3,30 @@
...
+        {
+          name: "flip",
+          options: {
+            allowedAutoPlacements: ["top", "bottom"],
+            rootBoundary: "viewport",
+          },
         },
-        preventOverflow: {
-          enabled: true,
-          escapeWithReference: false,
```
czabaj, line 16: The old `escapeWithReference` option is replaced with a flip modifier.

martijnrusschen (approved):
Thanks for getting this done, looks good!

src/popper_component.jsx
martijnrusschen: This seems fine to me.

martijnrusschen:
@czabaj It looks like this broke the documentation site. See https://github.com/Hacker0x01/react-datepicker/pull/2653

Issue #2572, opened 2020-12-04
Week numbers not applied correctly


akoel:
If "showWeekNumbers" prop is set to "true" the first week of 2020 should be the week of Sunday, Jan. 3. 2021. React datepicker says this is the 2nd week of 2021.

2020 should have 53 weeks, where the week ending on Jan 2, 2021 is the 53rd week. See: https://www.epochconverter.com/weeks/2021 as a reference for (official) weeknumbers. React datepicker seems to go only to 52 weeks.

tonila:
I noticed this problem also.

stephane-r:
Same for me.

But, i think the problem is with **date-fns**. This lib is used on **date_utils.js** file, and i've try this : 

`console.log(dfgetWeek(new Date())); // 2, it's wrong`

0xIvan:
Has this fix been released yet? When I go to the examples in the site and change locale to en-US and showWeekNumbers to true, it shows that week 1 starts on January 10th.

<img width="896" alt="Captura de pantalla 2021-07-06 a las 11 36 10" src="https://user-images.githubusercontent.com/83612730/124578410-7a8fec80-de4e-11eb-9831-93c46d7e1ae4.png">

stebogit:
I noticed though that, for some reason, the locale value changes the week result.
Also noticed that the code still uses a locale value [here](https://github.com/Hacker0x01/react-datepicker/blob/master/src/date_utils.js#L213), although it sohuld not matter at all since `date-fns/getISOWeek` only accepts [one date argument](https://date-fns.org/v2.23.0/docs/getISOWeek#syntax)

![Screen Shot 2021-07-23 at 1 04 35 PM](https://user-images.githubusercontent.com/12717225/126836200-e8d5fc1f-edf7-4669-aac1-60805ccfd241.png)

![Screen Shot 2021-07-23 at 1 04 47 PM](https://user-images.githubusercontent.com/12717225/126836202-4a495b24-1630-4c45-a259-5b368b226713.png)

`2021-07-23` should be W29 (source [wikipedia](https://en.wikipedia.org/wiki/ISO_week_date))
![Screen Shot 2021-07-23 at 1 05 40 PM](https://user-images.githubusercontent.com/12717225/126836205-404a4247-cad7-401e-98da-2ccf096423c6.png)

PR #2635, opened 2021-01-07
Fixing wrong weekNumber by using getIsoWeek

stephane-r:
The current `showWeekNumbers` return wrong result because (from `getWeek` date-fns function).

Example : 

`console.log(dfgetWeek(new Date())); // 2, it's wrong, want 1`
`console.log(dfgetWeek(new Date('2020-01-01'))); // 1, it's wrong, want 53`

We need to use `getISOWeek` function for return iso result : 

`console.log(getISOWeek(new Date())); // 1, good`
`console.log(getISOWeek(new Date('2020-01-01'))); // 53, good`

From #2572

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -18,7 +18,6 @@ import getMinutes from "date-fns/getMinutes";
 import getHours from "date-fns/getHours";
 import getDay from "date-fns/getDay";
 import getDate from "date-fns/getDate";
-import dfgetWeek from "date-fns/getWeek";
 import getISOWeek from "date-fns/getISOWeek";
 import getMonth from "date-fns/getMonth";
 import getQuarter from "date-fns/getQuarter";
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -19,6 +19,7 @@ import getHours from "date-fns/getHours";
 import getDay from "date-fns/getDay";
 import getDate from "date-fns/getDate";
 import dfgetWeek from "date-fns/getWeek";
+import getISOWeek from "date-fns/getISOWeek";
 import getMonth from "date-fns/getMonth";
 import getQuarter from "date-fns/getQuarter";
 import getYear from "date-fns/getYear";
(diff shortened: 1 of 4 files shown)


src/date_utils.js
```diff
@@ -19,6 +19,7 @@ import getHours from "date-fns/getHours";
 import getDay from "date-fns/getDay";
 import getDate from "date-fns/getDate";
 import dfgetWeek from "date-fns/getWeek";
```
martijnrusschen, line 21: Is this still being used?

martijnrusschen:
Can you add a test?

stephane-r:
> Can you add a test?

Hey @martijnrusschen , sorry for the delay :) i've add 3 tests for get the correct week number :+1:

estherleah:
I think that this fix is actually causing an issue for me. I am working with a custom date locale with `weekStartsOn: 0` and `firstWeekContainsDate: 4` so using `getISOWeek` doesn't work correctly as it doesn't take into account my locale or any specific options passed to it (see https://date-fns.org/v2.19.0/docs/getISOWeek vs https://date-fns.org/v2.19.0/docs/getWeek).
I think a potential fix would be passing an ISO options `{weekStartsOn: 1, firstWeekContainsDate: 4}` to `dfgetWeek` if no locale is given.

abenitesDC:
hello guys, is this change being release anytime soon? Do you need any help? Thanks!

rafchy:
Hey,
I believe the merged fix does not take into consideration locales. This results in wrong week numbers when `showWeekNumbers` is true.

To determine week numbers currently, the following function is used

```
export function getWeek(date, locale) {
  let localeObj =
    (locale && getLocaleObject(locale)) ||
    (getDefaultLocale() && getLocaleObject(getDefaultLocale()));z
  return getISOWeek(date, localeObj ? { locale: localeObj } : null);
```

According to [date-fns docs](https://date-fns.org/v2.0.1/docs/getISOWeek), getISOWeek does not take locale as 2nd param.

I believe  [https://date-fns.org/v2.22.1/docs/getWeek](getWeek) should be used instead because it does allow to pass locale config object.

I was able to find in date-fns docs only one version which does support passing locale to `getISOWeek`.  [2.0.0.alpha-8](https://date-fns.org/v2.0.0-alpha.8/docs/getISOWeekl).

[An Example for locale week numbers](https://codesandbox.io/s/vigorous-night-tjre6?file=/src/App.js)

Let me know what you think and I can take the time to submit a pr.

Issue #2647, opened 2021-01-10
Docs site broken


martijnrusschen:
**Describe the bug**
```
react.development.js:1465 Uncaught Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
```

**To Reproduce**

1. check out the latest master branch
2. run `yarn; yarn build-dev; yarn start`
3. See the error in your console

Is anyone able to troubleshoot what's going wrong? I don't have much time to investigate this currently. I suspect this started to happen after a dependency upgrade. Any help would be appreciated.

tu4mo:
Two different versions of React is imported, likely due to https://github.com/Hacker0x01/react-datepicker/commit/8487dd3ba1e5e14d76a74d003d481edcc9b78342.

martijnrusschen:
@tu4mo You're right, I've put up a fix to revert that change, doesn't look like we to require need react at all there.

PR #2653, opened 2021-01-11
Remove wrong react dependency to fix documentation

martijnrusschen:
Fixes https://github.com/Hacker0x01/react-datepicker/issues/2647

Change introduced in https://github.com/Hacker0x01/react-datepicker/pull/2626, but seems to break the documentation site. Reverting change to make the docs site work again.


martijnrusschen:
This fixes the docs site, but breaks the test suite. Looking for a better solution.

czabaj:
I mentioned the problem in [a comment here](https://github.com/Hacker0x01/react-datepicker/pull/2626#issue-548299565).

The problem was always present but reveals after the upgrade of react-popper, for it uses React hooks under the hoods which relies strictly on a single React instance.

This shall cause no problem in production, for devDependencies are not installed there, but breaks the app for local development.

The best solution is to separate tests into their own directory with their own peer dependencies.

See the linked comment for more info. Also, be aware that the upgrade of react-popper shall be a breaking change of react-datepicker, for propTypes is changed backward incompatibly and the positioning via `offset` modifier might break some customers logic.

martijnrusschen:
@czabaj You're right, I overlooked the comment and didn't take the consequences into account. It's currently breaking our reactdatepicker.com site which is used for public documentation and examples. 

I'm going to be reverting the PR to mitigate the broken docs site for now, but I do agree we need to look into moving the tests out to get rid of the double React dependency. Once this is fixed, I'll release the upgrade again in a major version upgrade as the breaking change shouldn't happen in a minor upgrade.

martijnrusschen:
Closing this PR as this isn't the right approach.

czabaj:
@martijnrusschen hey 👋 

should I re-open the pooper upgrade and draft a solution with the test folder?

PR #2686, opened 2021-01-21
[WIP] Upgrade react popper

czabaj:
This follows #2626 and also tries to isolate test dependencies to fix docs-site, see #2653


czabaj:
@martijnrusschen this is still a work in progress, but it yields the first working solution and I would like to discuss more possible approaches to fix the "two versions of React" issue

- currently, I only deleted `react` and `react-dom` dependencies from root `package.json`, this fixes the "two versions of React" problem but breaks the tests, so I set up the Webpack configuration for tests to resolve dependencies _also_ from `docs-site/node_modules` and I force installed dependencies in `docs-site/` dir via npm scripts. This has the following pros and cons:
    - ✅  there is only one React instance in the docs-site,
    - ✅  tests sources missing dependencies from the docs-site so it is necessary to run `yarn install` in the root and in the `docs-site/` which complies with current workflow,
    - ✅  there are no redundant dependencies installed, which safes storage,
    - ⚠️ the `yarn-install` yields many warnings, for some `devDependencies` relies on `react` and `react-dom`,
    - ⚠️ there is non obvious dependency of `tests/` on `docs-site/node_modules` hidden in `karma.conf.js`
- another approach to this could be to establish a private npm project in `test/`directory and link the `react-datepicker` to it the same way, we link it to the `docs-site/` this will clearly separate the tests from the `react-datepicker` but is more difficult to configure and one must not forget to link the `react-datepicker` before the tests.
    -  ✅  there is only one React instance in the docs-site,
    -  ✅  test/ directory has its own `package.json` it is clearly visible what dependencies belong to the test tooling and what is the build tooling for root `react-datepicker`,
    -  ✅  `yarn install` yields errors only for missing `peerDependencies` which are missing intentionally,
    - ⚠️ the test/ dir creates some redundant duplicates of dependencies in node_modules, thus this is slightly inefficient during the development
    - ⚠️ I tried set the repo up for this second case, but it is not simple and it will introduce major changes to the repository structure

karma.conf.js
```diff
@@ -26,44 +26,48 @@ module.exports = function(config) {
...
-        })
+          "process.env.NODE_ENV": JSON.stringify("test"),
+        }),
       ],
       resolve: {
-        extensions: [".jsx", ".js"]
+        extensions: [".jsx", ".js"],
+        modules: [
+          "node_modules",
+          path.join(__dirname, "docs-site/node_modules"),
```
czabaj, line 50: This tells to the Webpack "Hey, check the node_modules directory all the way down to the file system as you normally would, and, if you do not find what you want, look into path.join(__dirname, "docs-site/node_modules")"

package.json
```diff
@@ -123,9 +121,9 @@
     "precommit": "lint-staged",
     "sass-lint": "sass-lint --config .sass-lint.yml 'src/stylesheets/*.scss'",
     "lint": "run-p eslint flow sass-lint",
-    "start": "yarn --cwd docs-site install && yarn --cwd docs-site start",
-    "test": "cross-env NODE_ENV=test karma start karma.conf.js --single-run",
-    "test:watch": "cross-env NODE_ENV=test karma start karma.conf.js --watch",
+    "start": "yarn --cwd docs-site && yarn --cwd docs-site start",
+    "test": "yarn --cwd docs-site && cross-env NODE_ENV=test karma start karma.conf.js --single-run",
+    "test:watch": "yarn --cwd docs-site && cross-env NODE_ENV=test karma start karma.conf.js --watch",
```
czabaj, line 126: The dependencies in docs-site/ must be installed prior to the tests, so I borrowed a command from `start` but slightly shrank (`yarn` is equivalent to `yarn install`)

martijnrusschen:
@czabaj what's the easiest way forward? I'd like to get this shipped so we can make a breaking release with some new functionality. Any preference on how to get this over the finish line?

czabaj:
Give me some more time, I will write a proposal tomorrow.

czabaj:
Well, back to the problem again.

To sum it up, the problem originates from two instances of React in the application. This causes the peerDependencies being installed in the `react-datepicker` for the sake of the tests.

NPM does not install peerDependencies nor devDependencies of modules, but in your repository, you list peerDependencies in devDependencies and install all, then you link the root project to the docs sub-repository and 💥 BOOM 💥 .

Docs sub-repository has its own version of React, and Node.js module resolution mechanism works against us, when you load anything from the docs repo, it uses its own, single React instance, but when you import `react-datepicker` the Node.js resolves `react` to its private `node_modules`, which introduces second React version.

The cleanest way to handle this, is separate tests to its own sub-folder with its own `node_modules` and keep the root `node_modules` free of React and all other peer dependencies. This need quite a lot of changes, especially in the CI pipeline configuration.

The simplest possible solution I can think of is to adjust the docs sub-repo to always use it's own react with Webpack alias, but you are using create-react-app and this would require eject.

There might be some more options, but I can't see any better right now.

chalbert:
Hi! I've created the #2931 PR that continues from where this one left off. It fixes the docs problem by setting an alias for react, simplifies the setup and fixes some visual regression. Thanks for your work!

martijnrusschen:
Thanks @chalbert, let me take a look and get back to you.

martijnrusschen:
I merged in https://github.com/Hacker0x01/react-datepicker/pull/2931. This is a great achievement, and I want to thank both @czabaj who started this journey as well as @chalbert by taking the final step and bring it over the finish line. This is a team effort and I can't thank you enough for getting this done!

chalbert:
You're very welcome, thanks for reviewing it so quickly!

martijnrusschen:
Follow up was merged in https://github.com/Hacker0x01/react-datepicker/pull/2931

PR #2687, opened 2021-01-22
multi-language get es format accpet in array dateformat

kathirpandian1993:
i'm using locale nl-NL and dateFormat "array", while typing the date in english it returns null only. i need locale support for es and nl so i have copy paste the dateFormat "string" code for supporting es

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -88,8 +88,27 @@ export function parseDate(value, dateFormat, locale, strictParsing) {
       if (isValid(tryParseDate) && strictParsingValueMatch) {
         parsedDate = tryParseDate;
       }
+      if (!isValid(parsedDate)){
+        df = df
+        .match(longFormattingTokensRegExp)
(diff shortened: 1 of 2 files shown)


src/date_utils.js
```diff
@@ -88,8 +88,27 @@ export function parseDate(value, dateFormat, locale, strictParsing) {
       if (isValid(tryParseDate) && strictParsingValueMatch) {
         parsedDate = tryParseDate;
       }
+      if (!isValid(parsedDate)){
+        df = df
+        .match(longFormattingTokensRegExp)
+        .map(function(substring) {
+          var firstCharacter = substring[0];
+          if (firstCharacter === "p" || firstCharacter === "P") {
```
martijnrusschen, line 96: Not completely following what's happening here. Can you add a test case to cover your new behavior without adjusting the current implementations?

martijnrusschen (changes requested):

Issue #2497, opened 2020-10-26
Using a Function component as a customInput shows error


xdumaine:
**Describe the bug**

When using a react Function Component (`FC`) as the `customInput` to a `DatePicker` causes react to log an error `Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

**To Reproduce**

1. Create a function component
2. Render a DatePicker with that function component as the `customInput` value
3. View console

**Expected behavior**
No Errors are shown. 

**Observed behavior**
Using the function component as a custom input seems to work fine, but the react error is shown. It's possible some different behavior would be broken.

sashamorozov:
@xdumaine I faced the same error and found the solution in this comment https://github.com/Hacker0x01/react-datepicker/issues/2397#issuecomment-707048538

YanaLugina:
I fixed this bug like this:
// when create custom input added ref in function params
`    const ExampleCustomInput = React.forwardRef(({ value, onClick }, ref) => (
        <button ref={ref} className={style.timeInput} onClick={onClick}>
            <span>c</span>
            {value}
        </button>
    ));
    const ref = React.createRef();`

// in DatePicker i passed ref
`
<DatePicker
                selected={startTime}
                ...
                locale={ru}
                timeCaption='Время'
                dateFormat='HH:mm'
                customInput={<ExampleCustomInput ref={ref} />}
                onChange={(data) => {
                    setStartTime(data);
                }}
            />`

robations:
Does the documentation need to change?

https://reactdatepicker.com/#example-custom-input
https://github.com/Hacker0x01/react-datepicker/blob/master/docs-site/src/examples/customInput.js#L1

I can make a PR if this contribution is appropriate.

YanaLugina:
> Does the documentation need to change?
> 
> https://reactdatepicker.com/#example-custom-input
> https://github.com/Hacker0x01/react-datepicker/blob/master/docs-site/src/examples/customInput.js#L1
> 
> I can make a PR if this contribution is appropriate.

Yes, please!

robations:
@xdumaine `forwardRef()` has been added to the docs so you might be able to close this issue?

PR #2701, opened 2021-01-26
Fix doc-site example for custom input

robations:
Hello! I changed the example code for the doc site, which should answer #2497.

This was based on the example provided by @YanaLugina, and was tested working in my own project.

I read through the contributing guidelines etc but let me know if there's anything I've missed or should add.

Rob.

### DIFF
--- a/docs-site/src/components/Example/index.js
+++ b/docs-site/src/components/Example/index.js
@@ -44,7 +44,7 @@ export default class CodeExampleComponent extends React.Component {
               ...DateFNS,
               range,
               fi,
-              forwardRef
+              React
             }}
             theme={prismGitHubTheme}
           >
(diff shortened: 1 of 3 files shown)


robations:
> `React` is not defined

Will have to take another look. None of these example seem to use imports, I assumed React was in context for jsx.

YanaLugina:
> Will have to take another look. None of these examples seem to use imports, I assumed React was in context for jsx.

yes, React need imported in the context above, in the same place as the DatePicker itself. Or, if used React version >= 17.0.0 need import {forwardRef} from ‘react’

drews256:
I'll bump this PR. It should definitely be in the docs.

robations:
Thanks for the reminder @drews256, will try to take a look at the weekend. I thought I'd already submitted a fix to be honest #brainProblems

drews256:
I actually wanna bring this backup. I did use forwardRef like the changes here suggest and while it eliminated the ref error I was unable to get any sort of onClick callback to fire on my customInput. Have you guys seen this issue? If you haven't seen it, I may try to reproduce it in a small example and share it. After switching to a class-based component everything worked perfectly.

robations:
@drews256 IIRC, adding forwardRef fixed any issues for me

Issue #2720, opened 2021-02-10
filterTime does not autocorrect input value when invalid time value is typed


comatory:
**Describe the bug**
If I use `filterTime` prop to validate enabled times, it works fine for time picker UI. The times that are supposed to be disabled are not selectable. However when typing invalid date and time into input, it will **not** auto-correct the value.
When using `filterDate` the invalid date is auto-corrected to nearest valid date when typing it into the input.
I am not sure if this is a bug or intentional design. Since `filterDate` corrects value of the input and disables the dates in the picker, I assumed the auto-correcting input was a feature.

**To Reproduce**

1. Go to [https://codesandbox.io/s/react-datepicker-values-not-autocorrected-ek30v](https://codesandbox.io/s/react-datepicker-values-not-autocorrected-ek30v)
2. Click on second input field labelled "Not working".
3. Type in `02/01/2021, 08:00 AM`
4. Value in input field is left as it is even though that date should be selected.

**Expected behavior**
I would expect that it would autocorrect value to `02/01/2021, 08:30 AM` since that value is allowed. I am expecting this because when using `filterDate` method, it behaves in such a way. I've also provided example of the `filterDate` behaviour in the example.

**Desktop (please complete the following information):**

- OS: macOS
- Browser Chrome
- Version 88

**Additional context**
Package version `3.4.1`

comatory:
I have opened a PR for this. However I'm having hard time understanding the code, specifically the situation where the date is corrected (when it does not fall in the range of specified by `maxDate` and `minDate`). I have tried adding my code to places where I though it should work but to no avail. If anyone could help me with this, let's bring the discussion to the PR itself.

comatory:
Bump.

supersnager:
The same bug exists in v4.6.0

Desintegrator:
same

PR #2723, opened 2021-02-11
fix: attempt to validate dates when minTime and maxTime props are

comatory:
specified

I'm trying to address #2720 (https://github.com/Hacker0x01/react-datepicker/issues/2720). The value of the input should be reverted to date AND time that is within allowed constraints. This already happens for `maxDate` and `minDate` properties so it should be consistent for times as well.

However I am not able to pinpoint in the code where this normalization
occurs. When user presses enter or blurs the input, the value should
revert. I've tried adding validation code to parts where I thought this
occured but I am having no such luck.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -37,6 +37,8 @@ import {
   yearsDisabledBefore,
   getEffectiveMinDate,
   getEffectiveMaxDate,
+  getEffectiveMinDateTime,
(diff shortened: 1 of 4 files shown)


martijnrusschen (changes requested):
If anyone is able to help that would be appreciated. Seems like a valid issue.

PR #2767, opened 2021-02-25
Update datepicker.md

RoyBkker:
Remove utcOffset, as it is deprecated.


martijnrusschen:
This is an automated file, unless the functionlaity gets removed from the codebase this change won't do anything.

Issue #2661, opened 2021-01-12
Support for passing option to date-fns (useAdditionalWeekYearTokens)


rafchy:
Hello,

Starting with date-fnsv2, `format` and `parse` use [Unicode tokens](https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table). More information can be found  [on date-fns docs](https://github.com/date-fns/date-fns/blob/master/docs/unicodeTokens.md)

The problem we are facing is that we can't use `dateFormat="ww/YYYY"`. As a workaround, we replace YYYY with YYY.

Currently, react-datepicker does not support passing an option for allowing useAdditionalWeekYearTokens. If we were able to pass this additional option to date-fns the problem would not exist.

I've created a codesandbox showing the problem, to reproduce uncomment line 17.

[link](https://codesandbox.io/s/priceless-sammet-ouhge?file=/src/App.js)

piyushkathayat:
@martijnrusschen

mykolap:
See 
https://unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table
Year in “Week of Year” based calendars in which the year transition occurs on a week boundary; may differ from calendar year ‘y’ near a year transition. This numeric year designation is used in conjunction with pattern character ‘w’ in the ISO year-week calendar as defined by ISO 8601, but can be used in non-Gregorian based calendar systems where week date processing is desired. The field length is interpreted in the same was as for ‘y’; that is, “yy” specifies use of the two low-order year digits, while any other field length specifies a minimum number of digits to display.

PR #2768, opened 2021-02-26
Implement week picker component

### LINKED ISSUE #2033: Can we have a week picker ?
I see month picker , year picker but no week picker :) .

TiesWestendorp:
fixes #2033

Introduces the `showWeekPicker` setting, which allows picking a week. Internally, the date representing the week is the first day of that week. It's important that the `dateFormat` prop matches the `locale` in the sense that they have the same starting day of the week. In the same vein, the `selected` prop should refer to a date that's at the beginning of a week.

Peek 2021-03-01 16-15

**Question 1**: should left and right arrows navigate the months, rather than weeks? **Answer:** no.
**Question 2**: should the week component be considered selected/keyboard-selected, and the styling be derived from that, or should the days be considered selected/keyboard-selected like they currently are?


TiesWestendorp:
@martijnrusschen or someone else, can you take a look at this PR, and give your take on the questions I posed above?

tlehtimaki:
Q2: I think the selecting of all dates in the week seems logical, but then I think you should not be able to individually select the dates.
Q1: For how its implemented now the up/down arrows are logical for browsing the week. The browsing of month suddenly with left/right arrow might be confusing for the user, if user is used to browsing days with those keys.

TiesWestendorp:
> Q2: I think the selecting of all dates in the week seems logical, but then I think you should not be able to individually select the dates.

My question was actually more geared towards the internal state; each day component has a 'selected' and 'keyboard-selected' state, which I reuse for the week picker here. Another way would be to not consider these days selected or keyboard-selected in this case, but the week component that wraps these days. There's a bit of a trade-off here: the former option borrows heavily from the existing styling, meaning it doesn't introduce much complexity in the stylesheets, but is potentially less flexible. As to your comment: clicking on a particular date will currently select the corresponding date - do you mean to say that you don't agree with this? My issue with that is that when the week numbers are not present you won't have a meaningful thing to click on. A takeaway from this may be that the week numbers should always be shown when the week picker is used, but that is also currently not the case.

> Q1: For how its implemented now the up/down arrows are logical for browsing the week. The browsing of month suddenly with left/right arrow might be confusing for the user, if user is used to browsing days with those keys.

I agree, but wanted to lay the options on the table. :+1: Currently left/right behave the same as up/down in the week picker.

tlehtimaki:
I understood that the concept would be to select just weeks, so I thought there would be no need to select a single date, but here that single date would convert in to the week that the date belongs to.

TiesWestendorp:
> I understood that the concept would be to select just weeks, so I thought there would be no need to select a single date, but here that single date would convert in to the week that the date belongs to.

Correct, just having the week number select the first date of a week was already possible by using a callback on the week number. What this PR introduces is the additional logic required to properly display the state, changed keyboard navigation to match and a fallback that ensures the first day of a week is always used.

TiesWestendorp:
Could some one take a look at why the specs are failing? I'm not really sure what's going wrong.

martijnrusschen:
Can you try to pull the latest master in? I think the issue was fixed
earlier.

TiesWestendorp:
The failing spec seems to be fixed by rebasing, but I'm still getting errors during linting...

martijnrusschen:
It looks like something happened with the rebase, seeing some changes in the PR that already got merged. Regarding the lint errors, it seems they're related to the changes in this PR.

martijnrusschen:
Checking in, can you merge with the latest master so we can release this functionality?

TiesWestendorp:
> Checking in, can you merge with the latest master so we can release this functionality?

Did so just now, that at least cleans up the git history. :+1: I'm still not sure about the lint errors though. :shrug:

mykolap:
And proper format string for year/week will also be supported for this peeker? Currently we are forced to use literals "ww/YYY", with 3 letters Y, instead of YY or YYYY, because date-fns rejects ww/YYYY or ww/YY. And week should be used only with big Y, not small y. See https://github.com/Hacker0x01/react-datepicker/issues/2661. Week peeker without support of WeekYear tokens in date-fns will be not complete.

TiesWestendorp:
> And proper format string for year/week will also be supported for this peeker? Currently we are forced to use literals "ww/YYY", with 3 letters Y, instead of YY or YYYY, because date-fns rejects ww/YYYY or ww/YY. And week should be used only with big Y, not small y. See #2661. Week peeker without support of WeekYear tokens in date-fns will be not complete.

I'm not really familiar with the problem you're describing. The only thing that's important for this PR, is that `date-fns` is able to parse the the given format, and that in doing so it returns a date that corresponds to the first calendar cell in a given week. In the given example this means I've used the format `I/R`.

mykolap:
To show start of week correctly - you should pass locale to dateFns, after that if locale German - you will get Monday as a start, if US - Sunday, but issue is with year, because year for week and calendar year - are different, and only hack with 3 letters Y is working when we want to handle year correcly... And it's very ugly hack...

TiesWestendorp:
> To show start of week correctly - you should pass locale to dateFns, after that if locale German - you will get Monday as a start, if US - Sunday, but issue is with year, because year for week and calendar year - are different, and only hack with 3 letters Y is working when we want to handle year correcly... And it's very ugly hack...

I would say that this seems like a separate issue from what's being done in this PR. _Given_ that `format` and `parse` work correctly, the changes from this PR work correctly.

TiesWestendorp:
@martijnrusschen I've fixed the lint errors. Could you take a look at this PR?

src/day.jsx
```diff
@@ -259,10 +270,17 @@ export default class Day extends React.Component {
   getTabIndex = (selected, preSelection) => {
     const selectedDay = selected || this.props.selected;
     const preSelectionDay = preSelection || this.props.preSelection;
-
+    // TODO: use isStartOfWeek
```
martijnrusschen, line 273: Is this still something you want to change here?

TiesWestendorp: I think I just forgot to remove a comment there.

TiesWestendorp:
@martijnrusschen Could you still have a look at this PR?

martijnrusschen:
This looks good! Can you resolve the merge conflict so I can merge it in? I'm preping a major release for this week and would love to get this in.

TiesWestendorp:
:+1: Done.

martijnrusschen:
I ran the tests, and it looks like there's some things breaking. Can you take a look?

TiesWestendorp:
Should be fixed now.

test/week_picker_test.js
```diff
@@ -0,0 +1,53 @@
+import React from "react";
```
martijnrusschen, line 1: Thinking a bit on code coverage.. Might be good to add some tests that checks the behavior of showing/hiding the weekpicker as well. Making sure the weekpicker isn't shown when the prop isn't passed in and shown when it is passed in.

martijnrusschen: Codecov added some comments on coverage as well, would be great to improve the coverage to make sure we don't break this in the future.

martijnrusschen (changes requested):
Let's improve the test coverage to make sure we won't break this in the future.

georgecosmin98:
Hello!

Any news about week-picker? When can we expect a release?

johnnyxbell:
Any news on when we can get this merged?

sffirm:
Why is it taking so long?

manzoorwanijk:
Any update on this?

nacimgoura:
When will this feature be merged?

arikm66:
I really need this feature :)  When will it be merged?

martijnrusschen:
Perhaps someone can take a look and bring this to the finish line.

Seaviello:
Hi,
we are also currently interested in that feature. Is there something that outsider can do to help you out with making this happen or right now it's mostly in the reviewers hand?

akashraj9828:
Any update on this... really need this feature. Thanks in advance.

akashraj9828:
Not sure if anyone needs this anymore but I was able to implement the week picker with the existing library... 
Demo: 
https://codesandbox.io/s/react-datepicker-example-forked-89fn7?file=/src/WeekPicker.tsx

ngajhede:
Hey @correahi! Thanks for this. I was wondering how you display the `startDate to endDate` in the input box?

correahi:
> Hey @correahi! Thanks for this. I was wondering how you display the `startDate to endDate` in the input box?

I don't have access to that codebase anymore but i'm pretty sure it's just a span with some text inside, something like `${startOfISOWeek(selectedDate)} to ${endOfISOWeek(selectedDate)}` and `onClick` it opens the datepicker, it's not the default input.

gtaylor44:
Will this be merged into master?

PR #2778, opened 2021-03-03
upgrade react-popper to v2

hasanayan:
react-popper v1 depends on popper.js while the newer versions (v2.*) use @popperjs/core. In our projects we rely on react-popper v2 and currently this version conflict causes double import of those libraries into our bundle.  This pr makes sure a current, non-deprecated version of packages are used and reduces the bundle bloat for people who use @popperjs/core in their projects

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,11 +110,12 @@
     "react-dom": "^16.9.0 || ^17"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


martijnrusschen:
This upgrade involves a bit more work. See https://github.com/Hacker0x01/react-datepicker/pull/2686

PR #2815, opened 2021-03-15
Bugfix/make mindate and maxdate selectable via keyboard

### LINKED ISSUE #2429: Unable to select minDate using keyboard
**Describe the bug**
When using a `minDate` of e.g. `new Date()`, I'm not able to select the `minDate` using the keyboard. The `minDate` is styled as not disabled, and I can select it using the mouse, but using the keyboard I'm unable to navigate to it.

I _am_ able to navigate to it if I instead set `minDate` to `new Date().setHours(0,0,0,0)`. So there seems to be some sort of mismatch in how days are disabled (styling and click-handler) and how keyboard selection is "disabled" (what days you're allowed to navigate to via keyboard).

I.e. the disabling-mechanism seems to ignore time, while the "allowed keyboard navigation"-mechanism does not. So, the fix here should probably be to also ignore time in the keyboard navigation? 🤔

**To Reproduce**
1. Go to [Inline version](https://reactdatepicker.com/#example-inline-version)
2. Add a `minDate` of `new Date()`:  
```jsx
() => {
  return (
    <DatePicker
      selected={startDate}
      minDate={new Date()}
      inline
    />
  );
};
```
3. Using the keyboard, try navigating to the `minDate`.

**Expected behavior**
I expect both the disabled styling, selecting with mouse, and selecting with keyboard, to all allow me to select the `minDate`.

**Screenshots**

image

In the screenshot the keyboard selection is on the 6th, and I'm not able to arrow-left to the 5th, even though it shows as not disabled and I can select it using the mouse.

- OS: Windows 10
- Browser: Chrome
- Version: 85

relwiwa:
Selecting a minDate or maxDate via keyboard now works.

Furthermore, it is no longer necessary to neutralize the time fragment of dates (hours, minutes, seconds, milliseconds) as this is done internally now when comparing the pre-selected date with min/maxDate.

fixes #2429

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -73,17 +73,6 @@ export function newDate(value) {
   return isValid(d) ? d : null;
 }
 
-export function newDateWithNeutralTime(value) {
-  const copy = newDate(value);
-  if (copy !== null) {
-    copy.setHours(0);
-    copy.setMinutes(0);
-    copy.setSeconds(0);
-    copy.setMilliseconds(0);
-  }
-  return copy;
-}
-
 export function parseDate(value, dateFormat, locale, strictParsing) {
   let parsedDate = null;
   let localeObject = getLocaleObject(locale) || getLocaleObject(getDefaultLocale());
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -73,6 +73,17 @@ export function newDate(value) {
   return isValid(d) ? d : null;
 }
 
+export function newDateWithNeutralTime(value) {
+  const copy = newDate(value);
+  if (copy !== null) {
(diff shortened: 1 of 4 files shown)


martijnrusschen:
I see this is still in draft, are you planning on finishing this up?

relwiwa:
@martijnrusschen I finally came back to this issue and refactored the PR so now it should be ready to merge. I had quite some difficulty reproducing the bug via the tests, but eventually came up with two tests that actually failed when minDate or maxDate are set, but not when minDate and maxDate are set, which was the actual behavior. I think it would be a lot easier if there were end to end tests.

PR #2889, opened 2021-04-13
make the clear button customizable

ilkou:
I added a property named clearButtonClassName which will be added to the clear button classname property, to make it customizable !


martijnrusschen:
Can you add a test to make sure this keeps working in the future?

ilkou:
Here you go !
clearButtonClassName-test

Issue #2706, opened 2021-02-01
Time selection doesn't work from keyboard


vatsalyaagarwal:
**Describe the bug**
Time selection doesn't work on pressing enter

**To Reproduce**
1. run the date picker with witch "showTimeSelect" 
2. tab to the time selection list
3. try to press enter on the time 
 
**Expected behaviour**
Time should get selected on pressing enter

martijnrusschen:
Fixed in #3005

PR #2893, opened 2021-04-15
Select time on enter

oshirok:
Currently, the time selector is not accessible. As described in this issue https://github.com/Hacker0x01/react-datepicker/issues/2706 ,  times cannot be chosen by keyboard. 

So in this pull request I'd like to make the time selector usable from the keyboard by allowing users to select a time by the enter key.

### DIFF
--- a/src/time.jsx
+++ b/src/time.jsx
@@ -13,7 +13,7 @@ import {
   isEqual,
   isTimeInDisabledRange,
   isTimeDisabled,
-  timesToInjectAfter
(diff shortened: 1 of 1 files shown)


martijnrusschen (changes requested):
Can you add some test coverage to make sure this behavior doesn't break in the future?

Abdubek:
this was fixed in #3005, I think you can close this pr and this issue #2706

PR #2931, opened 2021-05-05
Upgrade react-popper to v2.x

chalbert:
This PR updates `react-popper` to its latest version while addressing the visual regressions the update caused. This required special care to the display of the arrow.

There's a lot of noise coming from the linting defined as a precommit hook, sorry about that.

This PR should also fix #2653 by setting an alias for react in the docs, see `docs-site/config-overrides.js`.

Thanks to @czabaj  and his PR #2686 that was the starting point of this PR.


martijnrusschen:
Looks like I had to manually approve to run the workflows. Just started them.

martijnrusschen (approved):
Tested manually as well and all seems to work as expected.

PR #2940, opened 2021-05-07
Prevent class name 'undefined' when clearButtonClassName is not provided

anneFly:
This fixes a most likely unwanted side effect that was introduced in https://github.com/Hacker0x01/react-datepicker/commit/f5402da72c25d2c40c9bd576f136a8ca19a36ffa.

When `clearButtonClassName` is not provided, the current code renders "undefined" as a class name of the clear button. This fix makes sure that the extra class name is only added if it was provided.


martijnrusschen:
Thanks!

PR #2957, opened 2021-05-12
Add : start day 

rohanvachheta:
- here I have added calendarStartDay which will show start weekday related to the passed number 

currently there is no start day flag , only week days starts with Sunday, after this you can pass any day and week day will be start from that onwards


martijnrusschen:
I like the approach, can you add some tests to make sure this behavior works as expected? Also, can you add an example to the docs site to show other users how to use this?

martijnrusschen (changes requested):
Requesting for changes based on previous comment

rohanvachheta:
@martijnrusschen https://github.com/Hacker0x01/react-datepicker/pull/2957/commits/60687f1e0dfc3d42ae64072c93d72921662d4860 here i've added test cases and example for the changes

martijnrusschen (approved):

rohanvachheta:
@tejalkukadiya thanks, I took reference of your code. thanks

PR #2982, opened 2021-05-20
Add monthDate to renderCustomHeader render props function

### LINKED ISSUE #2981: Add monthDate to renderCustomHeader render props function to get access to displayed month
**Is your feature request related to a problem? Please describe.**
When rendering a custom header with more than one months shown, it is necessary to have access to the currently displayed month, so that it can be displayed in the custom header in sync with the months displayed in the calendar. The date and customHeaderCount props aren't enough to achieve this.

**Describe the solution you'd like**
Add monthDate prop to renderCustomHeader render props function

**Describe alternatives you've considered**
This seems the easiest solution

relwiwa:
When rendering a custom header and displaying more than one month, like this, it becomes possible to sync the month displayed in header and in calendar

fixes #2981

### DIFF
--- a/docs-site/src/components/Examples/index.js
+++ b/docs-site/src/components/Examples/index.js
@@ -64,6 +64,7 @@ import CalendarContainer from "../../examples/calendarContainer";
 import RawChange from "../../examples/rawChange";
 import DontCloseOnSelect from "../../examples/dontCloseOnSelect";
 import RenderCustomHeader from "../../examples/renderCustomHeader";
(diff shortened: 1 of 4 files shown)


martijnrusschen:
Can you add some specs to make sure this won't break in the future?

relwiwa:
@martijnrusschen I added a test for the monthDate prop, I hope I got it right because with Dates it can be tricky for tests to work 24/7/365, please double-check

martijnrusschen:
Looking good, thanks!

PR #3045, opened 2021-06-10
Fix start of week not using the registered locale

### LINKED ISSUE #3015: Locale does not change calendar start day
**Describe the bug**

When applying a locale to a datepicker, the `weekStartsOn` value seems to no longer have an effect.

Our datepickers that use the `en-GB` locale for example would have Monday as the first day of the week as of `3.8.0`.

Since upgrading to `4.1.0` the weeks now start on Sunday. (This also happens in `4.0.0`)

The value can be updated manually using the new `calendarStartDay` prop, but I'm assuming it's supposed to still be picked up from the locale provided.

**To Reproduce**

1. Apply a locale with a `weekStartsOn` value that is not `0`.
2. Render datepicker.
3. See week starts on Sunday.

This looks to also be happening in the [locale example in the documentation](https://reactdatepicker.com/#example-locale).

**Expected behavior**

The week starts on the day specified by the locale.

**Screenshots**

`3.8.0`
Datepicker in 3-8-0

`4.1.0`
Datepicker in 4-1-0

**Desktop**

- OS: MacOS 11.4
- Browser: Chrome
- Version: 91.0.4472.77

Although this happens on all browsers checked (Firefox, Safari, Edge)

Amwam:
Fixes #3015

The locale we defaulting to 0, via the default props. This meant that 0 was passed all the way down to the function `getStartOfWeek`. The result of this is that the week will always start on a Sunday, ignoring the locale that is set.

The solution is just to make this undefined, which will then use the correct locale version if its set.
This should be fine as any uses where the number is overridden is still passed down, and those without a locale set will be defaulting to the US, which is 0 (sunday)

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -116,7 +116,7 @@ export default class DatePicker extends React.Component {
       showPopperArrow: true,
       excludeScrollbar: true,
       customTimeInput: null,
-      calendarStartDay: 0,
+      calendarStartDay: undefined,
     };
   }
 
--- a/test/datepicker_test.js
+++ b/test/datepicker_test.js
@@ -1,9 +1,10 @@
 import React from "react";
 import ReactDOM from "react-dom";
 import TestUtils from "react-dom/test-utils";
+import { enUS, enGB } from "date-fns/locale";
 import { mount } from "enzyme";
 import defer from "lodash/defer";
-import DatePicker from "../src/index.jsx";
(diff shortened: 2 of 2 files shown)


Amwam:
I'm not sure why the coverage has gone down in this PR, if you have any pointers happy to try and fix

martijnrusschen:
Thanks for fixing!

PR #3052, opened 2021-06-16
Uses react dom portals for portal version when portalId is provided

humphreyja:
## Changes
* providing `portalId` when used in conjunction with `withPortal` will render the portal version calendar in a ReactDOM portal (using the Portal component).
* Removes confusing logic of inline portals. Inline always renders the calendar. Providing the `withPortal` prop only wrapped the calendar in the portal class.
* Adds documentation for portal versions rendered with ReactDOM portals.
* Removes documentation for InlinePortal which was not used.
* Adds documentation for toggling an Inline calendar.
* Tests inline prop will ignore withPortal prop
* Tests withPortal and portalId will render a portal version with the portalId

### DIFF
--- a/docs-site/src/components/Examples/index.js
+++ b/docs-site/src/components/Examples/index.js
@@ -43,6 +43,7 @@ import OnBlurCallbacks from "../../examples/onBlurCallbacks";
 import ConfigurePopper from "../../examples/configurePopper";
 import Portal from "../../examples/portal";
 import PortalById from "../../examples/portalById";
+import WithPortalById from "../../examples/withPortalById";
 import TabIndex from "../../examples/tabIndex";
 import YearDropdown from "../../examples/yearDropdown";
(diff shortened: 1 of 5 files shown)


martijnrusschen:
Can you confirm this doesn't break any of the existing behaviors or does this require a major release?

humphreyja:
I don't think this requires a major release. Really what this does is give consistency to the 'portalId' attribute. As for the inline prop change, as I mentioned providing the withPortal prop ended up just wrapping it in the portal class (so render the calendar full page but you'd have no way to dismiss the calendar because it's also inline). To me that made no sense in terms of when it would be used and the documentation never actually demonstrated that behavior nor did any tests. That's why I opted to just simplify the code and write the docs and tests to clarify the expected behavior. If you disagree I can add back that behavior and I'll write the tests/docs to support it as well.

PR #3093, opened 2021-07-09
Bugfix - Navigation button icons are hidden in Safari

stephenbelyea:
## In this PR:

* Set explicit 0 width for navigation icon span container to render consistently on Safari

## Notes:

Chrome, Firefox, Edge, and other browsers implicitly sets the navigation button's width to 0 (collapsed) due to the combination of absolute and relative positioning. Safari, on the other hand, calculates a width based on left/right and other values, which leaves the icon outside of the button's hidden overflow. Setting an explicit width of 0 does not change rendering on other browsers, but helps Safari render the icon as expected.

Tested on Chrome, Firefox, Edge, Opera, and Safari.

## Screenshots:

**Original navigation rendering in Safari:**
image

**Updated navigation rendering in Safari:**
image

### DIFF
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -189,6 +189,7 @@
   position: relative;
   top: -1px;
   font-size: 20px;
+  width: 0;
 
   &::before {
     @extend %navigation-chevron;


stephenbelyea:
Hello! My manager found this rendering issue because for some reason he uses Safari. I fixed it in our app, which is using React Datepicker, but figured it may be helpful to offer the fix for other folks as well. Thanks!

martijnrusschen:
Thank you!!

PR #3101, opened 2021-07-20
update datepicker.scss with full width

devnull69:
Set width to 100% inside the component's container, so using it with e.g. the col-* classes of Bootstrap will become possible


devnull69:
Do you have any idea about when this change might be officially released?

Issue #2328, opened 2020-08-14
Include dates for Month picker


inouekun:
**Describe the bug**
Using `includeDates` with `showMonthYearPicker` does worked, but the issue is those not included dates are not showing as `disabled` style.
Upon selecting those `not included` dates doesn't do anything which is supposed to be.

**To Reproduce**
`<DatePicker
            ...
            showMonthYearPicker
            includeDates={[...]}
          />`

**Expected behavior**
`showMonthYearPicker` month list should `disabled` for those not included in `includeDates` array as per following example screenshot

or perhaps adding `includeMonths` with an array of months

**Screenshots**
![Screenshot 2020-08-14 at 11 51 15 AM](https://user-images.githubusercontent.com/27730503/90212089-0e1a9200-de25-11ea-87ad-46865dd41481.png)
<img width="220" alt="Screenshot 2020-08-14 at 3 05 28 PM" src="https://user-images.githubusercontent.com/27730503/90222957-ab82bf80-de3f-11ea-974f-07215fc84021.png">

wanhchai:
got the same issue here

ushev-s:
I have the same issue. Month disabling works only with minDate and maxDate. It would be great to use **includeMonths** for such datepickers

joelugh:
Had same issue. ^ That PR should at least allow month disabling (and quarter disabling) by includeDates and excludeDates. 

The existing includeDates check looks to see no dates in that month are included before disabling it. So for instance, if the includeDates contained only 24 July 2021, then July 2021 would not be disabled.

The caveat is that I think when you click a month it tries to select the first day of the month as the selected date. In this case, it would try to select the 01 July 2021 which is not in includeDates, and so actually selecting the month would fail.

It's not clear to me what the expected behaviour should be - so in the meantime I will reduce includeDates down to the first date of each included month when using showMonthYearPicker.

PR #3112, opened 2021-07-24
Allow Quarter and Month disable by includeDates and excludeDates

joelugh:
Removing minDate || maxDate requirement from Month and Quarter disabled checks to allow disable by includeDates and excludeDates.

Should fix: https://github.com/Hacker0x01/react-datepicker/issues/2328

NOTE:

The existing includeDates check looks to see no dates in that month are included before disabling it. So for instance, if the includeDates contained only 24 July 2021, then July 2021 would not be disabled.

The caveat is that I think when you click a month it tries to select the first day of the month as the selected date. In this case, it would try to select the 01 July 2021 which is not in includeDates, and so actually selecting the month would fail.

It's not clear to me what the expected behaviour should be - but a workaround here is to reduce includeDates down to the first date of each included month when using showMonthYearPicker.

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -428,7 +428,7 @@ export function isMonthDisabled(
   { minDate, maxDate, excludeDates, includeDates, filterDate } = {}
 ) {
   return (
-    isOutOfBounds(month, { minDate, maxDate }) ||
+    ((minDate || maxDate) && isOutOfBounds(month, { minDate, maxDate })) ||
     (excludeDates &&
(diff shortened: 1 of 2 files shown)


martijnrusschen (changes requested):
Can you add some tests to cover the changes you made?

joelugh:
Hi @martijnrusschen - happy to add some tests however as per the note above it was a bit ambiguous what the correct behaviour should be.

Do you agree with the behaviour described above? Happy to clarify further if unclear.

PR #3212, opened 2021-09-23
Added new property 'inlineWithInput'

xkomiks:
In fact, this is an extension of the already existing property `inline` that renders the calendar.

But if you use `inlineWithInput`, then the Input element will also be rendered


xkomiks:
Can you tell me how to improve coverage so that my PR gets approved? Need more tests?

Issue #3227, opened 2021-09-28
Increment react-onclickoutside to 6.12.0 to support shadow-dom


msnyMoodys:
**Is your feature request related to a problem? Please describe.**
react-onclickoutside recently pushed an update to support shadow-dom interactions. This update is working successfully, however when npm publishing using react-datepicker as a dependency it is not pulling in 6.12.0 version, instead it is respecting the yarn-lock and keeping version 6.10.0 which causes the resulting package not to support shadow-dom.

**Describe the solution you'd like**
Update react-onclickoutside to 6.12.0

**Describe alternatives you've considered**
If you want to create an npm package that uses react-datepicker, you will not be able to use the latest version of react-onclickoutside which support shadow dom. 

**Additional context**
Related react-onlickoutside PR for supporting shadow-dom https://github.com/Pomax/react-onclickoutside/pull/363

Related react-datepicker PR with updated dependency https://github.com/Hacker0x01/react-datepicker/pull/3228

msnyMoodys:
PR was merged, thank you!!

I submitted a follow-up issue to publish a new version with the updated dependencies https://github.com/Hacker0x01/react-datepicker/issues/3237

PR #3228, opened 2021-09-28
chore: updated react-onclickoutside support shadow-dom

msnyMoodys:
Related to issue https://github.com/Hacker0x01/react-datepicker/issues/3227

# Is your feature request related to a problem? Please describe.
react-onclickoutside recently pushed an update to support shadow-dom interactions. This update is working successfully, however when npm publishing using react-datepicker as a dependency it is not pulling in 6.12.0 version, instead it is respecting the yarn-lock and keeping version 6.10.0 which causes the resulting package not to support shadow-dom.

# Describe the solution you'd like
Update react-onclickoutside to 6.12.0

# Describe alternatives you've considered
If you want to create an npm package that uses react-datepicker, you will not be able to use the latest version of react-onclickoutside which support shadow dom.

# Additional context
Related react-onlickoutside PR for supporting shadow-dom Pomax/react-onclickoutside#363

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "classnames": "^2.2.6",
     "date-fns": "^2.24.0",
     "prop-types": "^15.7.2",
-    "react-onclickoutside": "^6.10.0",
+    "react-onclickoutside": "^6.12.0",
     "react-popper": "^2.2.5"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


martijnrusschen:
Awesome, thanks for bumping this!

PR #3268, opened 2021-10-14
Prevent opening calendar on click if preventOpenOnFocus is true

NicolaiF:
I was trying to utilize `onKeyDown` in my TimePicker input-only component, to change the time with the arrow-keys. 
However, I ran into trouble as [this part of the code](https://github.com/Hacker0x01/react-datepicker/blob/master/src/index.jsx#L653 ) expected a calendar to exist (or at least this.state.open is true, even though it is not rendered as I have applied `open={false}` and `showTimeInput={false}`

And technically, a mouse-click is also a focus-event, thus it is correct to avoid opening if `prevenOpenOnFocus={true}`.

Please let me know if the change is not wanted, or if you'll need further explanation of how my custom TimePicker component is setup.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -623,7 +623,11 @@ export default class DatePicker extends React.Component {
   };
 
   onInputClick = () => {
-    if (!this.props.disabled && !this.props.readOnly) {
+    if (
+      !this.props.disabled &&
+      !this.props.readOnly &&
+      !this.props.preventOpenOnFocus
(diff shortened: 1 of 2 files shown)


NicolaiF:
@martijnrusschen Do you have any questions regarding this pull request, or can it be merged?
I think the tests are failing as they're compared to a newer commit of master, as the coverage difference is not files I've edited.

PR #3272, opened 2021-10-15
Add time limit feature

kross77:
Time limit feature provides display times in the range


package.json
```diff
@@ -2,7 +2,7 @@
   "author": "HackerOne",
   "name": "react-datepicker",
   "description": "A simple and reusable datepicker component for React",
-  "version": "4.2.1",
+  "version": "4.2.3",
```
martijnrusschen, line 5: No need to change this version number. This will happen automatically when we merge and release the PR.

PR #3308, opened 2021-11-17
Add ability to exclude and include date intervals

fcolacilli:
Added properties to exclude and include dates from one or many intervals.

This is intended to make it easier to define these ranges when there are a large number of dates included (eg: a year or more), since it only needs pairs of start and end dates, instead of the array of all the specific dates needed for excludeDates and includeDates.

The term "intervals" is kept instead of "range" because both props need to be fed with intervals as defined by date-fns. It could be changed, however, if needed.


martijnrusschen:
Awesome, thanks for your contribution!

Issue #3262, opened 2021-10-13
dateFormat array avoids minDate enter via keyboard


urretxagar:
**Describe the bug**
When defining an array of dateFormat (i.e. dateFormat={["dd.MM.yyyy"]}) and setting a minDate (i.e "10/10/2021") then the date defined in the minDate can not be entered via keyboard in the date picker input.

**To Reproduce**

1. Go to https://stackblitz.com/edit/react-ts-vmuz4q?file=index.tsx
2. Click on: Not Working example
3. Enter 10.10.2021

**Expected behavior**
10.10.2021 should be accepted as Valid date (minDate)

**Additional context**
The problem is there since version 4.0.0 (in version 3.8.0 works fine).

PR #3310, opened 2021-11-18
Fix isValid helper so it accepts dates equal to minDate (#3262)

krzempekk:
Argument `minDate` suggest that date should be equal or higher than it, but previous version wasn't accepting equal date. This solves the issue of not being able to input `minDate` via keyboard when `dateFormat` array was set.

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -132,7 +132,7 @@ export { isDate };
 
 export function isValid(date, minDate) {
   minDate = minDate ? minDate : new Date("1/1/1000");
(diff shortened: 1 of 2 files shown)


anastasiiaonchukova:
@martijnrusschen or someone else, can you check this PR please? this is blocking update for us

martijnrusschen:
This is looking good, thanks for adding the tests!

darkowic:
@martijnrusschen when do you plan to release it?

Issue #3315, opened 2021-11-22
Incorrect display today's month font in month picker


kastetcs95:
**Describe the bug**
The current month's font isn't in bold. 
Today is november, but this month isn't in bold.
![image](https://user-images.githubusercontent.com/66264861/142855795-ef367afa-a896-42c1-9ca1-23cf8d7fa277.png)
Example for year picker:
![image](https://user-images.githubusercontent.com/66264861/142856010-3ab74923-aa2c-4820-b545-6c9db6188d34.png)

**Steps to reproduce**
Go to month picker example https://reactdatepicker.com/#example-month-picker
Go to year picker example https://reactdatepicker.com/#example-year-picker

PR #3326, opened 2021-11-30
Fix: month text today not displayed

zzwt:
To be consistent with display of year text today


zzwt:
Related issue: https://github.com/Hacker0x01/react-datepicker/issues/3315

Issue #3318, opened 2021-11-25
Year dropdown scroll defaults to max value


ardibegolli:
Im duplicating [this issue](https://github.com/Hacker0x01/react-datepicker/issues/1159) since it was closed due to inactivity.

I'm using the showYearDropdown prop on the datepicker but when I click the year dropdown, it will always start at the top (max year). Is there a way to set the default value of the dropdown to the current year?

How it opens now:

<img width="262" alt="Screenshot 2021-11-25 at 11 16 45 AM" src="https://user-images.githubusercontent.com/40717247/143423018-eba88a62-4584-4e33-a8a7-bb3447a3082e.png">

How it is expected to be opened:

<img width="261" alt="Screenshot 2021-11-25 at 11 17 39 AM" src="https://user-images.githubusercontent.com/40717247/143423119-21842279-1697-4d05-9cc6-e266ceb217c9.png">

javlund:
I had this issue as well. Turns out that it was due to me using `value` instead of `selected` for the value. Hope that helps someone.

PR #3366, opened 2021-12-08
Fix year dropdown scroll on open

zzwt:
Related issue:https://github.com/Hacker0x01/react-datepicker/issues/3318

Scroll the year to current selected year for scrollable year dropdown on open.

Before:
image

After: 
image


martijnrusschen:
Thanks

Issue #3229, opened 2021-09-29
Accessibility issue - current date should use aria-current="date"


SKaplan01:
**Describe the bug**

When a screenreader user navigates through the calendar, the current date is not announced as the current date.

**To Reproduce**

1. Go to the react-datepicker demo (https://reactdatepicker.com/)
2. Turn on a screenreader (I used Voiceover)
3. Open the calendar and press tab, focus will land on current date and screenreader will read something like "Choose Wednesday September 29, 2021, button"

**Expected behavior**

When the focus is on the current date, the screen reader should read something like "Choose Wednesday September 29, 2021, current date, button"

aria-current="date" can be added to the current date button.

PR #3388, opened 2021-12-20
Accessibility: current date should use aria-current="date"

xmakina:
**What does this PR do?**
This PR adds aria-current to the element which represents the current date/month/year.

Actual Result:
Narrator Behaviour: "Choose Wednesday September 29, 2021, button"

Expected Result:
Narrator Behaviour: "Choose Wednesday September 29, 2021, **current date**, button"

Related Issue
#3229

### DIFF
--- a/src/day.jsx
+++ b/src/day.jsx
@@ -223,6 +223,8 @@ export default class Day extends React.Component {
     );
   };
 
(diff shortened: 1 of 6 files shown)


martijnrusschen:
Looking great, thanks!

PR #3397, opened 2021-12-23
Provide option to reset monthSelectedIn in setOpen

bartzy:
This change provides the option to reset the `monthSelectedIn` state in `setOpen` calls.
It's helpful when you'd like to close the date picker and avoid "stale" state of the selected month when opening it again (while it's not unmounted yet).

I did not see a reason to add a docs example or tests here since there are no tests for the setOpen API.


bartzy:
Any updates on this perhaps?

PR #3402, opened 2021-12-25
Fix Markdown formatting.

RodolfoSilva:


martijnrusschen:
This is generated automatically, fixing it manually will break it next time we build the docs.

RodolfoSilva:
:(

PR #3407, opened 2021-12-29
inverted navigation arrows for right to left languages

fterraneo:
When datepicker is rendered under dir="rtl" (for example in an application in arabic language) the navigation arrows used to increment/decrement the month were not inverted. 

The change consists in rendering the buttons in a different order by playing on the css.


martijnrusschen:
This looks good, if you can add a test we can merge it.

PR #3433, opened 2022-01-11
chore: use the actual Popper types

amandapouget:
When using the current types, and passing something like `{ preventOverflow: { enabled: true } }`, it results in a typescript error for not passing an array of objects.

I'm thinking that perhaps the original type used here was assigned in error, and it was meant to be the Popper props.

I'm not sure what the comments `<datepicker /> props` was intended to indicate.

Unless we really intended people to pass an array here.... ?


amandapouget:
nevermind I'm conflating proptypes and typescript

PR #3461, opened 2022-02-04
Add one column month layout for month picker

tggreene:
I have a use case for showing an inline list of months in a single column layout; this was the simplest possible way of adding it and I'm open to making the column layout a parameter the user passes in when creating the component instead, perhaps with a utility function that produces the different possible layouts.

If it's not suitable or desired feel free to discard!


src/month.jsx
```diff
@@ -397,10 +399,26 @@ export default class Month extends React.Component {
       [8, 9],
       [10, 11],
     ];
+    const monthsOneColumn = [
```
martijnrusschen, line 402: Can you add a test to make sure this functionality is covered?

PR #3516, opened 2022-03-13
Enable specifying aria-label prefix for month

kimamula:
Though `<Month>` component has `ariaLabelPrefix` prop, it is not currently possible for react-datepicker users to specify this prop. Here I added `monthAriaLabelPrefix` prop to `<DatePicker>` component and pass it to `<Month>` component to fix this issue.

### DIFF
--- a/src/month.jsx
+++ b/src/month.jsx
@@ -329,7 +329,7 @@ export default class Month extends React.Component {
 
   getAriaLabel = (month) => {
     const {
-      ariaLabelPrefix = "Choose",
+      chooseDayAriaLabelPrefix = "Choose",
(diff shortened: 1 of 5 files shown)


src/month.jsx
```diff
@@ -329,7 +329,7 @@ export default class Month extends React.Component {
 
   getAriaLabel = (month) => {
     const {
-      ariaLabelPrefix = "Choose",
+      chooseDayAriaLabelPrefix = "Choose",
```
kimamula, line 332: Using `ariaLabelPrefix` prop here seems like a mistake.

Issue #3556, opened 2022-04-14
Use change raw handler with custom input


kolserdav:
**Describe the bug**
When using a custom input with a raw change handler, event `onBlur` fires on a every change in the input field.

**To Reproduce**

1. Type on custom input

**Expected behavior**
Event `onBlur` should not fire on every input

**Desktop (please complete the following information):**

- OS: [Windows]
- Browser [Firefox]
- Version [99.0.1]

**Additional context**
```jsx
import { forwardRef, useState } from 'react';
import DatePicker from 'react-datepicker';
import 'react-datepicker/dist/react-datepicker.css';
import './App.css';

function App() {
  const [date, setDate] = useState('');
  const [selected, setSelected] = useState(null);

  const changeDateRawHandler = (e) => {
    setDate(e.target.value);
  }

  const changeDateHandler = (e) => {
    setSelected(e);
  }

  const DateInput = forwardRef(
    ({ onClick, value }, ref) => (
      <div>
        <input
          value={value}
          ref={ref}
          placeholder="dd.mm.YYYY"
          onChange={changeDateRawHandler}
        />
        <button onClick={onClick} type="submit">
          click
        </button>
      </div>
    )
  );
  return (
    <div className="App">
      <DatePicker
          customInput={<DateInput />}
          onChange={changeDateHandler}
          selected={selected}
          required
          locale="ru"
          value={date}
          showYearDropdown
          showMonthDropdown
          scrollableYearDropdown
        />
    </div>
  );
}

export default App;
```

kolserdav:
It is solved that
```typescript
const changeDateRawHandler = (e) => {
    setDate(e.target.value);
    // added inputRef emit focus
    setTimeout(() => {
      if (inputRef.current) {
        inputRef.current.focus();
      }
    }, 10);
  }

  const changeDateHandler = (e) => {
    setSelected(e);
  }

  const DateInput = forwardRef(
    ({ onClick, value }, ref) => (
      <div>
        {/** inputRef to input */}
        <input
          value={value}
          ref={inputRef}
          placeholder="dd.mm.YYYY"
          onChange={changeDateRawHandler}
        />
        {/** ref to button */}
        <button ref={ref} onClick={onClick} type="submit">
          click
        </button>
      </div>
    )
  );
```

kolserdav:
https://github.com/kolserdav/issue-datepicker/blob/master/src/App.js

PR #3579, opened 2022-05-04
Set up dev server

kolserdav:
FIXED:
- Set up `dev:hello-world` command
- Added example `custom-input-rawhandler` command `dev:custom-input-rawhandler`
- Make `js:dev `and `cd example/package && npm run start` as a live reload dev server!
- Fixed some dev conflicts
- Connected examples as local one react

TODOS;
- Set up css dev server to concurrently stdout


kolserdav:
I needed to solve this problem https://github.com/Hacker0x01/react-datepicker/issues/3556 and I solved it in this example.

PR #3582, opened 2022-05-06
fix: formatFunc should take source params

YuCJ:
`props.formatWeekDay` of `DatePicker` will be passed to `getFormattedWeekdayInLocale` via argument `formatFunc` as a callback.

The naming of the prop suggests that it is designed to provide flexibility for users to format the weekday. So I think the expected behavior is taking the source date and locale as input and returning the formatted value (i.e. As the replacement of the built-in `formatDate`). But now it  takes the formatted string as input.

### Expected Usage Example

```js
import { format } from 'date-fns';

const myCustomFormatString = 'EEEEE'
const myCustomPrefix = '#'

function formatWeekDay(date, dateFnsLocale) {
  return myCustomPrefix + format(date, myCustomFormatString, { locale: dateFnsLocale })
}

function App () {
  // ...
  return (
    <DatePicker
      formatWeekDay={formatWeekDay}
      {/* ... */ }
    />
  )
}

```

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -377,7 +377,7 @@ export function getLocaleObject(localeSpec) {
 }
 
 export function getFormattedWeekdayInLocale(date, formatFunc, locale) {
-  return formatFunc(formatDate(date, "EEEE", locale));
+  return typeof formatFunc === "function" ? formatFunc(date, locale) : formatDate(date, "EEEE", locale));
 }
 
 export function getWeekdayMinInLocale(date, locale) {
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -377,7 +377,7 @@ export function getLocaleObject(localeSpec) {
 }
 
 export function getFormattedWeekdayInLocale(date, formatFunc, locale) {
-  return typeof formatFunc === "function" ? formatFunc(date, locale) : formatDate(date, "EEEE", locale));
+  return typeof formatFunc === "function" ? formatFunc(date, locale) : formatDate(date, "EEEE", locale);
 }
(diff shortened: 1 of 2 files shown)


src/date_utils.js
```diff
@@ -377,7 +377,7 @@ export function getLocaleObject(localeSpec) {
 }
 
 export function getFormattedWeekdayInLocale(date, formatFunc, locale) {
-  return formatFunc(formatDate(date, "EEEE", locale));
+  return typeof formatFunc === "function" ? formatFunc(date, locale) : formatDate(date, "EEEE", locale));
```
YuCJ, line 380: Fix typo:

```suggestion
  return typeof formatFunc === "function" ? formatFunc(date, locale) : formatDate(date, "EEEE", locale);
```

YuCJ:
The typo failed the linting is fixed 🙏

martijnrusschen:
I ran the jobs but looks like there's a failure.

YuCJ:
> I ran the jobs but looks like there's a failure.

I updated the test case for `formatWeekday` in `test/calendar_test.js`. And it fits the use case mentioned in PR description and can be passed now. 👌

maranomynet:
This was a breaking change, and down-stream projects are gradually starting to explode because of this, as is demonstrated by https://github.com/Hacker0x01/react-datepicker/issues/3918.

The proper way to do this would have been to add `date` and `locale` as the 2nd and 3rd arguments of the `formatWeekDay` callback.

That way code relying on the 1st parameter being  a string, would continue to work.

PR #3548, opened 2022-04-11
Allow React 18 peer dep

jonasws:

PR #3599, opened 2022-05-17
Fix test `isQuarterDisabled should be disabled if not in included dates`

asydikov:
Fixes the test currently failing for all builds as mentioned in https://github.com/Hacker0x01/react-datepicker/pull/3548#issuecomment-1124820851.

Test was wrongly assuming quarters had no more than 40 days. Fixed by using `addQuarters(day, 1)` rather than `addDays(day, 40)`.


martijnrusschen:
Thanks!

PR #3713, opened 2022-08-02
Navigate in month-year picker using up and down arrow keys

bidahmetov02:
Add two more cases to onMonthKeyDown function to handle navigating using up and down arrow keys.

### DIFF
--- a/test/month_test.js
+++ b/test/month_test.js
@@ -18,6 +18,10 @@ function getKey(key) {
       return { key, code: 37, which: 37 };
     case "ArrowRight":
       return { key, code: 39, which: 39 };
+    case "ArrowUp":
+      return { key, code: 38, which: 38 };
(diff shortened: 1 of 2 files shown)


test/month_test.js
```diff
@@ -534,6 +538,54 @@ describe("Month", () => {
...
+        preSelected = param;
+      };
+
+      const monthComponent = renderMonth({
+        selected: utils.newDate("2015-08-01"),
+        day: utils.newDate("2015-08-01"),
+        setPreSelection: setPreSelection,
+        preSelection: utils.newDate("2015-08-01"),
+      });
+      // monthComponent
```
anastasiiaonchukova, line 553: Can this be removed?

bidahmetov02: Yes, thanks.

PR #3755, opened 2022-09-11
added necessary step for getting correct css on calender

luddens:
When I followed the steps without doing yard css:dev or the equivalent npm command the dropdown calender had no CSS applied, so I added this step to ensure that anyone that runs it locally won't be confused

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,8 @@ To begin local development:
 
 1. `yarn install`
 2. `yarn build-dev`
-3. `yarn start`
+3. `yarn css:dev`
+4. `yarn start`
 
 The last step starts documentation app as a simple webserver on http://localhost:3000.
 


danMateer:
Hey @luddens heads up I wasn't able to reproduce this. Running `yarn build-dev` should update the build artifacts - so running it should apply any updates to the SCSS in:
```
src/stylesheets/datepicker.scss
```
For CSS assets that will be used (`dist/react-datepicker.css`). Any direct edits to `.css` files will be overwritten when you compile SCSS, but running `yarn build-dev` _should_ eliminate the need to run `yard css:dev`.

Any details you can provide about your workflow when you ran into this? It might also be an IDE thing. Which one are you using? There are some tools to just auto-compile SCSS on save (e.g., https://atom.io/packages/sass-autocompile).

danMateer:
@martijnrusschen I think this one can be closed out, unless @luddens I'm missing the mark here.

luddens:
Okay. I'll upload a video showing me do the npm install, the npm run
build-dev and then npm run start to show the css malfunctioning.

Weird that I have to run css:dev too.

danMateer:
> Okay. I'll upload a video showing me do the npm install, the npm run build-dev and then npm run start to show the css malfunctioning. Weird that I have to run css:dev too.

Ah, possible we're running `npm run start` here instead of `yarn start`?

luddens:
I don't have TEE installed, so that errors out and the CSS never builds properly.

luddens:
The same thing happens when I run yarn start or npm run start. Hmmm.
Interesting. I had to go to court today, spent half the day upstate so I
didn't get to make a video of the issue, yet. Basically every time I want
to display changes I have to do the css build.

PR #3787, opened 2022-10-06
fix: fix accessibility issues adding tab loop to datepicker with portal

gcoayla:
I'm using this date picker with the withportal prop, I noticed there's no tab loop enabled on this case, This PR uses the `TabLoop` component to wrap the portal contents.

Also, I added a onKeyDown handler to allow closing the datePicker using the `Escape` key, previously it was ignored especially when focus was on the header.


mikelnorth:
Can we get this merged? It would be a huge gain to have these accessibility bugs fixed.

Issue #3766, opened 2022-09-22
Screen reader is staying silent on activating 'Next month' & 'previous month' buttons.


msftedad:
**Describe the bug**
Screen reader is not announcing any information on activating 'Next month' & 'previous month' buttons of 'Date picker' dialog of 'Birthday', 'enroll date' & graduate date' edit fields.

**To Reproduce**
1. Open the above mentioned 'URL' in 'Microsoft Edge' & login with valid credentials.
2. 'Intern support' Home page will get open.
3. Navigate to 'Candidate' link via tab key & activate it using enter.
4. 'Candidates' detail page will get open.
5. navigate to 'New candidate' link via tab key & activate it using enter.
6. A page with 'Form' fields to create a candidate will get open.
7. Turn on the narrator using 'Ctrl+Windows+Enter' keys.
8. Navigate to the 'Next' & 'Previous' buttons present in 'Date picker' dialog of 'Birthday', 'enroll date' & graduate date' edit fields via tab key & activate it.
9. Observe the issue.

**Actual behavior**
Screen reader is not narrating anything on activating the 'next month' & 'Previous' month buttons.

**Expected behavior**
Screen reader should narrate name of the respective month, which appears upon activating 'next month' & 'Previous' month buttons.

**Screenshots**

**Desktop (please complete the following information):**
OS: Windows 11 Version: 21H2 (OS Build 22000.856)
Browser: Microsoft Edge Version 105.0.1343.42 (Official build) (64-bit)
URL: [https://internsupport.microsoft.com/](https://nam06.safelinks.protection.outlook.com/?url=https:%2f%2finternsupport.microsoft.com%2f&data=05%7c01%7cv-knarang%40microsoft.com%7cf76091e47ae04f67a01208da90094e27%7c72f988bf86f141af91ab2d7cd011db47%7c1%7c0%7c637980666901980033%7cUnknown%7cTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7c3000%7c%7c%7c&sdata=R1BmtgD%2f2S7RZRcKY6DbVJrEnckvGhdnsVy5XuBrbTQ%3D&reserved=0)
Screen Reader: Narrator

**Smartphone (please complete the following information):**

**Additional context**
Similar issue repro, as screen reader is only narrating 'page title' information on selecting any date.

[...]

Issue #3559, opened 2022-04-14
Accessibility issue - selected date is not announced by screen reader


Ke1sy:
**Describe the bug**
When you select a date with keyboard, it is expected that the new date will be announced by screen reader, but nothing is announced.

**To Reproduce**
1. Start the screen reader and open the date-picker with keyboard.
2. Navigate to any date and click on it.

**Expected behavior**
Selected date will be announced by screen reader.

**Desktop (please complete the following information):**
- OS: Windows 10
- Browser: Chrome (v. 100.0.4896.75), Firefox

martijnrusschen:
Fixed by https://github.com/Hacker0x01/react-datepicker/pull/3817

Issue #3558, opened 2022-04-14
Accessibility issue - selected month is not announced by screen reader


Ke1sy:
**Describe the bug**
When you change month, it is expected that the new month with year will be announced by screen reader. 
Currently after clicking next/prev month button, user doesn’t know that month has changed and what month is currently selected, because nothing is announced. It is recommended to add aria-live="polite" and announce changed information to user.

**To Reproduce**
1. Start the screen reader and open the date-picker with keyboard.
2. Navigate to previous or next month button, then click it

**Expected behavior**
Selected month with year will be announced by screen reader.

**Desktop (please complete the following information):**

- OS: [e.g. iOS] Windows 10
- Browser: Chrome (v. 100.0.4896.75), Firefox

martijnrusschen:
fixed by https://github.com/Hacker0x01/react-datepicker/pull/3817

PR #3817, opened 2022-11-10
Fix: Screen reader does not announce selected date, current month

SpaNb4:
This PR fixes some accessibility issues, and may help screen reader users navigate this datepicker library better.

- With my fix when you change the month, screen reader will announce the current month. Because with current implementation, when clicking on the next/previous month button, the user does not know which month the calendar is currently on
- Also made some changes so that screenreader will announce selected date

I believe that my PR will resolve these issue: 
https://github.com/Hacker0x01/react-datepicker/issues/3558
https://github.com/Hacker0x01/react-datepicker/issues/3559 
https://github.com/Hacker0x01/react-datepicker/issues/3766


SpaNb4:
@martijnrusschen Do I need to add more tests to increase the code-coverage and only then will you review this PR, right?

src/index.jsx
```diff
@@ -997,6 +1003,75 @@ export default class DatePicker extends React.Component {
...
+          ? "End date: " +
+            safeDateFormat(this.props.endDate, {
+              dateFormat: longDateFormat,
+              locale,
+            })
+          : ""
+      }`;
+    } else {
+      if (this.props.showTimeSelectOnly) {
+        ariaLiveMessage = `Selected time: ${safeDateFormat(
```
ch1qui, line 1031: Hardcoded english texts will ruin accessibility for all non-english users. This prefix should be diffferent prop.

martijnrusschen:
Can you resolve the merge conlfict so we can get this merged and released?

src/calendar.jsx
```diff
@@ -1004,6 +1040,7 @@ export default class Calendar extends React.Component {
           showPopperArrow={this.props.showPopperArrow}
           arrowProps={this.props.arrowProps}
         >
+          {this.renderAriaLiveRegion()}
```
SergioVanacloigCarlsberg, line 1043: Shouldn't this span be optional?

LukeBrandon: I agree, this should be optional.  This is showing up in my application and I cannot opt out of showing the "Selected date:" text.

Issue #3747, opened 2022-09-01
Year dropdown scroll position not working when either minDate / maxDate props is being specified


wenxiangtoh:
**Describe the bug**
The scroll position of the year dropdrown is not scrolled towards the currently selected year when either minDate / maxDate prop is being specified. The scroll position will always stay in the center.

**To Reproduce**

Possible Scenario 1
1. Specify minDate prop value as minDate = {new Date(1950,1,1)}
3. Open Calendar
4. Select the year dropdown
5. To Observe the scroll position (Centered)
6. Select a year like example: 1970

Possible Scenario 2
1. Specify maxDate prop value as maxDate = {new Date(2022,1,1)}
3. Open Calendar
4. Select the year dropdown
5. To Observe the scroll position (Centered)
6. Select a year like 2000

**Expected behavior**
Whenever the year dropdown is opened, the scroll position should be referencing the currently selected year

**Screenshots**
When maxDate is being specified
![maxDate_year_2022](https://user-images.githubusercontent.com/60644202/187826101-27977c95-51e8-4f9f-b49b-f0bddb4d8fc6.png)

When minDate is being specified
![minDate_year_1950](https://user-images.githubusercontent.com/60644202/187826104-89880af0-5667-4f0c-ad2a-bbcbe580af3a.png)

**Desktop (please complete the following information):**

- OS: [e.g. iOS]: iOS
- Browser [e.g. chrome, safari]: Google Chrome
- Version [e.g. 22]: 104.0.5112.101 

**Additional context**

Muhiq146:
If you use this prop, than I think your issue will be solved.

`dropdownMode="select"`

It will reference the selected or current year.

wenxiangtoh:
Hi, using the `dropdownMode = 'select' ` does help to solve the scroll position issue but It seems not compatible with CSS style, making it display a large list instead of a manageable size list.
<img width="77" alt="select_dropdown_mode_with_min_and_max_date" src="https://user-images.githubusercontent.com/60644202/189586820-bd8def17-6829-4e92-8e7b-8408af75276c.png">

kit52:
@wenxiangtoh Have you solved this problem?

Muhiq146:
@wenxiangtoh 
Hi, yes it creates this issue and looks bad in UI because it is not taking prop `yearDropdownItemNumber` into consideration. But it will solve your referencing issue and as well as min & max date issue.

I will try to solve, if you have any solution please share.

[...]

Issue #3646, opened 2022-05-30
Year dropdown scroll defaults to center of list


AbilityMaster:
I'm using the showYearDropdown prop and maxDate prop on the datepicker but when I click the year dropdown, it will always start at the center of list. Is there a way to set the default value of the dropdown to the current year?

**How it opens now:**
![image](https://user-images.githubusercontent.com/44331653/171069156-edee7929-65d7-4f1a-bf1c-505e49207d59.png)

**How it is expected to be opened:**
![image](https://user-images.githubusercontent.com/44331653/171069257-8fe49f68-0c34-449c-a49b-f9d569d8d50f.png)

anas510:
Same behavior is happening for me too
@AbilityMaster any solution so far?

kit52:
@anas510 , @AbilityMaster 
Is there a solution to this problem?

turkus:
> @anas510 , @AbilityMaster Is there a solution to this problem?

I'm affraid it's because of yearDropdownItemNumber, it doesn't adjust the height using scroll option. If you remove that prop, it centers.

SashaRepin2:
@AbilityMaster helped with the solution of this problem

PR #3844, opened 2022-12-03
Fix(year dropdown options): Year dropdown scroll defaults to center of list

SashaRepin2:
Description bug:
The scroll position of the year dropdown is not scrolled towards the currently selected year. The scroll position will always stay in the center.

Issues:
#3747 
#3646


martijnrusschen:
Looks like the checks aren't succeeding, can you check what's going on?

SashaRepin2:
Hi, I checked what the problem was and fixed it.

Issue #3876, opened 2022-12-27
month picker keyboard selected class names issue and update solution


sanghogu:
**Describe the bug**
disabledKeyboardNavigation props in month picker doesn't work

**Expected behavior**
css and side event issues with classname

**Additional context**
Solution Modify 
line 321 of src/month.jsx as follows: !this.props.disabledKeyboardNavigation && utils.getMonth(preSelection) === m,

Hvesyr:
Having the same issue. 
Anyone working on this project still?

sanghogu:
Perfect

yuki0410-dev:
Is there still a problem with this issue?
The fix appears to be complete.
(cc. @martijnrusschen )

PR #3887, opened 2023-01-06
fix(months): disabledKeyboardNavigation not hiding classname in month picker

Hvesyr:
Description bug:
This PR fixes the issue when using month picker with disabledKeyboardNavigation enabled still showing the keyboard navigation classes on the respective months resulting in weird styling.

Fixes this issue #3876


Rastaman231:
Nice work

Issue #3804, opened 2022-10-28
minDate date now cant select month current


chayut27:
<DatePicker
        selected={startDate}
        minDate={new Date()}
        onChange={(date) => setStartDate(date)}
        showMonthYearPicker
  />

10/28/2022

i cant select month oct, I think it should be selectable.

Issue #2912, opened 2021-04-26
Month and year view not working properly with minDate


toomski:
**Describe the bug**
As a user, I want to set minDate to now with a new Date().

I think is related to [https://github.com/Hacker0x01/react-datepicker/pull/2186](https://github.com/Hacker0x01/react-datepicker/pull/2186)

Below is the minimum code to reproduce the problem:

[https://codesandbox.io/s/mitorski-forked-5ccrm ](https://codesandbox.io/s/mitorski-forked-5ccrm )

**Result**
From the day view picker, all is fine.
From the month view picker, I can't select the current month (April)
From year view picker I can't select the current year (2021)

**To Reproduce**

1. Select month from select on left and try to select current month (April)
2. Select year from select on left and try to select the current year (2021)

**Expected behavior**
I want to select the current month (April) and year (2021)

**Desktop (please complete the following information):**

- OS: any
- Browser: any
- Version: any
- Version of react-datepicker - 3.8.0

AlexYanichev:
same here
![image](https://user-images.githubusercontent.com/40503291/157933478-860d20ed-d471-4682-a6d5-d7c2de34532a.png)

limes:
Why was the wontfix label added? The above described problem still occurs

shu1007:
I too faced this problem.
In fixing this, I think we first need to consider whether these are breaking the minDate convention by returning the first day of the month or the first day of the year.
(e.g. minDate is 2022/12/30 and selecting 2022 returns 2022/1/1 and selecting 2022/12 returns 2022/12/1 )

shu1007:
Another issue similar to minDate is the issue of excluded dates. Opinions differ as to how the exclusion dates should be treated when they are included.

PR #3888, opened 2023-01-07
Fix month and year picker disabled

shu1007:
Nice to meet you. We have fixed some of the issues mentioned in the issue.(#2912 #3804 )

- YearMonthPicker and yearPicker can now select the same month/year as the specified minDate and maxDate.
- Some Unit test case added


shu1007:
The code for which coverage is noted is not the one that has been corrected at this time.
https://app.codecov.io/gh/Hacker0x01/react-datepicker/pull/3888

Could you do a code review?

martijnrusschen:
Thanks for your fix and extensive tests on this subject

PR #3902, opened 2023-01-20
Calendar Icon Changes

Subrat1988:
Code changes for adding calendar icon to react-datepicker.
I've used react-icons to create the icon.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,6 +117,7 @@
     "classnames": "^2.2.6",
     "date-fns": "^2.24.0",
     "prop-types": "^15.7.2",
+    "react-icons": "^4.7.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -117,6 +117,7 @@
     "classnames": "^2.2.6",
     "date-fns": "^2.24.0",
     "prop-types": "^15.7.2",
+    "react-icons": "^4.7.1",
```
martijnrusschen, line 120: I'd prefer not add more dependencies to this package. Can you find another way to make this work?

martijnrusschen (changes requested):

PR #3903, opened 2023-01-23
fix: Inconsistent/broken behavior in `parseDate` 

maranomynet:
**See the expanded commit messages for details.**

After this fix, the parsing behavior of `react-datepicker`…

a) is more consistent between locales/browsers/OS platforms
b) has fewer bugs
c) may feel slightly more "strict" in certain situations.

Item "c" results from parseDate no longer falling back (with string values for `props.dateFormat`) on free-form `new Date(value)` parsing of input values when `date-fns` has already declared the value as "invalid".

However, there's **still** a lot of flexibility in the parsing, as `date-fns/parse` provides very sensible flexibility based on localized RegExp configs, and also since `props.dateFormat` can be an Array of valid formattings.

It thus seems fairly sensible (and more deterministic) to fully place the responsibility of parsing on `date-fns` and not second-guesss its results.

(In fact the previous `new Date(value)` parsing causes surprising and deceptively incorrect "false positive" parsing in situations where the dateFormat was `dd.MM.yyyy` and users type in a date in this format `dd.MM.yy`. Then their input gets "helpfully" re-parsed as `MM.dd.yy`, which was most definitely not intended by the developer setting up the datepicker input.)

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -70,39 +70,24 @@ export function newDate(value) {
   return isValid(d) ? d : null;
 }
 
-export function parseDate(value, dateFormat, locale, strictParsing, minDate) {
-  let parsedDate = null;
-  let localeObject =
+export function parseDate(value, dateFormat, locale, strictParsing, refDate) {
+  const localeObject =
     getLocaleObject(locale) || getLocaleObject(getDefaultLocale());
-  let strictParsingValueMatch = true;
-  if (Array.isArray(dateFormat)) {
-    dateFormat.forEach((df) => {
-      let tryParseDate = parse(value, df, new Date(), {
-        locale: localeObject,
-      });
-      if (strictParsing) {
-        strictParsingValueMatch =
-          isValid(tryParseDate, minDate) &&
-          value === formatDate(tryParseDate, df, locale);
-      }
-      if (isValid(tryParseDate, minDate) && strictParsingValueMatch) {
-        parsedDate = tryParseDate;
-      }
-    });
-    return parsedDate;
+
+  const formats = Array.isArray(dateFormat) ? dateFormat : [dateFormat];
+  refDate = refDate || newDate();
+
+  for (let i = 0, len = formats.length; i < len; i++) {
+    const format = formats[i];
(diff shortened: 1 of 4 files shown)


Greg-Smulko:
Hey @martijnrusschen, could you please share your thoughts about the future of this change w.r.t. discussions in https://github.com/Hacker0x01/react-datepicker/issues/3921 and https://github.com/Hacker0x01/react-datepicker/issues/3952?

It feels to me like this change was introduced without fully appreciating how much it is a breaking change. I'd be happy to see it being kept, I'd be (even more ;) ) happy to have it reverted, but it would be great to have clarity of the planned way forward with it.

Thanks in advance! :)

martijnrusschen:
Released a new version to revert the breaking change.

maranomynet:
@martijnrusschen do you have any plans for resolving the issues this PR was addressing?

The old (current?) parseDate behavior is causing real problems, for users outside the USA English usage domain.

I realise people have started to rely on the old mis-designed behavior, but if this PR were part of a v5 release, a migration guide (and updated documentation) could explain how to get the old behavior back by using `props.dateFormat`.

That way we can make everybody happy.

maranomynet:
BTW, I'm currently working around these issues by intercepting `props.dateFormat` and rewriting them like this:

```ts
  dateFormat={
    // NOTE: Force all dateFormat values into Array<string> to temporarily work around
    // a bug in the current version of react-datepicker where invalid **string** values
    // are re-parsed with `new Date()`, causing surprising (weird) false positives
    // AND where Arrayed formats get parsed in order of "increasing priority".
    // Revert back to the plain `dateFormat={dateFormat}` pass-through once this PR has been
    // accepted and released: https://github.com/Hacker0x01/react-datepicker/pull/3903
    // (NOTE: the PR was merged and release as 4.9.0, but then reverted in 4.11.0.
    // still awaiting resolution @2023-04-06)
    typeof dateFormat === 'string'
      ? [dateFormat]
      : dateFormat.slice(0).reverse()
    // dateFormat
  }
```

PR #3915, opened 2023-02-01
Calendar icon fix  #3525

### LINKED ISSUE #3525: Calender icon
datepicker comes with normal text field view, blank. Could there be an icon? which is kind of good for the UI, I guess.

febababy:
Calendar icon fix https://github.com/Hacker0x01/react-datepicker/issues/3525

Screenshot from 2023-02-01 11-28-29

### DIFF
--- a/docs-site/src/examples/default.js
+++ b/docs-site/src/examples/default.js
@@ -1,6 +1,10 @@
 () => {
   const [startDate, setStartDate] = useState(new Date());
   return (
-    <DatePicker selected={startDate} onChange={(date) => setStartDate(date)} />
+    <DatePicker
+      showIcon
+      selected={startDate}
+      onChange={(date) => setStartDate(date)}
+    />
   );
 };
--- a/docs-site/src/examples/default.js
+++ b/docs-site/src/examples/default.js
@@ -1,10 +1,6 @@
 () => {
   const [startDate, setStartDate] = useState(new Date());
   return (
-    <DatePicker
-      showIcon
-      selected={startDate}
-      onChange={(date) => setStartDate(date)}
-    />
+    <DatePicker selected={startDate} onChange={(date) => setStartDate(date)} />
   );
 };
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -503,11 +503,19 @@ export default class DatePicker extends React.Component {
       this.props.showTimeSelectOnly &&
       !isSameDay(date, this.props.selected)
     ) {
-      date = set(this.props.selected, {
-        hours: getHours(date),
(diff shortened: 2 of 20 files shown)


martijnrusschen:
I like the idea. To make this less disruptive, I'd like to have this as an optional prop that can be set. This makes it possible for users to add this icon if they want to, but aren't forced to use it.

febababy:
@martijnrusschen Sure will make the icon an optional prop

docs-site/src/examples/default.js
```diff
@@ -1,6 +1,10 @@
 () => {
   const [startDate, setStartDate] = useState(new Date());
   return (
-    <DatePicker selected={startDate} onChange={(date) => setStartDate(date)} />
+    <DatePicker
+      showIcon
```
martijnrusschen, line 5: Can you create a separate example in the docs site that shows how to make a datepicker with an icon in the input?

src/index.jsx
```diff
@@ -1176,8 +1177,22 @@ export default class DatePicker extends React.Component {
   };
 
   renderInputContainer() {
+    const { showIcon } = this.props;
     return (
-      <div className="react-datepicker__input-container">
+      <div
```
martijnrusschen, line 1182: Can you add a test that checks if the icon (or CSS class) is shown when the boolean is set to true.

src/stylesheets/datepicker.scss
```diff
@@ -517,6 +517,15 @@
   position: relative;
   display: inline-block;
   width: 100%;
+  .react-datepicker__calendar-icon {
+    position: absolute;
+    padding: 0.5rem;
```
martijnrusschen, line 522:
```suggestion

  .react-datepicker__calendar-icon {
    position: absolute;
    padding: 0.5rem;
```

src/stylesheets/datepicker.scss
```diff
@@ -517,6 +517,15 @@
...
+  }
+}
+.react-datepicker__view-calendar-icon {
+  input {
```
martijnrusschen, line 526:
```suggestion
}

.react-datepicker__view-calendar-icon {
  input {
```

src/index.jsx
febababy: Sure

febababy:
@martijnrusschen I have added an example and test cases.

martijnrusschen:
This looks good, in the screenshot you added the icon is shown for both the default and the example you added. Is that intentional?

febababy:
The icon is optional
image

martijnrusschen:
thank you!

PR #3925, opened 2023-02-08
Update index.jsx

SPLAYER-HD:
Hi,

I need to change the styles in real time, like the border to red when the field is invalid, So I added styles as additional prop.
at the moment I am patching your package in my local project, but it would be nice if you could merge this change


SPLAYER-HD:
Also, my theme is defined in real-time so my system is based on the styles properties of my components.
Thanks

martijnrusschen:
Can you add some tests for this functionality?

Girl-aventuries (approved):

martijnrusschen (changes requested):
Please take a look at the comments

Issue #3906, opened 2023-01-25
Date Range using input does not update date values on input update.


w-gluza:
**Describe the bug**
Changing the range value inside the input does not update the value in the calendar. 
Please note input works when you set up a range for the first time,  when updating the date, issues arise. 

**To Reproduce**
1. Create date range input using:
```
() => {
  const [dateRange, setDateRange] = useState([null, null]);
  const [startDate, endDate] = dateRange;
  return (
    <DatePicker
      selectsRange={true}
      startDate={startDate}
      endDate={endDate}
      onChange={(update) => {
        setDateRange(update);
      }}
      isClearable={true}
    />
  );
};
```
2. Set up a date range either by input or calendar GUI.
3. Try to change the date using the input field. The date does not update.

PS You can also see this bug in official docs under the:
[Date Range using input with clear button](https://reactdatepicker.com/#example-date-range-using-input-with-clear-button)

**Expected behavior**
Dates updates either if they are updated through calendar GUI or by typing inside the input field.

**Desktop (please complete the following information):**

- OS: [iOS]
- Browser [chrome]
- Version [109]

galmeida12:
Any updates on this issue? It seems that the current "solution" for this is to revert to v3.

MichaelMakesGames:
For others looking for an interim solution, you can use `onChangeRaw` prop with a function like the following:
```typescript
    function onChangeRaw(raw: null | undefined | string) {
        if (selectsRange) {
            if (raw != null) {
                let rawStart = raw;
                let rawEnd = '';
                if (raw.includes(' - ')) {
                    rawStart = raw.split(' - ')[0];
                    rawEnd = raw.split(' - ')[1];
                }

                const referenceDate = new Date(new Date().getFullYear(), 0, 1);
                let startDate: null | Date = null;
                try {
                    startDate = parse(rawStart, standardizedDateFormat, referenceDate);
                } catch (e) {
                    console.error('Failed to parse start date', e);
                }

                let endDate: null | Date = null;
                try {
                    endDate = parse(rawEnd, standardizedDateFormat, referenceDate);
                } catch (e) {
                    console.error('Failed to parse end date', e);
                }

                onChange([
                    startDate && isValid(startDate) ? startDate : null,
                    endDate && isValid(endDate) ? endDate : null,
                ]);
            }
        }
    }
```
Note: `parse` and `isValid` are imported from `date-fns`.
Maybe not the most robust solutions; but seems to be working fine for me.

[...]

Issue #3596, opened 2022-05-16
Changing range value in input does not update value in the calendar


serhiiminin:
**Describe the bug**
When changing the input value in range mode it does not trigger the change in the calendar.

```
() => {
  const [range, setRange] = useState([new Date("2014/02/08"), new Date("2014/02/10")]);
  const [startDate, endDate] = range;
  return (
    <>
      <DatePicker
        onChange={(date) => setRange(date)}
        selectsRange={true}
        startDate={startDate}
        endDate={endDate}
      />
    </>
  );
};
```

**To Reproduce**

1. Use my setup
2. Click on input to open the popup with the calendar
3. Try to change the input value
4. Observe that the calendar value is not changed

**Expected behavior**
If the value in input is a valid date then the value in the calendar is changed (similar to the single select behavior)

**Screenshots**
<img width="228" alt="image" src="https://user-images.githubusercontent.com/22465603/168579626-ab5ed6b9-c2f5-40db-bcd8-cc2055dc4e86.png">

**Desktop (please complete the following information):**

- OS: iOS
- Browser chrome

**Additional context**
I looked into the code and it seems the issue is with parsing. The parsing function can't handle ranges.
https://github.com/Hacker0x01/react-datepicker/blob/master/src/index.jsx#L486

dvelichuk:
Same issue for all browsers. Mac osx

eliranhasin1121:
Hi, do you have any updates about this issue?

jphawk:
Same issue here. It would be great to have a fix for this

laug:
See also #3906 and #4002 which report the same issue.

PR #3988, opened 2023-04-06
fix: Inconsistent/broken behavior in `parseDate` 

maranomynet:
> **Note**
> This is a re-opening of #3903 (since reverted by #3976) in the hope of it eventually getting accepted as a "breaking change".

**See the expanded commit messages for details.**

After this fix, the parsing behavior of `react-datepicker`…

a) is more consistent between locales/browsers/OS platforms
b) has fewer bugs
c) is the same whether `dateFormat` is string or an array of strings
d) treats arrayed `dateFormat` values as "most preferred first" (instead of the opposite).
e) is slightly more "strict".

Item "e" results from `parseDate` no longer falling back (with string values for `props.dateFormat`) on free-form, hard-to-predict `new Date(value)` parsing of input values when `date-fns` has already declared the value as "invalid".

It seems sensible (and more deterministic) to place the responsibility of parsing fully on `date-fns` and not (sometimes) second-guesss its results.

The `new Date(value)` parsing causes surprising and deceptively incorrect "false positive" parsing in situations where the dateFormat was `dd.MM.yyyy` and users type in a date in this format `dd.MM.yy`. Then their input gets "helpfully" re-parsed as `MM.dd.yy`, which was most definitely not intended by the developer setting up the datepicker input.

However, there's **still** a lot of flexibility in the parsing, as `date-fns/parse` provides very sensible flexibility based on localized RegExp configs, and also since `props.dateFormat` can be an Array of valid formattings.

As people are often relying on the old mis-behavior this PR needs to become part of a v5 release. A migration guide (and, possibly, updated documentation) could explain how to get the old behavior back by using `props.dateFormat` with multiple formats in order of decreasing priority.

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -72,39 +72,24 @@ export function newDate(value) {
   return isValid(d) ? d : null;
 }
 
-export function parseDate(value, dateFormat, locale, strictParsing, minDate) {
-  let parsedDate = null;
-  let localeObject =
+export function parseDate(value, dateFormat, locale, strictParsing, refDate) {
+  const localeObject =
     getLocaleObject(locale) || getLocaleObject(getDefaultLocale());
-  let strictParsingValueMatch = true;
-  if (Array.isArray(dateFormat)) {
-    dateFormat.forEach((df) => {
-      let tryParseDate = parse(value, df, new Date(), {
-        locale: localeObject,
-      });
-      if (strictParsing) {
-        strictParsingValueMatch =
-          isValid(tryParseDate, minDate) &&
-          value === formatDate(tryParseDate, df, locale);
-      }
-      if (isValid(tryParseDate, minDate) && strictParsingValueMatch) {
-        parsedDate = tryParseDate;
-      }
-    });
-    return parsedDate;
+
+  const formats = Array.isArray(dateFormat) ? dateFormat : [dateFormat];
+  refDate = refDate || newDate();
+
+  for (let i = 0, len = formats.length; i < len; i++) {
+    const format = formats[i];
+    const parsedDate = parse(value, format, refDate, { locale: localeObject });
+    if (
+      isValid(parsedDate /* , minDate */) &&
+      (!strictParsing || value === formatDate(parsedDate, format, locale))
+    ) {
+      return parsedDate;
+    }
   }
-
-  parsedDate = parse(value, dateFormat, new Date(), { locale: localeObject });
-
-  if (strictParsing) {
-    strictParsingValueMatch =
-      isValid(parsedDate) &&
-      value === formatDate(parsedDate, dateFormat, locale);
-  } else if (!isValid(parsedDate)) {
-    parsedDate = new Date(value);
-  }
-
-  return isValid(parsedDate) && strictParsingValueMatch ? parsedDate : null;
+  return null;
 }
 
 // ** Date "Reflection" **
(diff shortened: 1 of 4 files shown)


martijnrusschen:
It's been a long time since this was first submitted, but I agree with the thinking here. If you're still willing to bring this forward we can work towards releasing this as part of a major version bump.

mirus-ua:
@maranomynet hello 
Can you update your PR based on this one https://github.com/Hacker0x01/react-datepicker/pull/4707
and then we can do a major release, including your breaking changes

mirus-ua:
@martijnrusschen hello
this PR looks promising. Should we try to adapt it for v7 with the rest of the breaking changes?

martijnrusschen:
@mirus-ua I think that would be good.

src/date_utils.js
```diff
@@ -76,59 +72,24 @@ export function newDate(value) {
...
-      parsedDate = parse(value, dateFormat.slice(0, value.length), new Date());
-    }
+  const formats = Array.isArray(dateFormat) ? dateFormat : [dateFormat];
+  refDate = refDate || newDate();
 
-    if (!isValid(parsedDate)) {
-      parsedDate = new Date(value);
+  for (let i = 0, len = formats.length; i < len; i++) {
+    const format = formats[i];
+    const parsedDate = parse(value, format, refDate, { locale: localeObject });
```
WeatherGod, line 84: While I have not tried this myself (I'm not experienced enough to know how), I *believe* this would fix a bug of mine that I discovered last week. I'm using UTCDate objects with the datepicker (set as `selected`, `minDate`, and `maxDate`), but dates entered manually in the input box are not getting parsed as UTCDates, and each iteration of manual date input results in an additional timezone offset to the datetime. I *think* this is because the original call to `parse()` passed in a `new Date()` for the reference date to `date-fns`'s `parse()` instead of the component's `minDate` or some other UTCDate object.

My temporary workaround is to use `startOfDay(new UTCDate(newDate)).toISOString()` in the `onChange` function.

laug:
It seems this PR was skipped over for v7, is there any chance of getting it included in a future v8?
I would like to work on fixing  #3596 #3906 #4002 and this patch would be a prerequisite for fixing the parsing of ranges.

src/date_utils.js
```diff
@@ -146,21 +107,15 @@ export function formatDate(date, formatStr, locale) {
   if (locale === "en") {
     return format(date, formatStr, { awareOfUnicodeTokens: true });
   }
-  let localeObj = getLocaleObject(locale);
+  const localeObj =
+    getLocaleObject(locale) || getLocaleObject(getDefaultLocale()) || null;
```
laug, line 111: Adding `|| getLocaleObject(getDefaultLocale())` here will prevent the warning below from being shown when the `locale` string is invalid, as `localeObj` will not be nullish even when the `locale` string is invalid.

src/index.jsx
```diff
@@ -496,6 +496,7 @@ export default class DatePicker extends React.Component {
       this.props.dateFormat,
       this.props.locale,
       this.props.strictParsing,
+      this.props.selected,
       this.props.minDate
```
laug, line 500: `parseDate` only takes 5 parameters but is being passed 6 parameters here. This PR renames the 5th from `minDate` to `refDate`, but `minDate` is probably still needed to validate the parsed date correctly.

src/date_utils.js
```diff
@@ -76,59 +72,24 @@ export function newDate(value) {
   return isValid(d) ? d : null;
 }
 
-export function parseDate(value, dateFormat, locale, strictParsing, minDate) {
-  let parsedDate = null;
-  let localeObject =
+export function parseDate(value, dateFormat, locale, strictParsing, refDate) {
```
laug, line 75: Why was `minDate` replaced with `refDate`, instead of just adding `refDate` as an additional parameter?
`minDate` should be kept as a parameter, and passed to `isValid` below.

martijnrusschen:
closing in favor of https://github.com/Hacker0x01/react-datepicker/pull/5036

Issue #624, opened 2016-11-14
Add option to inherit font size


localjo:
Right now, sizes are set in `rem` units, which means that the component has a different size than it's parent both for the font and the spacing of the calendar. In my case, this makes the calendar extremely small compared to everything else on my page. I can override that will some clever CSS, but it would be nice if there was a prop to make the component use `em` units instead to inherit the sizes from it's parent. Something like `relativeSize={true}`. If this is something others want, I can open a PR.

DannyRyman:
I have the same problem when using alongside the stylesheet from the react-bootstrap.

josiahsprague:
I'd also be willing to open a PR that converts this component to use JS styles (quickly becoming the React norm) instead of a global stylesheet that has to be included separately. @Hacker0x01 do you have any thoughts on which direction you'd like to take this?

matthizou:
Same issue as DannyRyman. The Bootstrap stylesheet has a side effect on the component, and the calendar appears tiny.

acasaccia:
Thanks @josiahsprague for raising this! Would you share the CSS rules you used to override the default size while the PR is ready?

blueboxjesse:
Also experiencing this issue. Would also appreciate the CSS rules. Will try to hack my way through it tonight.

josiahsprague:
This is what I'm using;
```
.react-datepicker {
  font-size: 1em;
}
.react-datepicker__header {
  padding-top: 0.8em;
}
.react-datepicker__month {
  margin: 0.4em 1em;
}
.react-datepicker__day-name, .react-datepicker__day {
  width: 1.9em;
  line-height: 1.9em;
  margin: 0.166em;
}
.react-datepicker__current-month {
  font-size: 1em;
}
.react-datepicker__navigation {
  top: 1em;
  line-height: 1.7em;
  border: 0.45em solid transparent;
}
.react-datepicker__navigation--previous {
  border-right-color: #ccc;
  left: 1em;
}
.react-datepicker__navigation--next {
  border-left-color: #ccc;
  right: 1em;
}
```
[...]

PR #3998, opened 2023-04-14
Replace all `rem` units with `em` units

Ugoku:
This fixes small calendars on sites where something like `html { font-size: 62.5%; }` is set (see #624)


Ugoku:
It's been a month without any reply... I guess we'll just incorporate this CSS inside our project then. Happy to leave it open in case someone in the future decides to merge it.

martijnrusschen (approved):

martijnrusschen:
Sorry for the delay, this looks good.

Issue #3885, opened 2023-01-05
The Today Button is not accessible via Tab key for keyboard users


talatkuyuk:
The problem is: The keyboard users can not focus to the Today Button (actually it is a `div`) via `Tab` key.

Today button's tab index (tabIndex) actually should be set to 0. Otherwise it is not accessible from keyboard.

https://github.com/Hacker0x01/react-datepicker/blob/master/src/calendar.jsx#L734

marwanKettleCom:
Yes having the same issue.  Please fix.  Thank you!

LGLabGreg:
There is no need to add a button as there is already logic to set tabindex, but currently it is possible that no date has tabindex=0 if the datepicker loads with no selected date.

https://github.com/Hacker0x01/react-datepicker/blob/master/src/day.jsx#L282

So if not keyboardSelected/selected/preselected it should default to today or first day of the month.

LGLabGreg:
Here's a solution

```
const datePickerRef = useRef(null);

useEffect(() => {
  if (datePickerRef.current) {
    datePickerRef.current.setPreSelection(new Date());
  }
}, [datePickerRef]);

<DatePicker ref={datePickerRef} />
```

talatkuyuk:
My solution is inline with `selectsRange` prop of the date picker. Update yours accordingly. 

I assume you use `todayButton="Today"` prop already to see Today button at the bottom of the date picker.

Firstly, use `onCalendarOpen={setupTodayButton}` as a prop.

```tsx
const setupTodayButton = () => {
    setTimeout(() => {
      const todayButton = document.querySelector(
        ".react-datepicker__today-button",
      ) as HTMLElement | null;
      if (!todayButton) return;

      todayButton.setAttribute("tabindex", "0");
      todayButton.setAttribute("role", "button");

      const applyTodayRange = () => {
        const startDate = new Date();
        const endDate = new Date();
        endDate.setHours(23, 59, 59, 999);
        handleChange([startDate, endDate]); // replace it with your state setter
      };

      todayButton.onclick = applyTodayRange;

      todayButton.onkeydown = (e) => {
        if (e.key === "Enter" || e.key === " ") {
          e.preventDefault();
          applyTodayRange();
        }
      };
    }, 0);
  };
```
If you use it as **inline** date picker, then you should use an useEffect, I suppose.

PR #4006, opened 2023-04-21
Feature/today button can be accessed with keyboard

oherst:
Hello! This PR addresses issue https://github.com/Hacker0x01/react-datepicker/issues/3885

>The Today Button is not accessible via Tab key for keyboard users

Our apps have just been audited and not being able to reach the "Today" button with the keyboard has been flagged as an issue.

I believe this change backwards compatible.

- I've added a `button` element inside the existing `div` and moved the event handler on to the button.
- I've adjusted the styles so that anyone who has previously customised the today button shouldn't notice any changes.

Some considerations:

- I think I would have preferred to change the css class names so that the parent `div` would be `react-datepicker__today-button-container` and then the `button` could have used `react-datepicker__today-button` but that would break any existing custom styles.
- We could just pass a button as the todayButton prop (despite the prop type being specified as `string`) but the click handler would still be on the parent `div` - admittedly not a big deal.

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -706,11 +706,13 @@ export default class Calendar extends React.Component {
       return;
     }
     return (
-      <div
-        className="react-datepicker__today-button"
-        onClick={(e) => this.handleTodayButtonClick(e)}
-      >
-        {this.props.todayButton}
+      <div className="react-datepicker__today-button">
+        <button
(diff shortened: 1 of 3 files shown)


oherst:
This PR is not required. You can pass a `button` to the `todayButton` prop to make it accessible.

Issue #3952, opened 2023-03-03
Getting "Range Error: Invalid Time Value" on time picker setup 


alextselegidis:
**Describe the bug**
Hello and thanks for this great component. The following setup throws the aforementioned error. 

**To Reproduce**

1. Go to [reactdatepicker.com](https://reactdatepicker.com/)
2. Use the first example code box
3. Try the following code to use the time picker setup 
4. Type "11:1" in the time picker manually (do not select)
5. See the error

Code: 

```
() => {
  const [startDate, setStartDate] = useState(null);
  return (
    <DatePicker
      selected={startDate}
      onChange={(date) => setStartDate(date)}
      showTimeSelect
      showTimeSelectOnly
      timeIntervals={15}
      timeCaption="Time"
      dateFormat="HH:mm"
      timeFormat="HH:mm"
    />
  );
};
```

**Expected behavior**
The user must be able to complete the whole time input manually, till "11:11" for example. 

The error comes already as soon as "11:1" is typed. 

This only happens if you have set `null` as the default value of the time picker. 

**Screenshots**

<img width="452" alt="image" src="https://user-images.githubusercontent.com/4527441/222790213-1fd5b905-cb61-49ec-8258-e4db781e0f92.png">

stee-ren:
I'm seeing the same thing, looks like this issue was introduced with v4.9.0, if I had to guess, might have something to do this this contribution:
fix: Inconsistent/broken behavior in parseDate by @maranomynet
 https://github.com/Hacker0x01/react-datepicker/pull/3903

@alextselegidis looks like the quick way out is to roll back to v4.8.0. 
Also, I used the "p" formatting, so the locale would be part of the formatting. I suspect the issue relates to the dateFormat. This is used to format the input field, not the timeFormat. The timeFormat is only used to format the time drop down.

@mellis481 I took your example and added a locale. 
https://codesandbox.io/s/billowing-moon-sf06pt?file=/src/App.js
What I found is, if the dateFormat was removed, or set to something like "Pp" (localized date and time), it was rejected. If both dateFormat and timeFormat are set to "p" and the locale set to "de" (where 11:11 is correct), it explodes.

Also, I don't know if its a related regression but it looks like previously, setting the locale to a string e.g. locale="pt-BR" (taken directly from the documentation, it doesn't work. Console shows the warning:

`
A locale object was not found for the provided string ["pt-BR"]. 
`
[...]

PR #4036, opened 2023-05-22
Added check for date-fns set

AstroProjection:
This fix is related to https://github.com/Hacker0x01/react-datepicker/issues/3952

Specifically when, we try and Input a custom time into a blank input field, the component crashes.

It appears to be because the `set` method is being called on a null object ( i.e the.props.selected )

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -501,6 +501,7 @@ export default class DatePicker extends React.Component {
     // Use date from `selected` prop when manipulating only time for input value
     if (
       this.props.showTimeSelectOnly &&
+      this.props.selected &&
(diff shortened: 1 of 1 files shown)


akshayhk24:
https://github.com/Hacker0x01/react-datepicker/assets/122598646/46532506-904a-4442-bab5-13e1fad834f4

AstroProjection:
https://github.com/Hacker0x01/react-datepicker/assets/13699537/f36bdbf8-021f-4cd3-bebb-21992a92b6e7

martijnrusschen:
Thanks, good catch!

PR #4044, opened 2023-05-26
fix: Uneven margin in inline calendar

### LINKED ISSUE #4028: Ugly right "margin" appearing for inline calendar
At certain widths, the multiple month inline calendar displays a right "margin" with no content in it.  The additional space always appears to contain the right arrow control.

**To Reproduce**
1. Go to https://reactdatepicker.com/#example-multiple-months-inline
1. Scroll down to the Multiple months inline example
1. I use Firefox desktop.  
ReactDatepickerScreenshot_2023-05-19a

**Another Way to Reproduce**
1. This is an actual screenshot from my app, which is mobile/tablet first and repsonsive.  It starts off with a minor padding issue, then as the size increases, it gets worse, then it actually resolves itself when the screen width is large enough.

Bad:
ReactDatepickerScreenshot_2023-05-19c

Worse:
ReactDatepickerScreenshot_2023-05-19b

Resolved at wider screens:
ReactDatepickerScreenshot_2023-05-19d

**Expected behavior**
There should be no padding on the right.  The calendar should be "tight."

**Desktop: **

- Version 4.1.1
- OS: Windows 10
- Browser Firefox 103.0.1 (64-bit)
- React 17.0.2

**Smartphone:**
This was not tested on a smartphone.

**Additional context**
Appears to cause issues for small to medium size screen widths (from about 4 to 6 inches screen width) on desktop.  Mobile devices not tested.

navedqb:
Fixes #4028.

### DIFF
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -215,10 +215,6 @@
   }
 }
 
-.react-datepicker__month-container {
-  float: left;
-}
-
 .react-datepicker__year {
   margin: $datepicker__margin;
   text-align: center;


Ugoku:
This broke having 3 months next to each other :(

tkmcmaster:
This also broke having the time on the right of the month. It pushes it underneath which is ugly.

PR #4051, opened 2023-06-01
Fixes off centered clear button

### LINKED ISSUE #3974: The Datepicker Clear button is off centered only on windows
I can't find a quick fix sadly. Screenshot: 

image

Screenshot from the demo. It's still off-center.
image

navedqb:
Fixes #3974.

### DIFF
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -5,7 +5,6 @@
   display: inline-block;
   padding: 0;
   border: 0;
-  width: 100%;
 }
 
 .react-datepicker {


mellis481:
@navedqb @martijnrusschen This is a bad solution to solving the clear button problem as it will cause the `<input>` element used by the datepicker to no longer span the full widget of its container.

1Jesper1:
@mellis481 I have the same problem when updating.

1Jesper1:
@martijnrusschen @mellis481 Maybe we can add this as a breaking style change in the release? https://github.com/Hacker0x01/react-datepicker/releases/tag/v4.14.0

Issue #4066, opened 2023-06-12
aria-selected should read all range dates as selected


EimantasBlazevicius:
**Is your feature request related to a problem? Please describe.**
aria-selected should read all range dates as selected when selecting in day picker. That is accessibility issue, currently only one date is reported as selected.

**Describe the solution you'd like**
aria-selected should read all range dates as selected. 

**Describe alternatives you've considered**
no good ideas, not even hacky ones :)

**Additional context**
From what I have observed in this file:
https://github.com/Hacker0x01/react-datepicker/blob/master/src/day.jsx#LL360C7-L360C40
Changing this line from: 
aria-selected={this.isSelected()}
to
aria-selected={this.isSelected() || this.isInRange()}
Would solve issue with not reading range as selected.

I would be happy to do a PR for this.

EimantasBlazevicius:
https://github.com/Hacker0x01/react-datepicker/pull/4068

PR #4068, opened 2023-06-12
aria-selected true for range selected as well

EimantasBlazevicius:
Not sure how to link these, but this can be linked 
https://github.com/Hacker0x01/react-datepicker/issues/4066


martijnrusschen:
Are you able to add a test to cover this behaviour?

PR #4095, opened 2023-06-23
[FEAT] Custom content renders for month, quarter and year

KevinFabre-ods:
I work for an organization that requires this UI for day, month and year elements in selected / selecting range.

| Day | Month | Year |
|---|---|---|
|Screenshot 2023-06-22 at 15 35 47|Screenshot 2023-06-22 at 15 42 29|Screenshot 2023-06-23 at 11 19 34|

To do this UI, it required to use custom content renders:
- One div `.react-datepicker__{precision}` for the background.
- Another div `.react-datepicker__{precision}-content` to wrap the text in some kind of chip.

This PR aims to offer custom render for month, quarter and year content in the same way as custom day content render.

### Question:

- Should these props be singular or plural ? renderYearContent / renderYearContent**s**
_I prefer the singular form as the element has only one content._

### DIFF
--- a/src/date_utils.js
+++ b/src/date_utils.js
@@ -8,8 +8,6 @@ import addWeeks from "date-fns/addWeeks";
 import addMonths from "date-fns/addMonths";
 import addQuarters from "date-fns/addQuarters";
 import addYears from "date-fns/addYears";
-import subMinutes from "date-fns/subMinutes";
-import subHours from "date-fns/subHours";
 import subDays from "date-fns/subDays";
 import subWeeks from "date-fns/subWeeks";
 import subMonths from "date-fns/subMonths";
@@ -35,7 +33,6 @@ import min from "date-fns/min";
 import max from "date-fns/max";
 import differenceInCalendarDays from "date-fns/differenceInCalendarDays";
 import differenceInCalendarMonths from "date-fns/differenceInCalendarMonths";
-import differenceInCalendarWeeks from "date-fns/differenceInCalendarWeeks";
 import differenceInCalendarYears from "date-fns/differenceInCalendarYears";
 import startOfDay from "date-fns/startOfDay";
 import startOfWeek from "date-fns/startOfWeek";
@@ -272,16 +269,7 @@ export { addMinutes, addDays, addWeeks, addMonths, addQuarters, addYears };
 
 // *** Subtraction ***
 
-export {
-  addHours,
-  subMinutes,
-  subHours,
-  subDays,
-  subWeeks,
-  subMonths,
-  subQuarters,
-  subYears,
-};
+export { addHours, subDays, subWeeks, subMonths, subQuarters, subYears };
 
 // ** Date Comparison **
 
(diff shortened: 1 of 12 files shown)


martijnrusschen:
> Should these props be singular or plural ? renderYearContent / renderYearContents
> I prefer the singular form as the element has only one content.

Agreed, I think that makes more sense.

KevinFabre-ods:
> 



> > Should these props be singular or plural ? renderYearContent / renderYearContents
> > I prefer the singular form as the element has only one content.
> 
> Agreed, I think that makes more sense.

So, maybe it's a good time to rename renderDayContents to renderDayContent ?

src/date_utils.js
```diff
@@ -8,8 +8,6 @@ import addWeeks from "date-fns/addWeeks";
 import addMonths from "date-fns/addMonths";
 import addQuarters from "date-fns/addQuarters";
 import addYears from "date-fns/addYears";
-import subMinutes from "date-fns/subMinutes";
```
martijnrusschen, line 11: I assume this is just clean up?

src/index.jsx
```diff
@@ -119,10 +119,6 @@ export default class DatePicker extends React.Component {
       timeInputLabel: "Time",
       enableTabLoop: true,
       yearItemNumber: DEFAULT_YEAR_ITEM_NUMBER,
-
-      renderDayContents(date) {
```
martijnrusschen, line 123: was this unused?

martijnrusschen (approved):

martijnrusschen:
> > 
> 
> > > Should these props be singular or plural ? renderYearContent / renderYearContents
> > > I prefer the singular form as the element has only one content.
> > 
> > 
> > Agreed, I think that makes more sense.
> 
> So, maybe it's a good time to rename renderDayContents to renderDayContent ?

sounds good to me.

martijnrusschen:
> > > 
> > 
> > 
> > > > Should these props be singular or plural ? renderYearContent / renderYearContents
> > > > I prefer the singular form as the element has only one content.
> > > 
> > > 
> > > Agreed, I think that makes more sense.
> > 
> > 
> > So, maybe it's a good time to rename renderDayContents to renderDayContent ?
> 
> sounds good to me.

Only concern is that it may be  a breaking change. Perhaps leave that for now, if it's breaking the API.

src/index.jsx
KevinFabre-ods: Yes, as there is [already a check](https://github.com/Hacker0x01/react-datepicker/blob/e67fe3a6f46d27e8765b9193f1ec90a383dcad0e/src/day.jsx#L343) if this prop is defined.

```javascript
   return this.props.renderDayContents
      ? this.props.renderDayContents(getDate(this.props.day), this.props.day)
      : getDate(this.props.day);
```

src/date_utils.js
KevinFabre-ods: Yes, during build there were three warnings about unused imports

martijnrusschen: ah great, should probably implement a linter for that :)

PR #4096, opened 2023-06-23
Add eslint plugin to detect unused imports

martijnrusschen:
Based on the config from https://www.npmjs.com/package/eslint-plugin-unused-imports


martijnrusschen:
fyi @KevinFabre-ods

PR #4099, opened 2023-06-30
github actions: update to latest versions where applicable

Zarthus:
Giving some love to existing github actions and paving the path to upgrading the CI to Node 18 (and 20) later down the line.

In a nutshell:

- Upgrades actions to the latest major versions where it is safe to do so
- Remove Node 14 from the CI matrix as it is unsupported
  - https://endoflife.date/nodejs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,10 +13,10 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [16.x]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -13,10 +13,10 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x]
```
martijnrusschen, line 16: Yay!

Zarthus, line 16: https://endoflife.date/nodejs - node 14 is EOL so we shouldn't test on it anymore, testing on 18 should become the default and 20 would be ideal.

martijnrusschen (approved):

Issue #4133, opened 2023-07-11
Node 18 Support


Zarthus:
We'd like to make our dev env available in Node 18 / Node 20, this requires some things:

- [x] Webpack update (https://github.com/Hacker0x01/react-datepicker/pull/4132)
- [x] Deps update (https://github.com/Hacker0x01/react-datepicker/pull/https://github.com/Hacker0x01/react-datepicker/pull/4100)
- [ ] https://github.com/Hacker0x01/react-datepicker/issues/4107 (delegated)
- [x] Switch to craco for the docs-site? (not necessary)
- [x] Remove `@babel/plugin-proposal-private-property-in-object": "^7.21.11",`

Zarthus:
We can close this for now; as the last remaining entry has an issue already.

PR #4150, opened 2023-07-12
investigate failure on CI

Zarthus:
Refs https://github.com/Hacker0x01/react-datepicker/issues/4133


Zarthus:
Merging as CI is green

PR #4189, opened 2023-08-05
This enhancement aims to add a holiday list as parameters or props to the react-datepicker component.

sushilkundu143:
The background will be set to orange for holidays in the calendar. If a holiday is selected, its background color will change to blue. Additionally, on mouseover, a tooltip will display the details of the corresponding holiday.


martijnrusschen:
This looks very similar to https://reactdatepicker.com/#example-highlight-dates-with-custom-class-names-and-ranges. Can you explain how this would be different and can't be solved by the example above?

.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
```
martijnrusschen, line 1: This isn't related to this PR. Please revert

docs-site/yarn.lock
martijnrusschen, line 1: Please revert

martijnrusschen (changes requested):
Code contains a bunch of unrelated changes. After checking the code, this looks very similar to the highlight day function. Can't you achieve what you're trying to do with the existing functions? Test coverage seems to be missing too.

Issue #4122, opened 2023-07-09
Module not found: Can't resolve 'react-onclickoutside'


Satish789748:
"react-datepicker": "4.13.0",
    In this version getting this issue.
    error - ../../node_modules/react-datepicker/dist/react-datepicker.min.js:1:15
Module not found: Can't resolve 'react-onclickoutside'

Zarthus:
I was unable to reproduce, can you give more insights?

The dep is also visible in the 4.13 tag

yuki0410-dev:
The PR for this issue has already been merged and appears to be resolved.

( cc. @martijnrusschen )

PR #4196, opened 2023-08-10
fixed/Module not found: Can't resolve 'react-onclickoutside' #4122

gyananshu07:
fixed/Module not found: Can't resolve 'react-onclickoutside' #4122

A dependency react-onclickoutside was missing.


martijnrusschen:
Not sure why this should solve the issue you're referring to, but bumping the version is fine.

Issue #4107, opened 2023-07-03
Migrate test framework away from Karma


Zarthus:
**Is your feature request related to a problem? Please describe.**

Karma does not support Node JS 18+ (unless you run `export NODE_OPTIONS=--openssl-legacy-provider` [NOT RECOMMENDED]). GitHub Actions actively prohibits toggling this option on.

NodeJS 16 will EOL in 2 months time: https://endoflife.date/nodejs

It is unclear for how long GitHub Actions will keep supporting Node 16, but we should assume the EOL date.

**Describe the solution you'd like**
We need to migrate away from Karma to another testing framework, e.g. Jest

This may be partially automatable with something like https://jestjs.io/docs/migration-guide#jest-codemods

**Describe alternatives you've considered**

1. Run Github Actions on Node 18 with `--openssl-legacy-provider` (not possible)
2. Keep running the CI on node 16 but target development on 18 (not feasible long-term)
3. Use a different CI infrastructure than GitHub actions to run our testsuite (not feasible long-term)

**Additional context**

More context: https://github.com/Hacker0x01/react-datepicker/pull/4100

https://github.com/karma-runner/karma/commit/450fdfdac5b999967daec1020f1ac69cf9b854ab

martijnrusschen:
This is an important initiative. Any help we can get to migrate to Jest would be welcome. Using code mods may be give us a good start on getting the migration done.

PR #4229, opened 2023-08-31
karma to jest migration completed

polbene95:
Closes https://github.com/Hacker0x01/react-datepicker/issues/4107

In this PR we are migrating from Karma to Jest.

Extra changes: 
-  In `day.jsx`, `month.jsx`, & `week.jsx` `PropTypes.instanceOf(Element)` was changed for `PropTypes.object` because the first one doesn’t work on server side render making the test fail.

- In `day.jsx` we changed `shouldFocusDay && this.dayEl.current.focus({ preventScroll: true });` for `shouldFocusDay && this.dayEl.current?.focus({ preventScroll: true });` because in some test `this.dayEl.current` was `null` making the test fail.

What's missing:

- Scripts for `test`, `test:ci` & `test:watch` need to be updated.


package.json
```diff
@@ -90,15 +84,15 @@
     "rollup-plugin-replace": "^2.2.0",
     "rollup-plugin-terser": "^7.0.2",
     "rollup-plugin-uglify": "^6.0.4",
-    "sass": "1.66.1",
```
martijnrusschen, line 93: Why is this needed?

martijnrusschen:
Looking good overall, would be good to get CI scripts running so we can have the test suite running.

package.json
```diff
@@ -63,17 +65,9 @@
     "highlight.js": "^11.0.1",
     "husky": "8",
     "istanbul-instrumenter-loader": "^3.0.1",
-    "karma": "^6.4.2",
-    "karma-chai": "^0.1.0",
-    "karma-coverage": "^2.2.1",
```
martijnrusschen, line 68: We should consider finding an alternative for this. I think Jest may have support for this, but not sure how to set this up.

package.json
```diff
@@ -124,7 +118,7 @@
     "lint": "yarn run eslint && yarn run sass-lint",
     "prettier": "prettier --write '**/*.{js,jsx}'",
     "start": "yarn --cwd docs-site install && yarn --cwd docs-site start",
-    "test": "NODE_OPTIONS=--openssl-legacy-provider NODE_ENV=test karma start karma.conf.js --single-run",
+    "test": "jest --verbose",
     "test:ci": "NODE_ENV=test karma start karma.conf.js --single-run",
     "test:watch": "NODE_OPTIONS=--openssl-legacy-provider NODE_ENV=test karma start karma.conf.js --watch",
```
martijnrusschen, line 123: These 2 lines need to be updated as well.

package.json
polbene95: I don't know how did this got here. I reverted the change.

package.json
polbene95: Just updated the scripts.

package.json
polbene95: What do we need to find an alternative for exactly?

martijnrusschen: The code coverage tool

Zarthus (approved):
Thanks so much for doing this!

I reviewed all the non-test files and it gets a big thumbs up from me.

I skimmed over the new tests and they look good to me as well.

Issue #4283, opened 2023-10-01
Test Case Failing Due to Incomplete Data Display in DatePicker Component


balajis-qb:
### Issue:
The test case located in `datepicker_test.test.js`, within the `describe("selectsRange with inline")` block and named `it("should change dates of range change startDate when endDate set before startDate")` is currently failing under specific conditions.

### Problem:
In the above mentioned test-case, we attempt to find a date three days before the current date within a `DatePicker` component. 
![Test case issue](https://github.com/Hacker0x01/react-datepicker/assets/145029139/5b4ad87e-d9c0-4eb3-8175-bc8d5bcf9e69)

However, this test case is failing in certain scenarios.  The issue arises when the current date falls at the beginning of a month. In such cases, the `DatePicker` component does not display the previous month's data, which prevents us from selecting a date three days before the current date.
![The issue screenshot](https://github.com/Hacker0x01/react-datepicker/assets/145029139/5974c2a0-7cd5-434f-ab40-8bd826b876cd)

### To Reproduce
Normally, this issue will not occur, but I accidentally caught it on Oct. 1, 2023, as there were no fall-over days from last month in this month's week, so the component is not rendering the last month's data.
![Issue Screenshot](https://github.com/Hacker0x01/react-datepicker/assets/145029139/013361ba-8792-4e24-ad9b-4fce49f546c0)

### Suggested Solution:
To address this issue, we propose the following solution:

1. When the current date is at the beginning of the month and the `DatePicker` component does not display the previous month's data, we should navigate to the previous month programmatically.
2. After navigating to the previous month, we can then attempt to select the desired date three days before the current date.

### Expected behavior
With the suggested solution in place, the test case should be able to find and select a date three days before the current date, regardless of whether it falls at the beginning of the month.

### Screenshots
![The issue screenshot](https://github.com/Hacker0x01/react-datepicker/assets/145029139/f49d02c2-b903-4403-91b2-dc38f0300e64)

balajis-qb:
You can assign this issue to me.  I fixed the test case, I'll raise a PR for it

martijnrusschen:
Nice catch, yes this test needs to be fixed. Pretty useless if it doesn't work on the 1st of the month :)

balajis-qb:
Thank you @martijnrusschen for assigning the issue.  I fixed the test case.  Kindly review my PR https://github.com/Hacker0x01/react-datepicker/pull/4284 and let me know if any changes required

PR #4284, opened 2023-10-01
Fix: Ensure Proper Date Selection in DatePicker Test Case

balajis-qb:
Closes https://github.com/Hacker0x01/react-datepicker/issues/4283

### Proposed Changes:

- Added logic to the it("should change dates of range change startDate when endDate set before startDate") test case in datepicker_test.test.js to ensure proper date selection.
- Fixed the issue where the test was failing under specific conditions, particularly when the current date was at the beginning of a month.

### DIFF
--- a/test/datepicker_test.test.js
+++ b/test/datepicker_test.test.js
@@ -24,6 +24,24 @@ function getSelectedDayNode(datePicker) {
   );
 }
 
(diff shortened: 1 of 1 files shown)


martijnrusschen (approved):
Nice!

Issue #3635, opened 2022-05-19
Accessibility issue - when date picker closes focus does not return on input


TautvydasPetrauskas:
**Describe the bug**
When date picker closes pressing `Escape` or when `shouldCloseOnSelect` property is passed, focus does not return on input

**To Reproduce**

1. Go to (https://reactdatepicker.com/#example-open-to-date)
2. On the first default example add `shouldCloseOnSelect` property for `<DatePicker />`
3. Open date picker and select a date or press `Escape`
4. See that focus does not return on input.

**Expected behavior**
After close focus suppose to go on input.

**Screenshots**

https://user-images.githubusercontent.com/49157108/169313648-05350b56-317a-4043-a3d3-34980fe5c937.mov

PR #4285, opened 2023-10-02
fix(3635): a11y - Sets focus to input on select

justinseiter:
# Fixes: https://github.com/Hacker0x01/react-datepicker/issues/3635 : [Accessibility issue - when date picker closes focus does not return on input](https://github.com/Hacker0x01/react-datepicker/issues/3635) 

### Issue
When using the `Escape` or `Enter` keys as well as after selection with pointer, focus did not return to input.

Accessibility [guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/) suggest that in these instances focus should return to the input.

### What were the specific changes made?

- Adds a `sendFocusBackToInput` method for handling returning focus. There were already two separate (nearly identical) implementations to handle this within `handleSelect` and `onPopperKeydown`. The `sendFocusBackToInput` is simply an abstraction to facilitate reuse in those instances as well as several others.
- Adds tests around possible consumer implementations.

### Additional notes/context
This work makes considerations not only for default implementations but as well as:
- `isClearable` - Clicking clear now returns focus to input
- `customInput` - This one Just Works ™️ as long as implemented as outlined in docs (ie with `forwardRef`)
- `shouldCloseOnSelect` - Don't hide calendar on date selection means we'll keep focus within picker
- `disabledKeyboardNavigation`- Pointer only
- `showTimeSelect` - When this prop is enabled the picker currently closes once a time is selected. This work just makes sure that focus only returns on that event and not on date selection.
- `inline` - Focus remains on picker
- `withPortal` - Again, Just works ™️ thanks to previous contributors. :) 
- Pressing `Escape` while in the input field - If picker was open and focus was on input, pressing `Escape` would send focus to `body`. Now returns to input.

https://github.com/Hacker0x01/react-datepicker/assets/78192/5c668e2e-4104-46af-8a64-768e0713889d

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -302,6 +302,7 @@ export default class DatePicker extends React.Component {
   constructor(props) {
     super(props);
     this.state = this.calcInitialState();
+    this.preventFocusTimeout = null;
   }
 
   componentDidMount() {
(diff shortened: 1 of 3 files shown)


martijnrusschen (approved):
Thanks, this is looking great.

Issue #4291, opened 2023-10-06
How to make Calendar icon show in React Date Picker?


AtGbemiga:
**Describe the bug**
I'm using React Date Picker, which I installed via npm i react-datepicker. Everything works except the calendar icon which doesn't display. The docs says to add showIcon to the JSX element to display the calendar icon, this does not show the icon, it only makes the JSX element bigger.

**To Reproduce**
```tsx
"use client";

import { useState } from "react";
import DatePicker from "react-datepicker";
import "react-datepicker/dist/react-datepicker.css";


export default function Form() {
  const [dateRange, setDateRange] = useState<[Date | null, Date | null]>([
    null,
    null,
  ]);
  const [startDate, endDate] = dateRange;

  return (
    <form>
      <DatePicker
        showIcon
        selectsRange={true}
        startDate={startDate}
        endDate={endDate}
        onChange={(update) => {
          console.log("Selected Date Range:", update);
          setDateRange(update);
        }}
        minDate={new Date()}
        withPortal
      />
    </form>
  );
}
```

**Expected behavior**
A calendar icon should be displayed.

**Screenshots**
[No icon](https://github.com/AtGbemiga/car-rental-frontend/blob/main/No%20icon.PNG)

**Desktop (please complete the following information):**

- OS: Windows
- Browser chrome
- Version "^4.16.0"

frankops11:
@AtGbemiga show https://github.com/Hacker0x01/react-datepicker/pull/4295

AtGbemiga:
> @AtGbemiga show #4295

I don't understand what your response is stating.

frankops11:
Try the last versión and if problem persist please added a Sanbox or Stackblitz :)

AtGbemiga:
> Try the last versión and if problem persist please added a Sanbox or Stackblitz :)

Alright, I will as quickly as possible :)

frankops11:
Hey bro, you know, check it out and see if it's true. But the problem is more than anything because of a style:

After
![image](https://github.com/Hacker0x01/react-datepicker/assets/50973629/f6f01020-8507-4152-87cd-09e2b4a3b39c)

Before
![image](https://github.com/Hacker0x01/react-datepicker/assets/50973629/26365c83-26d8-4cc4-ae28-652fbbf4f836)

I'm going to review and look for the cause, because analyzing X's comment, this came from the last version. The funniest thing is that he never replied to me

skoldborg:
[...]

PR #4298, opened 2023-10-09
Fix: Display Issue of Calendar Icon in React Date Picker (#4291)

frankops11:
#### **🐛 Problem:**
The calendar icon in the React Date Picker was not displaying as expected due to a conflict arising from a global `* {box-sizing: border-box;}` style, as discussed in issue #4291. Specifically, the icon container's `padding` interfered with its available space for content, causing the icon to not render visibly.

#### **🛠 Solutions Considered:**
Two main solutions were pondered:
1. **Remove Padding:** Adjust the position using `left` and `right` properties.
2. **Change Box-Sizing:** Explicitly set `box-sizing: content-box;` for the icon.

#### **🚀 Adopted Solution:**
The second solution was opted for in this PR due to its straightforwardness and minimal impact on existing style setups. By assigning `box-sizing: content-box;` to the icon, it assures the padding does not encroach on the icon’s space, thereby permitting it to be displayed even when a global `box-sizing: border-box;` is applied.

#### **🧹 Additional Changes:**
A small code adjustment was also introduced to enhance developer flexibility while interacting with the default SVG. Previously, resolving certain issues or introducing styles could inadvertently modify the base class, potentially causing unintended outcomes. The specific code change made is as follows:
```diff
- className="react-datepicker__calendar-icon"
+ className={`${defaultClass} ${className}`}
```
Here, `defaultClass` holds the essential CSS class for the calendar icon, whereas `className` can be used to inject additional classes for style customization or alteration without influencing the base. This affords developers greater control and flexibility over the icon’s appearance and behavior while ensuring the integrity of the default styling remains intact.

### **🧪 Test Adjustments:**

In the process of resolving the primary issue, an adjustment was made to a test to ensure reliability and accuracy in its assertions. The change, as it pertains to the class name of the calendar icon, was as follows:

```javascript
- expect(showIconClass).toBe("react-datepicker__calendar-icon");
+ expect(showIconClass).toMatch(/^react-datepicker__calendar-icon\s?$/);
```

This modification was brought into play due to the trailing whitespace within the `class` attribute, a nuance not addressed by the original test assertion. Two options were contemplated to manage this: employing `.trim()` to dismiss the whitespace, or utilizing a regex pattern to account for its potential presence. 

The regex method was chosen for a few key reasons:
- **Clarity:** It transparently accommodates the potential for a trailing space without masking it through string manipulation, maintaining a truer reflection of the DOM attribute’s state.
- **Robustness:** It doesn’t change the string being tested, thereby reducing the chance of inadvertently altering or misrepresenting the attribute under examination.
- **Flexibility:** It leaves room for accommodating variations in whitespace handling in future code adjustments without demanding test alterations.

Given the above, the chosen test modification provides a balanced blend of transparency, reliability, and foresight. Although `.trim()` could be utilized for a cleaner string comparison, we deemed the clarity in reflecting actual attribute values to be paramount, and thus, the regex approach was favored.

For future reference and clarity, the explanation is documented here in the PR. The incorporation of such a rationale directly in the codebase was considered. However, maintaining a lean code and accommodating these specifics in the PR/documentation might serve a dual purpose of avoiding inline clutter while still ensuring knowledge is shared and archived for posterity.

#### **🎯 Outcome:**
With these changes, the calendar icon will display consistently, regardless of global box-sizing styling, and developers are afforded enhanced flexibility concerning icon styling and customization.

#### **🙏🏽 Request:**
We kindly urge the maintainers to consider merging this PR promptly as it addresses a user-facing issue, ensuring enhanced usability and experience across implementations.

#### **How to Test:**
```jsx
<DatePicker
  showIcon={true}
/>
```
Ensure that the calendar icon displays correctly even in environments where `* {box-sizing: border-box;}` is set.

---

### **Final Notes:**
This fix aims to seamlessly integrate with existing implementations, ensuring no disruption while enhancing reliability in varied styling contexts. Your feedback and considerations for a swift merge would be immensely valued.

---

### **Additional Suggestion:**
Moving forward, it might be beneficial to refine how we handle bug reports and feature requests in this repository. Enforcing a policy where issue creators provide a reproducible example, possibly through a StackBlitz or CodeSandbox setup, could streamline the debugging process. Having a readily available environment that mimics the reported issue will likely expedite the identification and resolution of potential defects, enabling us to maintain a robust and reliable codebase for all users.

We humbly propose this approach in the spirit of continuous improvement and welcome discussion on its potential implementation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "author": "HackerOne",
   "name": "react-datepicker",
   "description": "A simple and reusable datepicker component for React",
-  "version": "4.19.0",
+  "version": "4.19.1",
   "license": "MIT",
   "homepage": "https://github.com/Hacker0x01/react-datepicker",
   "main": "dist/index.js",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "author": "HackerOne",
   "name": "react-datepicker",
   "description": "A simple and reusable datepicker component for React",
-  "version": "4.19.1",
+  "version": "4.19.0",
   "license": "MIT",
   "homepage": "https://github.com/Hacker0x01/react-datepicker",
   "main": "dist/index.js",
--- a/src/calendar_icon.jsx
+++ b/src/calendar_icon.jsx
@@ -22,7 +22,7 @@ const CalendarIcon = ({ icon, className }) => {
   // Default SVG Icon
   return (
     <svg
-      className="react-datepicker__calendar-icon"
+      className={`${defaultClass} ${className}`}
       xmlns="http://www.w3.org/2000/svg"
       viewBox="0 0 448 512"
     >
--- a/test/datepicker_test.test.js
+++ b/test/datepicker_test.test.js
@@ -2234,7 +2234,7 @@ describe("DatePicker", () => {
       datePicker,
       "react-datepicker__calendar-icon",
     ).getAttribute("class");
-    expect(showIconClass).toBe("react-datepicker__calendar-icon");
+    expect(showIconClass).toMatch(/^react-datepicker__calendar-icon\s?$/);
   });
 
   describe("Year picker", () => {
--- a/src/stylesheets/datepicker.scss
+++ b/src/stylesheets/datepicker.scss
@@ -513,6 +513,7 @@
   .react-datepicker__calendar-icon {
     position: absolute;
     padding: 0.5rem;
+    box-sizing: content-box;
   }
 }
 


package.json
```diff
@@ -2,7 +2,7 @@
   "author": "HackerOne",
   "name": "react-datepicker",
   "description": "A simple and reusable datepicker component for React",
-  "version": "4.19.0",
+  "version": "4.19.1",
```
martijnrusschen, line 5: You don't need to change this, please revert.

src/calendar_icon.jsx
```diff
@@ -22,7 +22,7 @@ const CalendarIcon = ({ icon, className }) => {
   // Default SVG Icon
   return (
     <svg
-      className="react-datepicker__calendar-icon"
+      className={`${defaultClass} ${className}`}
```
martijnrusschen, line 25: Where do you define the default now?

martijnrusschen (changes requested):

src/calendar_icon.jsx
martijnrusschen, line 25: Can you explain why this change was needed? In the PR description you explained how the CSS change was needed, but this seems unrelated.

test/datepicker_test.test.js
```diff
@@ -2234,7 +2234,7 @@ describe("DatePicker", () => {
       datePicker,
       "react-datepicker__calendar-icon",
     ).getAttribute("class");
-    expect(showIconClass).toBe("react-datepicker__calendar-icon");
+    expect(showIconClass).toMatch(/^react-datepicker__calendar-icon\s?$/);
```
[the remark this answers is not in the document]
frankops11: Appreciate your insights! 🚀 I opted for expect(showIconClass).toMatch(/^react-datepicker__calendar-icon\s?$/) over using trim() because it directly allows for potential, intentional trailing spaces, signaling an openness for additional class names. It kind of sets a clear expectation, without assuming unnecessary whitespaces.

In regards to in-code comments, I believe this PR and the associated discussion provide a decent historical record for future code wanderers. However, if inline comments are the norm here, happy to adapt! 😄

src/calendar_icon.jsx
frankops11: @martijnrusschen 
Thank you for pointing that out! 🙌 I've gone ahead and updated the PR description to encompass all the adjustments and their respective rationales. Apologies for any confusion caused and I appreciate your vigilance in ensuring all changes are accurately documented and justified! 🚀

Thanks again for your time and keen eye! 🧐🚀

martijnrusschen (approved):

Issue #4300, opened 2023-10-09
Rewrite usage of ReactDOM.render to React 18


martijnrusschen:
> ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it’s running React 17.

Blocking https://github.com/Hacker0x01/react-datepicker/pull/4207 currently. See https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis for guidance on how to use the new render.

martijnrusschen:
@frankops11 If you're looking for a nice challenge, there are 13 usages across 4 files in the code base. These need to be rewritten to the new React 18 way.

frankops11:
Challenge accepted!🔍🛠️

frankops11:
@martijnrusschen accepting the challenge has been quite an adventure. After making changes in 'docs-site' and 'examples', no issues arose; everything went smoothly. However, when we dove into the universe of unit testing, that's where the house came tumbling down, haha! Given that we are using Enzyme and, to this day, there is no support for React 18 nor anything official about it, many tests simply break, and honestly, I don’t know if you have tried to tackle something similar. One strategy that occurred to me was to migrate some of those tests to Testing Library, but I'd love to hear your thoughts on this.

![image](https://github.com/Hacker0x01/react-datepicker/assets/50973629/998f3c62-d6ce-467e-80cd-3e10d1090493)

martijnrusschen:
It looks like the problematic part is the Enzyme adapter that's specifically for React 17. It looks like Enzyme and Jest go well together though: https://enzymejs.github.io/enzyme/docs/guides/jest.html. Can we get rid of the React 17 adapter?

frankops11:
Thank you for pointing out the issue and providing additional resources 🙏. 

I've explored the [Enzyme's compatibility thread](https://github.com/enzymejs/enzyme/issues/2524#issuecomment-1116246491) and the article ["Enzyme is dead, now what?"](https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl) to grasp the current situation with Enzyme, particularly concerning its compatibility with React 17 and the forthcoming versions.

It seems like the path forward with Enzyme is a bit murky 🌫️, especially considering the move towards `react-testing-library` by many in the community due to the uncertain future of Enzyme. The particular concern with React classes in future versions does pose a valid concern 🚧. 

[...]

PR #4318, opened 2023-10-14
Fix #4317: Fix Enzyme Test file in exclude_times_test.js and migrate it to use @testing-library/react

### LINKED ISSUE #4317: Invalid Test Case Results from Enzyme's.find() - Returns inconsistent Results for <DatePicker /> component in exclude_times_test.test.js
**Problem**
In the `exclude_times_test.test.js` file, we have encountered an issue when using Enzyme's `.find()` method in conjunction with `mount()` to render a `<DatePicker />` component and querying elements by class. The `.find()` method is returning an empty wrapper even when the specified class does not exist, which is leading to inconsistencies in our test results and making it challenging to determine the validity of our test cases.

image

In the above screenshot, the test case should fail, as the `<DatePicker />` is not opened or inline, it won't render any date or time, hence the test case would fail obviously.  But as we're expecting `expect(datePicker.find('xxx')).not.toBeNull()` the test case is getting ignored.

So, in this file, I found 2 issues

1. `<DatePicker />` is not opened, hence the test case fails
2. Using `.not.toBeNull` instead of checking specific assertion like length check

I also found one another important bug, that is related to the use of `setTime` helper while setting `excludeTimes` prop
image

If you notice the above screenshot, we're passing invalid argument to the `setTime` helper.  For example, instead of `hour`, we are passing `hours` and instead of `minute` we're passing `minutes`.

**Proposed Solution**

- Fix the `setTime` call
- Open the `<DatePicker />` component to test it, by passing the `open` prop
- Use .`toHaveLength(4)` instead of using `.not.toBeNull`

balajis-qb:
Closes #4317

### Summary
This PR addresses the issues in the test file `exclude_times_test.js` and also migrated the test cases to use `@testing-library/react` instead of `enzyme` as requested by @martijnrusschen

### Changes Made

- Fix the `setTime` call
- Open the `<DatePicker />` component to test it, by passing the `open` prop
- Use `.toHaveLength(4)` instead of using `.not.toBeNull`
- Migrated the file to use `@testing-library/react`

### DIFF
--- a/test/exclude_times_test.test.js
+++ b/test/exclude_times_test.test.js
@@ -1,12 +1,13 @@
 import React from "react";
-import { mount } from "enzyme";
+import { render } from "@testing-library/react";
 import { setTime, newDate } from "../src/date_utils";
 import DatePicker from "../src/index.jsx";
 
 describe("DatePicker", () => {
   it("should disable times specified in excludeTimes props", () => {
-    var now = newDate();
-    var datePicker = mount(
+    const now = newDate();
+
+    const { container: datePicker } = render(
       <DatePicker
         open
         showTimeSelect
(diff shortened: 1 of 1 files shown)


martijnrusschen:
Nice that was quick! looking great :) We have more usage of Enzyme, if you're up feel free to remove those as well. We need to remove Enzyme before we can safely migrate to the React 18 render methods.

martijnrusschen:
More info can be found here: https://github.com/Hacker0x01/react-datepicker/issues/4300

PR #4336, opened 2023-10-23
Fix #4170: Clear the previously selected date on empty input with showTimeSelectOnly

### LINKED ISSUE #4170: DatePicker with "showTimeSelectOnly" is not possible clear the value
**Describe the bug**
When the datepicker is configured with property "showTimeSelectOnly" it's not possible to clear the value

**To Reproduce**

0. Go to the page 'https://reactdatepicker.com/' and find the "Select Time Only" example (https://reactdatepicker.com/#example-select-time-only)
1. Click on the datepicker
2. Clear the value in the input (backspace or delete on the keyboard)
3. Click outside
4. The previous value reappears in the input

**Expected behavior**
The datepicker should be cleared, withuout value

**Screenshots**
Step 1
immagine
Step 2
immagine
Step 3
immagine

- OS: Windows 10
- Browser Chrome
- Version 115.0.5790.102 (Official Build) (64-bit)
- Version ReactDatepicker 4.16.8

**Additional context**
I tryed in a previous versione of react-datepicker (4.8.0) and it worked, so i think is a regression

balajis-qb:
Closes #4170

### Summary
This PR addresses an issue where the DatePicker was not clearing the previously selected time when the user deletes the value in the date input via keyboard.  This issue was only coming when the `showTimeSelectOnly` is enabled.

### Changes Made:

- I updated the new date value only if the value exist, else set the null value, instead of restoring back to the previously selected value
- I added test cases to validate the above case both with and without `showTimeSelectOnly` using React Test Library

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -542,21 +542,14 @@ export default class DatePicker extends React.Component {
     if (
       this.props.showTimeSelectOnly &&
       this.props.selected &&
+      date &&
       !isSameDay(date, this.props.selected)
     ) {
-      if (date == null) {
-        date = set(this.props.selected, {
-          hours: getHours(this.props.selected),
-          minutes: getMinutes(this.props.selected),
-          seconds: getSeconds(this.props.selected),
-        });
-      } else {
-        date = set(this.props.selected, {
-          hours: getHours(date),
-          minutes: getMinutes(date),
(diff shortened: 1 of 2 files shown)


martijnrusschen:
Nice refactor and fix

PR #4346, opened 2023-10-30
Add Enzyme deprecation linter to avoid more usage of Enzyme

martijnrusschen:
Relates to https://github.com/Hacker0x01/react-datepicker/issues/4300

### DIFF
--- a/test/calendar_icon.test.js
+++ b/test/calendar_icon.test.js
@@ -14,6 +14,7 @@ afterAll(() => {
 describe("CalendarIcon", () => {
   it("renders a custom SVG icon when provided", () => {
     const wrapper = mount(
+      // eslint-disable-line enzyme-deprecation/no-mount
       <CalendarIcon showIcon icon={<IconParkSolidApplication />} />,
     );
     expect(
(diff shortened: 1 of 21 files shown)


test/calendar_icon.test.js
```diff
@@ -14,6 +14,7 @@ afterAll(() => {
 describe("CalendarIcon", () => {
   it("renders a custom SVG icon when provided", () => {
     const wrapper = mount(
+      // eslint-disable-line enzyme-deprecation/no-mount
```
Zarthus, line 17: did you mean to move this one line up?

Zarthus (approved):
The actual changes (eslintrc and package.json) look good to me, it seems that at a few places GitHub disagrees with the linter and I think the line has to be moved up one line, but at the same time it's also working fine for other LOCs, it seems like the differentiator is whether it's assigned to a variable by a quick look.

test/calendar_icon.test.js
martijnrusschen: It looks like somehow prettier is moving things to a new line. Not sure if there's a good workaround for this.

PR #4373, opened 2023-11-12
Fix: reflect the `holidays` prop change

### LINKED ISSUE #4372: Changes to the `holidays` prop is not reflected to the selector of day
**Describe the bug**
Once the datepicker is rendered without the `holidays` prop, change the `holidays` prop and re-render the datepicker. At this time, I want the datepicker to reflect the changed holidays, but it does not.
As an example, I'm assuming a case where I render the datepicker with `holidays === undefined` and then resolve `holidays`  asynchronously .

**To Reproduce**
https://codesandbox.io/s/charming-tristan-6cgh9s?file=/src/App.js

**Expected behavior**
The changed `holidays` should be reflected and the `--holidays` modifier should be given to day cells.

**Screenshots**
スクリーンショット 2023-11-12 10 08 37

- OS: macOS Sonoma 14.0
- Browser: Arc (Chromium) (1.15.2), Chrome (119.0.6045.123), Safari (17.0)

Not tried.

shimech:
## Issue
The `holidays` prop is not reflected even if it is updated after initial rendering, because `holidays` is managed by state, and it is set at only initial rendering.
fix #4372 

## Fixing Approach
Calculate `holidays` at every rendering instead of managing it in state

## Cons
If very large `holidays` is given, rendering cost will increase.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -362,20 +362,18 @@ export default class DatePicker extends React.Component {
       ? this.props.endDate
       : newDate();
 
+  // Convert the date from string format to standard Date format
+  modifyHolidays = () =>
+    this.props.holidays?.reduce((accumulator, holiday) => {
+      const date = new Date(holiday.date);
+      if (!isValid(date)) {
+        return accumulator;
+      }
+
+      return [...accumulator, { ...holiday, date }];
+    }, []);
+
   calcInitialState = () => {
-    // Convert the date from string format to standard Date format
-    const modifiedHolidays = this.props.holidays?.reduce(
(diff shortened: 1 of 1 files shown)


shimech:
Thank you for your review.
Is there anything left for me to do to get this PR merged?

TinnesWasTaken:
Hi. Is this change in the latest npm package (4.23.0)? Looks it was merged in the same day as the package was released, but I can't find a change log and I'm still getting the issue. It would be good to confirm it's if just not published yet or if I'm doing something wrong! Thanks

PR #4384, opened 2023-11-21
containerRef div shouldnt affect styling

joaopaulo-capy:
Currently the calendar is being rendered wrapped in a div that the sole intent, as i understand, is to wrap a ref arround it.
Considering this and that there are no props to style that div, it should have the display contents property to not affect styling. In some very specific cases it can create issues, for example in the print below, there is a little gap under the calendar*

*Maybe if i used height 100% in the calendar style it would fix it, however it could introduce other issues in the way i want the style to behave.

image

With a flex parent:

image
image

### DIFF
--- a/src/calendar.jsx
+++ b/src/calendar.jsx
@@ -1074,7 +1074,7 @@ export default class Calendar extends React.Component {
   render() {
     const Container = this.props.container || CalendarContainer;
     return (
-      <div ref={this.containerRef}>
+      <div style={{display: 'contents'}} ref={this.containerRef}>
         <Container
           className={classnames("react-datepicker", this.props.className, {
             "react-datepicker--time-only": this.props.showTimeSelectOnly,


sebastienbarre:
@joaopaulo-capy -- This definitely had unfortunate negative side effects on inline calendars that I don't know how to fix.

Before:

image

After (because of `display: 'contents'`):

image

joaopaulo-capy:
Im sure this can be fixed easily. What kind of container is your inline calendar? Does your calendar have any styling? and is the expected behaviour your first print?

Here are some experiments:

My datepicker from the original post is also inline and just works

image

However if i set a width to the container  (in this example 100% width), your bug is reproduced

image
image

This is kind of expected considering the month container is not responsive

image

Here is month container with 100% width (still not the best solution because the day grid is not responsive and it would be hard to make it responsive, considering we would need to calculate the font-size responsively)

image
image

Does wrapping your calendar arround a container (without any width style) fix your problem?

sebastienbarre:
Yes, wrapping in a `div` does fix it, but this is still a breaking behavior for everybody else, right? You would need people who have an inline calendar to fix it too, as opposed to use `display: contents` only when the `inline` prop is not true in the first place. Thanks.

PR #4469, opened 2024-01-24
Fix options passed to date-fns/format

emilecantin:
`awareOfUnicodeTokens` has been removed in https://github.com/date-fns/date-fns/pull/1069


emilecantin:
@martijnrusschen Would you mind releasing a new version?

Looks like the last one is about a month old, and there have been a few changes. Also, linking directly to the git version doesn't work as it's not built.

martijnrusschen:
New version has been released.

emilecantin:
I just saw that, thank you very much!

PR #4481, opened 2024-01-29
Upgrade date-fns to v3

ethanve:
Upgrades `date-fns` to their latest version.

Changes that were needed:
1. All default exports were removed.
2. Intervals now normalize if the start date is after the end date


test/date_utils_test.test.js
```diff
@@ -249,15 +249,15 @@ describe("date_utils", () => {
       ).toBe(true);
     });
 
-    it("should throw error if excluded date interval is invalid", () => {
+    it("should be enabled and normalize negative intervals correctly", () => {
```
ethanve, line 252: [See date-fns' release](https://github.com/date-fns/date-fns/releases/tag/v3.0.0):
>  Functions that accept Interval arguments now do not throw an error if the start is before the end and handle it as a negative interval. If one of the properties in an Invalid Date, these functions also do not throw and handle them as invalid intervals.

martijnrusschen:
This looks good, are there any breaking changes that require a major version bump of the package?

ethanve:
@martijnrusschen I believe this should be a major version bump given the breaking changes in date-fns. This will not work with date-fns v2.

PR #4498, opened 2024-02-02
Fix - Removed defaultProps from Function Components for React 18.3.

### LINKED ISSUE #4477: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.

shyk001:
As of react 18.3, a warning will be displayed when using defaultProps inside a function component. Therefore, I have removed the defaultProps used in the function component.

Fix #4477

### DIFF
--- a/src/with_floating.jsx
+++ b/src/with_floating.jsx
@@ -25,22 +25,29 @@ export const popperPlacementPositions = [
 
 export default function withFloating(Component) {
   const WithFloating = (props) => {
+    const alt_props = {
+      ...props,
+      popperModifiers: props.popperModifiers || [],
+      popperProps: props.popperProps || {},
+      hidePopper:
+        typeof props.hidePopper === "boolean" ? props.hidePopper : true,
+    };
     const arrowRef = React.useRef();
     const floatingProps = useFloating({
-      open: !props.hidePopper,
+      open: !alt_props.hidePopper,
       whileElementsMounted: autoUpdate,
-      placement: props.popperPlacement,
+      placement: alt_props.popperPlacement,
       middleware: [
         flip({ padding: 15 }),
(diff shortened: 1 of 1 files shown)


reinrl:
Is the extra `typeof` check in the line `hidePopper: typeof props.hidePopper === "boolean" ? props.hidePopper : true` because otherwise with this manner you couldn't ever set `hidePopper` to `false`? Why not just use nullish coalescing (e.g., `hidePopper: props.hidePopper ?? true`), or just create a default props object within the component body and then layer the incoming props in on top of that?

```js
const defaultProps = {
  popperModifiers: [],
  popperProps: {},
  hidePopper: true,
};
const alt_props = {
  ...defaultProps,
  ...props,
};
```

shyk001:
@reinrl
I think it would have been good code to use nullish coalescing like you said, I missed it. Thanks for the advice.

For the second method, I don't think it will work as you intended because the props will overwrite the default_props.

```js
const props = {
    a: "1qe",
    b: "asdf",
    c: undefined,
    d: {a: "gkgk"},
    e: undefined
}

const default_props = {
    c: [],
    d: {},
    e: true,
}

const alt_props = {
    ...default_props,
    ...props
}

# alt_props console log result
# { c: undefined, d: { a: 'gkgk' }, e: undefined, a: '1qe', b: 'asdf' }
```

reinrl:
@shyk001 good catch - explicitly passing in undefined for any of the props was an edge case that I hadn't considered. Just took me longer than I care to admit to figure out the "why" behind the Boolean type check in your change...and then once I figured it out, I started thinking through other alternatives I have used elsewhere (and your point is a good one that is going to make me revisit some past code elsewhere).

PR #4533, opened 2024-02-26
Fix #4483: Enable onKeyDown handler for the month picker view and the year picker view

### LINKED ISSUE #4483: onKeyDown function is not triggered in Month and Year Pickers.
**Describe the bug**
onKeyDown function is not triggered in Month and Year Pickers. It is triggered only in a day selection view.

**Expected behavior**
Function should be triggered in any view.

- OS: Windows 11
- Browser Chrome
- Version 120.0.6099.225

balajis-qb:
Closes #4483

**Description**
This PR addresses the missing event handling to handle the key-press event for the month picker and the year picker.  For the Year Picker component, I added `onKeyDown` and mapped it to the corresponding prop.  However, the MonthPicker (`month.jsx`) is already receiving `onKeyDown` event, but that was mapped to the handler `handleOnDayKeyDown` which is the handler specifically we're using on `day.jsx`.  In `month.jsx` we're not directly using it, but passing that to the `Week` component.  Hence for the month.jsx alone, I didn't change any of its current behavior, and I added a new prop named `handleOnMonthKeyDown` which is linked to the `onKeyDown` event handler we pass from the DatePicker component.  So users can still use `onKeyDown` itself for any datepicker view (day/month/year), but internally `month.jsx` alone is handling it via `handleOnMonthKeyDown`.

**Changes Made**
- Updated Year picker and Month picker to handle the `onKeyDown` event.
- Updated the corresponding test cases.

### DIFF
--- a/test/month_test.test.js
+++ b/test/month_test.test.js
@@ -1,5 +1,5 @@
 import React from "react";
-import ReactDOM from "react-dom";
+import { render, fireEvent } from "@testing-library/react";
 import Month from "../src/month";
 import Day from "../src/day";
 import DatePicker from "../src";
@@ -941,7 +941,7 @@ describe("Month", () => {
     it("should call onKeyDown handler on any key press", () => {
       const onKeyDownSpy = jest.fn();
 
-      const datePicker = TestUtils.renderIntoDocument(
+      const { container } = render(
         <DatePicker
           selected={new Date()}
           dateFormat="MM/yyyy"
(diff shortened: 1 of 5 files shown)


test/month_test.test.js
```diff
@@ -935,6 +937,29 @@ describe("Month", () => {
...
+
+      const datePicker = TestUtils.renderIntoDocument(
+        <DatePicker
+          selected={new Date()}
+          dateFormat="MM/yyyy"
+          showMonthYearPicker
+          onKeyDown={onKeyDownSpy}
+        />,
+      );
+      TestUtils.Simulate.focus(ReactDOM.findDOMNode(datePicker.input));
```
martijnrusschen, line 952: I believe this is deprecated behavior. Can you use the new RTL library?

test/month_test.test.js
```diff
@@ -935,6 +937,29 @@ describe("Month", () => {
...
+
+      const month = TestUtils.findRenderedDOMComponentWithClass(
```
martijnrusschen, line 954: Same

martijnrusschen:
Thanks for the fix!

test/month_test.test.js
balajis-qb: Updated

PR #4542, opened 2024-02-28
fix: Change onMouseXXX event to onPointerXXX event

yuki0410-dev:
Thank you for allowing us to use your wonderful library.

I found a bug when using the `selectsRange` props.
When I select startDate and then select a date before that date, I get the `react-datepicker__day--in-selecting-range` style for a moment.
This occurs on touch devices and on the desktop Chrome touch device simulation.
It also seems to occur more noticeably when using `monthsShown` props and many calendars in a row.

[code sandbox](https://codesandbox.io/p/sandbox/react-datepicker-df2zky)

I think that onPointerXXX events should be used rather than onMouseXXX events to more accurately capture events on touch devices.
(By using `onPointerXXX events`, the earlier decrease no longer occurs.)

- [mdn](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events)
- [camiuse](https://caniuse.com/pointer)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -74,18 +74,18 @@
 | `onChange`                      | `func`                         | `() {}`                    |             |
 | `onChangeRaw`                   | `func`                         |                            |             |
 | `onClickOutside`                | `func`                         | `() {}`                    |             |
-| `onDayMouseEnter`               | `func`                         |                            |             |
+| `onDayPointerEnter`             | `func`                         |                            |             |
 | `onFocus`                       | `func`                         | `() {}`                    |             |
 | `onInputClick`                  | `func`                         | `() {}`                    |             |
 | `onInputError`                  | `func`                         | `() {}`                    |             |
 | `onKeyDown`                     | `func`                         | `() {}`                    |             |
 | `onMonthChange`                 | `func`                         | `() {}`                    |             |
-| `onMonthMouseLeave`             | `func`                         |                            |             |
+| `onMonthPointerLeave`           | `func`                         |                            |             |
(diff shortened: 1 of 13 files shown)


yuki0410-dev:
thanks for the review.

---

> This change looks good to me by using `pointer` event to replace `mouse` events. The only concern I have is regarding legacy browser/device compatibility/support.

> My only concern with this PR is just that it seems to be changing the existing interface. Doing so would likely require a release with a version bump to the major version. My suggestion is to instead make these changes backward-compatible by maintaining the existing interface.

Thank you for your feedback on legacy browser support and backward compatibility of props.
For example, what do you think of the policy of adding `usePointerEvents (bool) props` (default: false) and binding the handler to onPointerXXX Event instead of onMouseXXX Event only if true?

---

> It is also worth noting that switching to pointer events, specifically `pointerdown` events, can cause different behavior that is well-defined, but often not expected, compared to `mousedown` events.
> 
> This is referred to as implicit pointer capture and is covered in the [spec](https://www.w3.org/TR/pointerevents/#implicit-pointer-capture).
> 
> This second point is not actionable for this PR, but should be kept in mind and referenced if any further migration to pointer-based events occurs.

Certainly we will need to be careful when adding new Event handlers in the future.
However, the currently used onXXXEnter and onXXXLeave appear to work fine with Pointer.

yuki0410-dev:
I close this PR because I consider the other [PR](https://github.com/Hacker0x01/react-datepicker/pull/4553) more useful.
Thanks for taking your guys time to review!

PR #4548, opened 2024-02-29
Corrected Validation OpenToDate

RobertAlmeida:
Add a validation "openToDate" before access property


RobertAlmeida:
When called without any parameters it assumes the current date.

PR #4552, opened 2024-02-29
Fix: default time selection

kiselevskym:
1. Datepicker with `showTimeSelect` and `minTIme`.
2. The datepicker should be empty.
3. Select any day you want .

Actual Result (AR):
Despite disabling the start time, the Datepicker defaults to selecting the beginning of the day.

Expected Result (ER):
The Datepicker should select the minimum allowable time upon choosing a day.

image
image

### DIFF
--- a/test/min_time_test.test.js
+++ b/test/min_time_test.test.js
@@ -1,7 +1,6 @@
 import React, { useState } from "react";
 import DatePicker from "../src/index.jsx";
-import { mount } from "enzyme";
-import Day from "../src/day.jsx";
+import { fireEvent, render } from "@testing-library/react";
 
 const DatePickerWithState = (props) => {
   const [selected, setSelected] = useState(null);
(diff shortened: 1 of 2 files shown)


martijnrusschen:
Please make sure to add a test case to check if the min time gets set

kiselevskym:
> Please make sure to add a test case to check if the min time gets set

Test cases were added

test/min_time_test.test.js
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from "react";
+import DatePicker from "../src/index.jsx";
+import { mount } from "enzyme";
```
martijnrusschen, line 3: Can you avoid using enzyme? You can use React Testing Library.

kiselevskym: no prob, can you check it now?

kiselevskym: @martijnrusschen

PR #4559, opened 2024-03-05
fix: enzyme to RTL (some files)

yuki0410-dev:
When we heard that the new tests were going to use RTL instead of enzyme, we decided that we needed to rewrite the existing tests with RTL.
Since I do not have the time to change everything at once, I created this PR to try to change things little by little.

(Sorry if there is an existing issue, PR.)


martijnrusschen:
This is great, and appreciate the effort! Ideally we would remove all enzyme usage from this repo.

martijnrusschen (approved):

PR #4553, opened 2024-03-01
feat : add usePointerEvent props

yuki0410-dev:
I want to fix a [bug](https://github.com/Hacker0x01/react-datepicker/pull/4542#issue-2159347503) I reported in a previous PR,
Inspired by the [comments](https://github.com/Hacker0x01/react-datepicker/pull/4542#discussion_r1506667247) in the previous PR, we have modified the policy to allow users to freely choose by Props.

PR #4575, opened 2024-03-10
feat: run prettier check in ci

yuki0410-dev:
Since the prettier execution affects the Codecov Report, we thought it should be enforced by CI.

-- Background --
I did not run prettier on #4553.
When I ran prettier on another PR, a line break occurred and Codecov Report's CI was NG, so I needed to add a test.
https://github.com/Hacker0x01/react-datepicker/pull/4574/commits/dec832a628176f5f513113f8fb6082ed10f650de

This case was not a problem because it was code I created myself, but I don't think well if it was code created by someone else.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,6 +41,9 @@ jobs:
       - name: Install dependencies
         run: yarn install
 
+      - name: Check Code Formatter
+        run: yarn prettier:check
+
       - name: Lint
         run: yarn lint
 
--- a/package.json
+++ b/package.json
@@ -117,6 +117,7 @@
     "sass-lint": "stylelint 'src/stylesheets/*.scss'",
     "lint": "yarn run eslint && yarn run sass-lint",
     "prettier": "prettier --write '**/*.{js,jsx}'",
+    "prettier:check": "prettier --check '**/*.{js,jsx}'",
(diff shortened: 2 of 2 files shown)


yuki0410-dev:
The fix is reflected in the CI for this PR, which is running without problems.
https://github.com/Hacker0x01/react-datepicker/actions/runs/8221884722/job/22482785292?pr=4575

.github/workflows/test.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         run: yarn install
 
       - name: Lint
-        run: yarn lint
+        run: yarn lint && yarn prettier:check
```
martijnrusschen, line 45: Alternatively, you could split this to a separate step after the linter for more clarity.

martijnrusschen:
Agree, I think this has caused issues in the past and would be good to enforce.

.github/workflows/test.yml
yuki0410-dev: @martijnrusschen 

Thanks for the prompt review.
Added as another step.
```
      - name: Check Code Formatter
        run: yarn prettier:check
```
https://github.com/Hacker0x01/react-datepicker/actions/runs/8221991678/job/22483001616?pr=4575

Issue #4587, opened 2024-03-12
Add .d.ts


yuki0410-dev:
Thanks for the continuous OSS development.

**Is your feature request related to a problem? Please describe.**
I use react-datepicker along with typescript in my work. However, @types/react-datepicker is currently in a separate repository, so new features of react-datepicker are not immediately available when they are released.

**Describe the solution you'd like**
I would like to distribute d.ts files in this repository so that the latest react-datepicker functionality is available in typescript without @types/react-datepicker.

**Describe alternatives you've considered**
Wait for the latest @types/react-datepicker to be released. Or create a PR.
We also considered rewriting this entire repository to TS, but due to the scale of the project, we felt it would not be practical.

**Additional context**
We need to introduce a check of d.ts by CI with reference to @types/react-datepicker.

yuki0410-dev:
refs #4000

yuki0410-dev:
refs #4700

PR #4592, opened 2024-03-14
add types

yuki0410-dev:
## Description
**Linked issue**: #4587

**Problem**

**Changes**
- Based on [@types/react-datepicker](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-datepicker), incorporating changes from v6.3.0 
    - add types (d.ts)
    - add test
    - add CI
    - fix md

## To reviewers
Merging and releasing this PR will allow react-datepicker type completion to work without installing @types/react-datepicker.
You will also see the TS icon (painted blue) on npmjs.com.

## Contribution checklist


yuki0410-dev:
Thanks for the review.

> My only other suggestion would be to find any instances in the README or elsewhere (e.g., documentation) that involve notes on contributing, and mention that TypeScript types should be updated when making relevant changes. A good example of this is actually your recent PR (https://github.com/Hacker0x01/react-datepicker/pull/4553).

I agree and will edit the md file.

> The primary maintainers of this repository will need to have the final say on supporting TypeScript at this level. I think it seems like a beneficial addition to the project, but we will have to wait for their feedback before finalizing this.

I agree with you.
I would be happy if the main maintainer of this repository would understand the advantages and disadvantages of this PR and adopt it.

martijnrusschen:
@yuki0410-dev I'm not an expert, but I understand the benefits of supporting this. Is there any reason why these definitions are also maintained in a separate repo like you linked? I guess that's the same of what we're now adding in this repo.

yuki0410-dev:
@martijnrusschen 
Thank you for your understanding regarding TS support.

When this PR is released, users of later versions will no longer need to use @types/react-datepicker. However, the current version of @types/react-datepicker must be retained for users of versions prior to that release. (No further updates will be made.)

Initially there is no difference between the two repositories, but each time a parameter such as Props is added in this repository, it will be reflected in the types in this repository, and the difference between the two repositories will become larger and larger.

In this case, we are adding the type usePointerEvent as an example.
https://github.com/Hacker0x01/react-datepicker/pull/4592/commits/363cbb6b6abeb96e15d18f70e91d4353ddfc973a

martijnrusschen:
Thanks, will wait for the recommended updated.

yuki0410-dev:
@martijnrusschen 
Thank you for your understanding.
I will make the corrections pointed out by linter and add the documentation.

yuki0410-dev:
I did CI tweaks, Lint error handling, and md file corrections.

yuki0410-dev:
Sorry, but I'm going to have to rethink the idea once and I'll make it a draft.

mirus-ua:
This is a bold idea, but maybe we should consider adopting TS instead of just merging the repo with `.d.ts`? 

This is almost impossible to do at once, but with a proper community road map, we can do it for all files individually. Moreover, the most important part is the datepicker itself, and docs and even tests can stay written in js for any amount of time

martijnrusschen:
Yes agree, I'm open to it but don't have the time to get this going. Open to suggestions.

mirus-ua:
What path I see:
- create an issue similar to the mui team https://github.com/mui/material-ui/issues/41273
- prepare rollup and the rest tools to work with typescript files 
- from my experience, we should adapt ts from small modules to the bigger ones

mirus-ua:
I can prepare a draft of the migration issue

martijnrusschen:
@mirus-ua sounds good!

mirus-ua:
@martijnrusschen here is the draft, let's discuss it inside the issue https://github.com/Hacker0x01/react-datepicker/issues/4700

yuki0410-dev:
#4700 is supported so that the type definitions will be output, and this PR will be closed.

PR #4616, opened 2024-03-19
add JSDoc (same files)

yuki0410-dev:
## Description

**Problem**
Type completion does not work with jsx, and with huge Component, the type of Porps and State (null or undefined) cannot be guessed, which is not a good prospect.
By adding JSDoc correctly, type completion will work at least on VSCode, which we believe will improve development efficiency and reduce the incidence of defects due to omissions of empty checks.

**Changes**
We have included JSDoc in two files as a sample.
I would like to hear your opinion on whether it is worth expanding to other files.

If all files are to be replaced with ts/tsx, it is better to start with JSDoc, which is easier to install, because of the large impact of rollup setting changes, etc.

## Contribution checklist

### DIFF
--- a/src/month_dropdown_options.jsx
+++ b/src/month_dropdown_options.jsx
@@ -1,6 +1,19 @@
+//@ts-check
 import React from "react";
 import PropTypes from "prop-types";
 
+/**
+ * @typedef {Object} Props
+ * @property {number} month
+ * @property {VoidFunction} onCancel
+ * @property {(month:number) => void} onChange
+ * @property {string[]} monthNames
+ */
+
+/**
+ * @class
+ * @extends {React.Component<Props, {}>}
+ */
 export default class MonthDropdownOptions extends React.Component {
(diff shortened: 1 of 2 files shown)


martijnrusschen:
I wonder if this is the right approach, especially with the other PR that adds the types to this repo. What alternatives do we have? I understand converting to TypeScript is a big effort, but would that be a better path forward?

Furthermore, there's a script that generates static markdown docs for each file already on each release. See https://github.com/Hacker0x01/react-datepicker/tree/main/docs

yuki0410-dev:
@martijnrusschen 
Thanks for the review.

> Furthermore, there's a script that generates static markdown docs for each file already on each release. See 
> https://github.com/Hacker0x01/react-datepicker/tree/main/docs

The documentation seems to be generated based on PropsType, but I found it lacking because there is no enforcement like TS in the source code and no mention of State.

I will consider another approach.

PR #4632, opened 2024-03-23
feat: #3798 Replaced classnames with clsx

### LINKED ISSUE #3798: Replace clasnames with clsx
Could you please consider replacement of classnames library with clsx?
After react-grid-layout did the replacement react-datepicker is the last library in my dependencies that still uses classnames.

"Now, as a free tip to at least 2x the performance of the thousands of classNames calls you probably make, just switch to clsx. It’s a similar library to classNames, but a bit more fresh-y, and with even fewer LOCs.

The API is identical to the classNames one, but without the additional versions (which you probably don’t need anyway). And while the performance gains might not be noticeable, it still means that there’s speed left on the table, and the multitude of calls can quickly add up to something more."

https://areknawo.com/you-dont-know-the-classnames-library/

yuki0410-dev:
## Description
**Linked issue**: close #3798

**Problem**
See issue #3798

**Changes**
Replaced classnames with clsx

## Contribution checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
   },
   "dependencies": {
     "@floating-ui/react": "^0.26.2",
-    "classnames": "^2.2.6",
+    "clsx": "^2.1.0",
     "date-fns": "^3.3.1",
     "prop-types": "^15.7.2",
     "react-onclickoutside": "^6.13.0"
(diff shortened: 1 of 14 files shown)


yuki0410-dev:
Thanks for the review.

I removed classnames from global in rollup.config.mjs.
The clsx did not affect the build result whether it was present in globals or not.

martijnrusschen:
Thanks this looks good.

PR #4643, opened 2024-03-26
fix: #4558 Rename calendarIconClassname props to calendarIconClassName props

### LINKED ISSUE #4558: Property naming confusion for class names
**Describe the bug**
some class name properties is named xyzClassName, while others are called xyzClassname, this creates a bit of confusion when trying to override styling.

Examples: 
- calendarIconClassname
- DayClassName

yuki0410-dev:
## Description
**Linked issue**: close #4558

**Problem**
See issue #4558

**Changes**
fix props : calendarIconClass`n`ame => calendarIconClass`N`ame

## Contribution checklist


yuki0410-dev:
@martijnrusschen 
According to the reasons given in the issue.

> some class name properties is named xyzClassName, while others are called xyzClassname, this creates a bit of confusion when trying to override styling.

yuki0410-dev:
I tried to make the old props apply if you are not using the new props.
I also tried to output console.warn if you are using the old properties, what do you think?

yuki0410-dev:
Thanks for the review.

> I'm not certain what the project's guidelines are around console statements since they are usually removed for production code.

Agreed. However, since console.warn is often not deleted in development builds, I think it makes sense to leave the message in console.warn.

> I think some sort of notice of deprecation in the README would be nice or even removing the calendarIconClassname in the README and let it silently work for existing projects but ensure future projects only use calendarIconClassName.

> I agree, I have marked it as deprecated in datepicker.md, but not in index.md, as it will probably be overwritten by auto-generation, so no details are given.

> These are likely decisions to be made by the project owner though since there are many different ways to accomplish deprecating fields.

Agreed.

@martijnrusschen
What do you think about this PR response?

martijnrusschen:
There are currently 3 PRs in the list that contain breaking changes. Happy to make a major version upgrade if we can get all 3 merged.

martijnrusschen:
@yuki0410-dev Can you merge the latest main branch to make sure this PR is still compatible?

yuki0410-dev:
@martijnrusschen 
I have imported the latest master and the CI results seem to be fine.

PR #4689, opened 2024-04-15
fix: popper dosen't close if swapRange prop was passed

mirus-ua:
---
name: Fix input behavior
about: input doesn't close popper if swapRange prop passed 
---

**Problem**
After [the 6.7.0 release](https://github.com/Hacker0x01/react-datepicker/releases/tag/v6.7.0), I found that the new `swapRange` feature doesn't allow to close datepicker popper automatically when the range selected in reverse order

**Changes**
Add swapRange into the related check + a new test case

## Screenshots
The first video is before the fix

https://github.com/Hacker0x01/react-datepicker/assets/29524085/49ac07c7-106d-458c-8e79-91e4111f341a

The second one is after the fix

https://github.com/Hacker0x01/react-datepicker/assets/29524085/a0561d94-f654-480e-9c16-862b5e8a96f4

## Contribution checklist

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -602,7 +602,11 @@ export default class DatePicker extends React.Component {
 
       const { startDate, endDate } = this.props;
 
-      if (startDate && !endDate && !isDateBefore(date, startDate)) {
+      if (
+        startDate &&
+        !endDate &&
+        (this.props.swapRange || !isDateBefore(date, startDate))
+      ) {
         this.setOpen(false);
       }
     }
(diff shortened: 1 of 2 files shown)


mirus-ua:
> I believe these changes to be correct, but I question if the provided test already passes without the changes? It is not necessary to choose a different end date to exercise the new code?
> 
> Image of Graham C **Graham C**
> 

The original code swaps the dates on the final check and doesn't provide a possibility of handling`[null, endDate]` state
https://github.com/Hacker0x01/react-datepicker/pull/4654/files#diff-ec767e14d91498b5b31a4ce826c4596061eb83a47d3dee00f6cbe1078232a75cR697

PR #4702, opened 2024-04-17
feature: add aria-disabled to Month component

MatteoPieroni:
## Summary
This PR addresses an issue where the Month component is missing the attribute aria-disabled on its disabled items.

## Changes made
- Updated the Month component to add the aria-disabled attribute over its disabled items
- Made the logic to find the disabled month unit shareable in a reusable helper function
- Updated the corresponding test cases to also check for the existence of the aria-disabled attribute along with the classname

### DIFF
--- a/src/month.jsx
+++ b/src/month.jsx
@@ -545,31 +545,27 @@ export default class Month extends React.Component {
     }
   };
 
+  isMonthDisabled = (month) => {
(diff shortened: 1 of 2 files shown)


MatteoPieroni:
Thank you for your reviews and having merged this

Issue #4700, opened 2024-04-17
[milestone] Convert the library to typescript


mirus-ua:
# Contribution is welcome 

### Why? 

Everything started [here](https://github.com/Hacker0x01/react-datepicker/pull/4592) from the willing to merge a separate package with the corresponding type definitions into this repo. 
The idea is great, but we, as a community, can go further and migrate the entire code base into typescript. 

### Any benefits?
- only one package to install for end users
- easier to maintain
- your help to the lovely library
- ??? 
- profit

## Our steps to achieve the goal 

- [x] extend rollup and rest tools to support `.ts` and `.tsx` modules https://github.com/Hacker0x01/react-datepicker/pull/4704
- [x] migrate datepicker modules first
- [x] add a GitHub action to check types @mirus-ua https://github.com/Hacker0x01/react-datepicker/pull/4749
- [ ] update docs
- [x] migrate test cases
- [x]  use `ts-jest` @mirus-ua 
- [ ] migrate docs (optional)

### Basic rules and advice:
- start with the smallest isolated modules, like `calendar_container.jsx`
- prevent the use `any` or other loose types if you don't have strong arguments
- prevent using "[slow types](https://jsr.io/docs/about-slow-types)", give this option to the end users 
- use interfaces instead of types because the large amount of the code base is class-based
- add JSDocs where you can to improve overall DX. Everyone likes a good DX

### Steps to contribute

1. Write in the comments the name of the module that you want help to migrate 
2. Me or a maintainer will assign your name to the list to prevent doing the same job twice 
3. Migrate according to the list of advice above
4. Highly recommended to migrate test cases together with the module
5. Create a PR with a name format `[typescript-migration] calendar.jsx` etc.

- [x] calendar_icon.jsx https://github.com/Hacker0x01/react-datepicker/pull/4706
- [x] calendar_container.tsx https://github.com/Hacker0x01/react-datepicker/pull/4704
- [x] date_utils.js https://github.com/Hacker0x01/react-datepicker/pull/4707
- [x] inputTime.jsx https://github.com/Hacker0x01/react-datepicker/pull/4708
- [x] calendar.jsx @yuki0410-dev https://github.com/Hacker0x01/react-datepicker/pull/4766
- [x] popper_component.jsx @yuki0410-dev https://github.com/Hacker0x01/react-datepicker/pull/4740
- [x] week_number.jsx @yuki0410-dev https://github.com/Hacker0x01/react-datepicker/pull/4714
[...]

PR #4749, opened 2024-04-30
[typescript-migration] add type-check to gh actions

mirus-ua:
---
name: add type-check to gh actions
---

## Description
**Linked issue**: https://github.com/Hacker0x01/react-datepicker/issues/4700

**Changes**
I extended the existing Github action called `test.yml` rather than creating a new one because this action already contains the rest of the checks, including eslint and prettier

## Contribution checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Check Code Formatter
         run: yarn prettier:check
 
-      - name: Typecheck
+      - name: Type check
         run: yarn type-check
 
       - name: Lint
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Check Code Formatter
         run: yarn prettier:check
 
-      - name: Type check
+      - name: Type Check
         run: yarn type-check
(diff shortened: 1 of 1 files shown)


mirus-ua:
@yuki0410-dev hello. could you help me and tell if the `DatePicker › should update the preSelection state when a day is selected with mouse click` test is also failing on the latest main branch `0cb6f84237a5db65dc3fb0c454b3ac3a40683f0c`

yuki0410-dev:
@mirus-ua 
I tried it locally and it also failed on the main branch.
As far as I have tried, it occurs when I rewind the PC date by a week or so, which means that the date (end of the month?) is not the same as the date. I think it has something to do with the date (end of the month?) and the day of the week.
I have fixed that error in #4747.

mirus-ua:
> @mirus-ua I tried it locally and it also failed on the main branch. As far as I have tried, it occurs when I rewind the PC date by a week or so, which means that the date (end of the month?) is not the same as the date. I think it has something to do with the date (end of the month?) and the day of the week. I have fixed that error in #4747.

Thank you for your help. So, if you already fixed the problem in your PR, we'll merge your first and then go with this one

PR #4800, opened 2024-05-18
Change the calendar Icon unit test to TS

alvaromartinez986:
---
name: Pull Request
about: Create a pull request to improve this repository
title: "Change the calendar Icon unit test to TS"
labels: ""
assignees: "alvaromartinez986"
---

## Description

**Problem**
Change the unit test to ts, also I noticed the icon prop was not defaulted,so I changed to default

**Changes**

## Contribution checklist

### DIFF
--- a/src/calendar_icon.tsx
+++ b/src/calendar_icon.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 
 interface CalendarIconProps {
-  icon: string | React.ReactNode;
+  icon?: string | React.ReactNode;
   className?: string;
   onClick?: (event: React.MouseEvent) => void;
 }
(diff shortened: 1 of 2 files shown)


src/calendar_icon.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 
 interface CalendarIconProps {
-  icon: string | React.ReactNode;
+  icon?: string | React.ReactNode;
```
martijnrusschen, line 4: Any reason why this is now optional?

alvaromartinez986: > Any reason why this is now optional?

On the calendar_icon.tsx there is a default svg icon on the line 62.... so if the user doesn't pass any icon should be return this icon.... therefore this parameter should be optional to return this svg default icon

PR #4809, opened 2024-05-22
Fix code coverage report

martijnrusschen:
Attempting to fix code coverage report, which was only fetching test files for some reason. Ref #4700


martijnrusschen:
I can see in the output that all files are now considered for the code coverage again. I'm not sure what did the trick, but followed some best practices in terms of config.

PR #4816, opened 2024-05-23
Update codecov config

martijnrusschen:
It seems the previous config was still using a tokenless config


martijnrusschen:
Fixed this based on https://docs.codecov.com/docs/github-2-getting-a-codecov-account-and-uploading-coverage

PR #4891, opened 2024-06-12
[typescript-migration] some tests and configs

yuki0410-dev:
## Description
**Linked issue**: #4700

**Changes**
- migrate src/test/datepicker_test.test.tsx
- migrate src/test/helper_components/custom_input.tsx
- migrate rollup.config.mjs

## Contribution checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,8 +30,6 @@ jobs:
         with:
           path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
           key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-yarn-
 
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v4
--- a/src/test/datepicker_test.test.tsx
+++ b/src/test/datepicker_test.test.tsx
@@ -29,18 +29,15 @@ import CustomInput from "./helper_components/custom_input";
 import TestWrapper from "./helper_components/test_wrapper";
 import { getKey } from "./test_utils";
 
-function getSelectedDayNode(datePicker) {
+function getSelectedDayNode(container: HTMLElement) {
   return (
-    datePicker.calendar &&
-    datePicker.calendar.componentNode.querySelector(
-      '.react-datepicker__day[tabindex="0"]',
-    )
+    container.querySelector('.react-datepicker__day[tabindex="0"]') ?? undefined
   );
 }
 
-function findSelectedDay(datePicker, targetDate) {
+function findSelectedDay(container: HTMLElement, targetDate: Date) {
   const days = Array.from<HTMLElement>(
-    datePicker.querySelectorAll(".react-datepicker__day"),
+    container.querySelectorAll(".react-datepicker__day"),
   );
   return days.find(
     (d) =>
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -30,8 +30,6 @@ jobs:
         with:
           path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
           key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
```
martijnrusschen, line 33: Can you expand on this?

martijnrusschen (commented):
This is a major milestone! I think this marks the complete conversion to TS. Nice work!

.github/workflows/test.yml
yuki0410-dev: @martijnrusschen 
"restore-keys" searches cache by forward matching and uses the latest cache hit.
I have finished the ts-jest migration. (CI was ok).
I removed babel-jest as it is no longer needed. (CI was NG as the version dependency of the module used internally seemed to be broken.)
Drop the commit of ↑ (CI was NG as well as ↑)

From the above, I think it is a good idea to reduce CI time by using close cache as a base for yarn install, though,
However, if a cache with incorrect dependencies remains somewhere, there is a risk that subsequent CI will be NG, including other branches, so I thought it would be better to remove it.
What do you think?

Translated with DeepL.com (free version)

martijnrusschen: Yes the intent of this cache is to reduce the CI time overall since it gets pretty slow if you don't have this. I haven't seen issues with cache hits before on dependency changes. I'd assume it would not match if there's a change in the yarn lock file since we use that in the key.

src/test/datepicker_test.test.tsx
```diff
@@ -787,11 +797,12 @@ describe("DatePicker", () => {
         isClearable
       />,
     );
+    expect(instance).toBeTruthy();
     const clearButton = container.querySelector(
       ".react-datepicker__close-icon",
     );
     fireEvent.click(clearButton ?? new HTMLElement());
-    expect(instance.state.inputValue).toBeNull();
+    expect(instance!.state.inputValue).toBeNull();
```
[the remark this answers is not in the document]
yuki0410-dev: Since the decision was made a little above here, I did not think it was necessary to make the decision again.

.github/workflows/test.yml
yuki0410-dev: @martijnrusschen 
Indeed, cache is created with hash and os in yarn.lock as keys.
However, according to the official documentation, if restore-keys is set, when using cache, it first searches for cache using hash and os in yarn.lock as keys; if cache is a hit, it is used, but if not, only os is used as key. If there is no hit, the repository is configured to use the most recent cache with only os as a key.
https://github.com/actions/cache?tab=readme-ov-file#inputs

I would expect that removing this setting will not slow down all PR CIs, but will make them take longer because cache is no longer allocated to PR CIs that include some yarn.lock changes.

yuki0410-dev: The latest CI shows that cache is being used correctly without restore-keys as long as yarn.lock has not been changed.

https://github.com/Hacker0x01/react-datepicker/actions/runs/9486148892/job/26139812754?pr=4891

```
Run actions/cache@v4
Received 301[9](https://github.com/Hacker0x01/react-datepicker/actions/runs/9486148892/job/26139812754?pr=4891#step:5:10)89888 of 651797708 (46.3%), 287.1 MBs/sec
Cache Size: ~622 MB (651797708 B)
/usr/bin/tar -xf /home/runner/work/_temp/fbc7044b-c1cb-40eb-a8[10](https://github.com/Hacker0x01/react-datepicker/actions/runs/9486148892/job/26139812754?pr=4891#step:5:11)-6c1a5ad6255b/cache.tzst -P -C /home/runner/work/react-datepicker/react-datepicker --use-compress-program unzstd
Received 651797708 of 651797708 (100.0%), 310.2 MBs/sec
Cache restored successfully
Cache restored from key: Linux-yarn-d94afddf929d782fe04721c5606e6ca65a0a1ee89384f90043903a1cc9339958
```

martijnrusschen: ok thanks for clarifying, that sounds like a good approach to me.

Issue #4002, opened 2023-04-18
Date Range Picker not working properly when try to edit directly on input box


lechu24:
**Describe the bug**
When I try to edit the dates in date range picker directly in the input box without using the range picker the start date and end date is not getting updated rather it is getting reseted to old values.

**To Reproduce**

1. Go to date range picker
2. Select a date range (ex: 02/11/2023 - 15/12/2023)
3. Now try to edit the date directly on the input box (example change the end date from 15/12/2023 to 18/12/2024)
4. You will see that the date will not get updated it will be reseted again to 15/12/2023.

**Expected behavior**
It should change to 18/12/2024.
The expected behaviour is that the date should change even when we directly change that in the input box without changing it through date range picker

**Screenshots**
1.![image](https://user-images.githubusercontent.com/45351359/232743910-e6702a98-d0e0-434b-92e8-71a47b5dd0a5.png)

2.
![image](https://user-images.githubusercontent.com/45351359/232744232-ec0d6589-0cea-4565-a5ca-f50095352be0.png)

3.
![image](https://user-images.githubusercontent.com/45351359/232744293-3338d416-ba2a-4eed-839f-6da2b90c1ca2.png)

4.
![image](https://user-images.githubusercontent.com/45351359/232744498-703148fa-0120-4d6f-833d-070c3f64751a.png)

aziyatali:
@lechu24  That's because you are passing selected props startDate
And your startDate is new Date which is current date.

Try something like this and make sure you are selecting both date when props change.

```
<DatePicker
        dateFormat={dateFormat}
        locale={chosenLocale}
        selected={startDate}
        startDate={startDate}
        endDate={endDate}
        onChange={(date) => setStartDate(date)}
      />

    <DatePicker
      dateFormat={dateFormat}
      locale={chosenLocale}
      selected={endDate}
      onChange={(date) => {
        console.log(date);
        setEndDate(date);
      }}
    />
```

rickysandoval:
@aziyatali that's not really an answer. Your snippet if for a separate solution: using two date pickers to make a range.

The issue is with using the selectsRange prop. When using that prop you cannot set or change the entire range using the input. If you type in dates it'll set the start date but it won't recognize any further input past that.

mats363:
[...]

PR #5060, opened 2024-08-29
Parse date range

laug:
## Description
**Linked issue**: #3596 #3906 #4002 

**Problem**
When selectsRange is false, users can input "03/04/2024" in text box and automatically have 4 March 2024 selected in the calendar.
However when selectsRange is true, if users input "03/04/2024 - 05/06/2024" in the text box, nothing happens.

**Changes**
With this PR, when selectsRange is true and the user inputs "03/04/2024 - 05/06/2024", the start date will be set to 4 March 2024 and the end date will be set to 6 May 2024.

## Contribution checklist
- [X] I have followed the [contributing guidelines](https://github.com/Hacker0x01/react-datepicker/blob/main/CONTRIBUTING.md).
- [X] I have added sufficient test coverage for my changes.
- [X] I have formatted my code with Prettier and checked for linting issues with ESLint for code readability.

### DIFF
--- a/src/date_utils.ts
+++ b/src/date_utils.ts
@@ -794,7 +794,7 @@ export function isDayDisabled(
         if (excludeDate instanceof Date) {
           return isSameDay(day, excludeDate);
         } else {
-          return isSameDay(day, excludeDate.date ?? new Date());
+          return isSameDay(day, excludeDate.date);
         }
       })) ||
     (excludeDateIntervals &&
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -565,6 +565,7 @@ export default class DatePicker extends Component<
     }
   };
 
+  // handleChange is called when user types in the textbox
   handleChange = (
     ...allArgs: Parameters<Required<DatePickerProps>["onChangeRaw"]>
   ) => {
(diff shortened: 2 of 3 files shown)


src/date_utils.ts
```diff
@@ -794,7 +794,7 @@ export function isDayDisabled(
         if (excludeDate instanceof Date) {
           return isSameDay(day, excludeDate);
         } else {
-          return isSameDay(day, excludeDate.date ?? new Date());
+          return isSameDay(day, excludeDate.date);
```
laug, line 797: When determining whether to disable a day due to it being in the list of excluded dates, it is not correct to check whether `day` is the same day as `new Date()` (today) as today is not necessarily in the list of excluded dates.
If `day` is non null and `excludeDate.date` is null, isSameDay will return false, which is correct.

src/index.tsx
```diff
@@ -585,36 +586,77 @@ export default class DatePicker extends Component<
...
+      );
+      const startChanged = startDate?.getTime() !== startDateNew?.getTime();
+      const endChanged = endDate?.getTime() !== endDateNew?.getTime();
+
+      if (!startChanged && !endChanged) {
+        return;
+      }
+
+      if (startDateNew && isDayDisabled(startDateNew, this.props)) {
+        return;
```
martijnrusschen, line 625: Can you check if there's test coverage for this case?

src/index.tsx
```diff
@@ -585,36 +586,77 @@ export default class DatePicker extends Component<
...
+        valueEnd ?? "",
+        dateFormat,
+        this.props.locale,
+        strictParsing,
+      );
+      const startChanged = startDate?.getTime() !== startDateNew?.getTime();
+      const endChanged = endDate?.getTime() !== endDateNew?.getTime();
+
+      if (!startChanged && !endChanged) {
+        return;
```
martijnrusschen, line 621: Can you check if there's test coverage for this case?

src/index.tsx
```diff
@@ -585,36 +586,77 @@ export default class DatePicker extends Component<
...
+      }
+
+      if (startDateNew && isDayDisabled(startDateNew, this.props)) {
+        return;
+      }
+      if (endDateNew && isDayDisabled(endDateNew, this.props)) {
+        return;
```
martijnrusschen, line 628: Can you check if there's test coverage for this case?

laug: I've added coverage for the 3 early returns in commit 55f49a0, so it should be good now hopefully.

PR #5069, opened 2024-09-02
test: fix a test that would always fail if run on first day of month

laug:
## Description
**Problem**
The test suite failed on September 1st, and every 1st of the month in the past, because this test assumes `today` and `subDays(today, 1)` are in the same month.

**Changes**
We check if today is the first of the month, and if so, we use `addDays(today, 1)` instead of `subDays(today, 1)`.

## Contribution checklist
- [X] I have followed the [contributing guidelines](https://github.com/Hacker0x01/react-datepicker/blob/main/CONTRIBUTING.md).
- [X] I have added sufficient test coverage for my changes.
- [X] I have formatted my code with Prettier and checked for linting issues with ESLint for code readability.

### DIFF
--- a/src/test/exclude_dates.test.tsx
+++ b/src/test/exclude_dates.test.tsx
@@ -1,14 +1,18 @@
 import { render } from "@testing-library/react";
-import { subDays } from "date-fns";
+import { addDays, subDays } from "date-fns";
 import React from "react";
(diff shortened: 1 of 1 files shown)


martijnrusschen:
Nice :)

Issue #5193, opened 2024-10-25
The "previous month" button appears when selecting a date


ahrybouskaja-easyjet:
**Describe the bug**
The "previous month" button appears when selecting a date

**To Reproduce**

1. Go to https://codesandbox.io/p/sandbox/reactdatepicker-forked-kxtkd8
2. Prev button absent
![image](https://github.com/user-attachments/assets/638f3309-e8fe-49dc-b255-b8cd02e99967)
4. Select date in second month
5. Prev button appears
![image](https://github.com/user-attachments/assets/be1eb83d-fbf9-4cdb-9eff-e4484cc3daf3)

**Expected behavior**
the "previous month" button does not appear because the calendar starts in October

**Additional context**
v.6.9.0 also reproduced in 7.4.0

pmacmillan:
This seems to fix the issue. Basically the current code relies on the date selected. However if multiple calendars are selected then it doesn't know if the min month is rendered.

Borrowed some code from `renderMonths` to do the necessary math (`monthSelectedIn` is the relative calendar offset that the date is selected in -- so in your first example it's 0, and in the second example it's 1).

![image](https://github.com/user-attachments/assets/7178765b-23f5-49ed-941a-6311f24b7dfe)

So the fix is likely to consider which was the 0th calendar rendered by applying the same math. (i.e. if the first month calendar is October and the selected date is in November, then we need to subtract 1 to ensure we're checking the correct date.

not creating a PR because tests (even on a clean checkout) aren't passing for me.

```diff
diff --git a/src/calendar.tsx b/src/calendar.tsx
index 461e1c81..799de807 100644
--- a/src/calendar.tsx
+++ b/src/calendar.tsx
@@ -543,7 +543,17 @@ export default class Calendar extends Component<CalendarProps, CalendarState> {
         );
         break;
       default:
-        allPrevDaysDisabled = monthDisabledBefore(this.state.date, this.props);
+        const monthsShown =
+          this.props.monthsShown ?? Calendar.defaultProps.monthsShown;
+        const monthsToSubtract = this.props.showPreviousMonths
+          ? monthsShown - 1
+          : 0;
+
+        const monthSelectedIn = this.props.monthSelectedIn ?? monthsToSubtract;
+        const fromMonthDate = subMonths(this.state.date, monthSelectedIn);
+
+        allPrevDaysDisabled = monthDisabledBefore(fromMonthDate, this.props);
+
         break;
     }
```

[...]

PR #5214, opened 2024-11-07
fix: "previous month" button appears when selecting a date

OlegDev1:
## Description
**Linked issue**: #5193

**Problem**
Even though the min date is provided, when selecting the second month, "previous month" button still appears. It shouldn't happen, because the calendar from the screenshots starts at November. 

**Changes**
Changed the calculation logic that determines whether previous month button should appear or not (borrowed some code from `renderMonths`, big thanks to @pmacmillan). Added a test case, in which a calendar is rendered with min date, after that the date from the second month is selected and it is expected for "previous month" button not to appear.

## Screenshots
*Calendar starts at November 7*

**Before**
image
**After**
image

## Contribution checklist

### DIFF
--- a/src/calendar.tsx
+++ b/src/calendar.tsx
@@ -528,6 +528,14 @@ export default class Calendar extends Component<CalendarProps, CalendarState> {
       return;
     }
 
+    const monthsShown =
+      this.props.monthsShown ?? Calendar.defaultProps.monthsShown;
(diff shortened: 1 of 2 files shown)


OlegDev1:
Thank you very much for the review, I really appreciate it! 

I have added an additional button check in the test case as Frank suggested. I also replaced the non-null assertion with an error check

PR #5292, opened 2024-12-27
:ambulance: Fix the problem where pressing Enter after deleting a value in the custom input results in the first value being incorrectly formatted as '2001'.

ftrdzputrap:
I have an issue where, after typing the year and pressing Enter in the customInput field of React Datepicker, the value is automatically formatted when I delete it.

I'm handling the issue by blurring the field when Enter or Escape is pressed, and ensuring the selected value is preserved.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -962,6 +962,7 @@ export default class DatePicker extends Component<
       const copy = newDate(this.state.preSelection);
       if (eventKey === KeyType.Enter) {
         event.preventDefault();
+        (event.target as HTMLInputElement).blur();
         if (
           this.inputOk() &&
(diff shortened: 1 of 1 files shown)


ftrdzputrap:
> It may be worth switching from casting to a type guard (`if (event.target instanceof HTMLInputElement) { ... }`) and handling the case where that isn't true, in order to help avoid future bugs. Other than that, this seems pretty straightforward and looks good to me.
> 
> Image of Ryan Lester <@buu700> **Ryan Lester <@buu700>**
> 

so i must handle this event using instanceof HTMLInputElement to get approval? or is there another way?

PR #5128, opened 2024-09-25
fix: remove date-fns as external package

johannrp27:
---
name: fix: remove date-fns as external package
about: Update of date-fns library from 3.6 to 4.1 was running into a issue with current rollup configuration
title: Remove date-fns as external package
labels: ""
assignees: "[martijnrusschen](https://github.com/martijnrusschen)"
---

## Description
**Linked issue**: #5108 

**Problem**
Get back online page with latest version of date-fns

**Changes**
Rollup issue remove

## Contribution checklist
- [x] I have followed the [contributing guidelines](https://github.com/Hacker0x01/react-datepicker/blob/main/CONTRIBUTING.md).
- [x] I have added sufficient test coverage for my changes.
- [x] I have formatted my code with Prettier and checked for linting issues with ESLint for code readability.

PR #5326, opened 2025-01-20
chore(deps): upgrade date-fns to v4.1.0

paolostyle:
## Description
**Linked issue**: #5177, #5128

**Problem**
This PR update `date-fns` to the latest version, v4.1.0. Dependabot update breaks the docs website (which is a separate topic in itself, the build system over there is extremely outdated - happy to help with that), my changes fix these issues so the website is still functional.

**Changes**
- bumped version in package.json
- removed the logic in Rollup that marks date-fns exports as external
- adjusted imports to use `date-fns` directly, also in tests for consistency

## To reviewers
This might seem like a bad decision, after all previous contributors were using direct date-fns exports to reduce the bundle size! This isn't actually correct after v4.1.0. Since v4, `date-fns` is marked as `type: "module"`, so really any somewhat up-to-date bundle will handle tree shaking correctly (in fact this might've been the case even earlier as date-fns was providing ESM exports back in v3, too). I have a hard time imagining that people stuck on super old bundlers that don't work with ESM are keeping this library up to date.

On another note, if this was an issue, someone would likely report it already, because there was a direct `date-fns` import in `src/calendar.tsx` for a while now, so for CJS builds this would've had an impact already anyway.

As a small positive side effect, the final bundle size is slightly smaller for every produced target.

After this is merged I believe both linked issues can be closed; one is the dependabot's PR, the other contains a subset of my changes.

## Contribution checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
   "dependencies": {
     "@floating-ui/react": "^0.27.0",
     "clsx": "^2.1.1",
-    "date-fns": "^3.6.0"
+    "date-fns": "^4.1.0"
   },
   "scripts": {
     "eslint": "eslint --ext .js,.jsx,.ts,.tsx ./src",
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -24,25 +24,6 @@ const banner = `/*!
   Released under the ${pkg.license} License.
 */`;
 
-// it's important to mark all subpackages of data-fns as externals
-// see https://github.com/Hacker0x01/react-datepicker/issues/1606
-// We're relying on date-fn's package.json `exports` field to
-// determine the list of directories to include.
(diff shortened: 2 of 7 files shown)


martijnrusschen:
@paolostyle Let's give this a try. Thanks for pushing this along. The added context in the description will be helpful for future references. If you're up with helping to upgrade outdated logic feel (such as the build system) free to open a PR.

martijnrusschen:
OK, it seems we're not fully functional: https://reactdatepicker.com/

paolostyle:
Okay I must've messed up something during my local development as it indeed does not work after I did a clean set up, more specifically the UMD build used by the docs page does not work. I will revisit this and get back to you.

PR #5332, opened 2025-01-21
chore(deps): upgrade date-fns to v4.1.0 (attempt no. 2)

paolostyle:
## Description
**Linked issue**: #5177, #5128, #5326

**Problem**

This PR is a second attempt after #5326 to update `date-fns` to the latest version, v4.1.0. Dependabot update breaks the docs website (which is a separate topic in itself, the build system over there is extremely outdated - happy to help with that), my changes fix these issues so the website is still functional (this time for real, I triple checked).

**Changes**
- bumped `date-fns` version in package.json
- adjusted exports in package.json, namely removed `browser` field entirely and added more modern `exports` field
- removed the logic in Rollup that marks date-fns exports as external
- fixed UMD build so it actually works (it was probably broken for a very long time, though)
- adjusted imports to use `date-fns` directly, also in tests for consistency

## To reviewers
So it turns out that the main cause of the docs page failing after all was the `browser` field in package.json. It was pointing to the UMD build, which just wouldn't work with `date-fns` after the update. It points `browser` to an ESM file exporting all the other ESM files with individual functions. ESM and UMD are not compatible with each other (oh well), so there was just no way it would work.

What I did is I simply removed the `browser` field. I honestly think it was just used incorrectly until now. We could keep it (pointing to the ESM build) but if `browser` field is missing bundlers use `module` field anyway, and some (namely `esbuild`) have additional behavior where they can use `main` instead if imported with `require`. From what I've checked most frontend libraries nowadays tend to omit this field, but if you REALLY want to keep it, I can - just need to change it to the ESM build.

I kept the changes from previous PR, namely using only main `date-fns` export as an import source. This is part of the "fixing UMD bundle" effort which I decided to do as soon as I found out... it doesn't work at all. Or I guess it could work but it would require an enormous amount of effort. After extending the `globals` field in Rollup config it's now doable to use it with only CDN-imported dependencies.

I also added the `exports` field which I initially thought was relevant to this. It is not, however I think it's still valuable to keep it. If you'd prefer to do it in a separate PR (I will open one tomorrow re/ docs site), I can do that, too.

I think that's all.

## Contribution checklist

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -26,7 +26,10 @@ const banner = `/*!
 
 const globals = {
   react: "React",
-  "prop-types": "PropTypes",
+  "react-dom": "ReactDOM",
+  clsx: "clsx",
+  "date-fns": "dateFns",
+  "@floating-ui/react": "FloatingUIReact",
 };
 
 // NOTE:https://rollupjs.org/migration/#changed-defaults
@@ -40,7 +43,7 @@ const config = {
   input: "src/index.tsx",
   output: [
     {
-      file: pkg.browser,
+      file: pkg.unpkg,
       format: "umd",
       name: "DatePicker",
       globals,
@@ -49,7 +52,7 @@ const config = {
       plugins: [terser()],
     },
     {
-      file: pkg.browser.replace(".min.js", ".js"),
+      file: pkg.unpkg.replace(".min.js", ".js"),
       format: "umd",
       sourcemap: "inline",
       name: "DatePicker",
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -24,25 +24,6 @@ const banner = `/*!
   Released under the ${pkg.license} License.
 */`;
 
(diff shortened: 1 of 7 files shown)


rollup.config.mjs
```diff
@@ -24,28 +24,12 @@ const banner = `/*!
...
-  .filter((key) => key !== "." && key !== "package.json")
-  .map((key) => `date-fns/${key}`);
-
 const globals = {
   react: "React",
-  "prop-types": "PropTypes",
+  "react-dom": "ReactDOM",
+  clsx: "clsx",
+  "date-fns": "dateFns",
+  "@floating-ui/react": "FloatingUIReact",
```
martijnrusschen, line 32: These seem new, are these required?

paolostyle: They make the UMD build work in non-bundler environments. So not strictly necessary in the context of this PR but overall I think they should be kept

Issue #3797, opened 2022-10-19
Accessibility issue - screen reader readout of "Su Mo Tu We Th Fr Sa"


Nina0295:
**Describe the bug**

When a screenreader user arrows through the content in the datepicker, the screen reader reads the calendar header's days of the week as "Su" "Mo" "Tu" We" (I tested with voiceover)

**To Reproduce**

Go to the react-datepicker demo (https://reactdatepicker.com/)
Turn on a screenreader (I used Voiceover)
Arrow through the calendar to the days of the week at the top and observe how they are announced by the screen reader
Expected behavior

It would be better if either:

The weekdays had aria-labels and were read by the screen reader as "Sunday" "Monday" etc.
These elements were skipped over altogether (aria-hidden=true) since the actual dates are read as "Sunday August 1 2021" and already include the day, so being able to focus on the weekdays isn't really useful.

**Desktop (please complete the following information):**

MAC OS
Browser: Chrome

jphawk:
It would be good to have a possibility to customise the day names' aria-labels for Date and MonthPickers. Right now. unfortunately, it's really confusing for the end user.

tsederberg:
I strongly agree with this accessibility improvement suggestion. 

The current behavior of reading abbreviated day names ("Su", "Mo", etc.) provides little value and can be confusing

SaiBalaji202:
I'll work on it

PR #5988, opened 2025-11-02
fix calendar weekday name with custom styles

atarek12:
## Description

**Problem**
There is an issue with weekday name when using `datepicker` component without using datepicker styles 
i.e: `import "react-datepicker/dist/react-datepicker.css";`

**Changes**
The issue has been introduced in this [pr](https://github.com/qburst/react-datepicker-3/pull/56) that fixes this [issue](https://github.com/Hacker0x01/react-datepicker/issues/3797). 
But instead of using `sr-only` class on an extra `span` element which will be rendered if we don't import the stylesheet. 
I have reverted it back to use `aria-label` but kept the inner `span` with `aria-hidden="true"`; so the screen-readers will skip it and announce the `aria-label` prop.

## Screenshots
- Before: Notice the duplicate of weekday -> `SundaySu`
Screenshot 2025-11-02 at 9 48 09 PM

- After
Screenshot 2025-11-02 at 9 48 56 PM

## Contribution checklist

### DIFF
--- a/src/calendar.tsx
+++ b/src/calendar.tsx
@@ -481,6 +481,18 @@ export default class Calendar extends Component<CalendarProps, CalendarState> {
       this.props.calendarStartDay,
     );
 
+    const srOnlyStyles: React.CSSProperties = {
+      position: "absolute",
+      width: "1px",
+      height: "1px",
+      padding: 0,
+      margin: "-1px",
+      overflow: "hidden",
+      clipPath: "inset(50%)",
+      whiteSpace: "nowrap",
+      border: 0,
+    };
+
     const dayNames: React.ReactElement[] = [];
     if (this.props.showWeekNumbers) {
       dayNames.push(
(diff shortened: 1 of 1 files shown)


martijnrusschen:
Please check the output of the failing CI

martijnrusschen:
Looks like there's a failing test here. Please resolve that.

martijnrusschen:
Thank you for your PR! Unfortunately, we can't merge this change.

The aria-label approach breaks axe accessibility tests with "empty-table-header" violations. When columnheader elements only contain <span aria-hidden="true">, screen readers have no text to announce.

The existing sr-only span pattern is the correct accessibility approach - it provides visible text for screen readers while hiding it visually. This pattern requires the CSS stylesheet to be imported, which is already expected for the component to function properly (layout, colors, positioning, etc.).

If you're using the datepicker without importing the stylesheet, we'd recommend importing it or providing your own styles that include the .react-datepicker__sr-only class definition.

Issue #5939, opened 2025-10-16
Calendar doesn't display the `startDate` as first month


tcpstrh:
**Describe the bug**
The calendar doesn't display the `startDate` as first month if `monthsShown={2}` and `selectsRange` is set.

**To Reproduce**

Example:
```
const SelectRangeIssueWith2Months = () => {
  const [startDate, setStartDate] = useState<Date | null>(new Date());
  const [endDate, setEndDate] = useState<Date | null>(null);

  const onChange = (dates: [Date | null, Date | null]) => {
    const [start, end] = dates;
    setStartDate(start);
    setEndDate(end);
  };

  return (
    <DatePicker
      selected={startDate}
      onChange={onChange}
      startDate={startDate}
      endDate={endDate}
      selectsRange
      monthsShown={2}
    />
  );
};

render(SelectRangeIssueWith2Months);
```

1. Go to input field and select end date
2. Click outside
3. Click into input field

**Expected behavior**
The calendar display the `startDate` as first month.

**Screenshots**
The calendar display the `startDate` in second month as if `showPreviousMonths` would be set.

<img width="915" height="564" alt="Image" src="https://github.com/user-attachments/assets/9a325ae7-9274-4c1b-abfa-a8805f302f03" />

**Desktop:**

- OS: MacOS
- Browser: chrome
- Version 1.4.1
- react-datepicker version: 8.7.0

**Additional context**
As a workaround the issue can be fixed by setting the `focusSelectedMonth` prop. But this introduces another behaviour of the calendar component - selected month is focused (directly set as first month) during user input.

tcpstrh:
Seems to be related to #5916

martijnrusschen:
This should be fixed in v8.10.0. The fix for #5939 (PR #5948) also addresses this issue by resetting monthSelectedIn to 0 when selectsRange is enabled, ensuring the calendar always displays starting from the current month after clearing a selection.

Please upgrade to v8.10.0 and let us know if you're still experiencing this issue.

PR #6103, opened 2025-12-04
fix: prevent month view jump when selecting from second calendar in range mode

### LINKED ISSUE #5275: When displaying 2 months and having an embedded calendar, the months shift when the user selects the second calendar
**Describe the bug**
When selecting a start date from the second calendar in a multi-month react-datepicker component, the selected month of the second calendar automatically shifts to the next month, and the focus switches back to the first calendar.

**To Reproduce**

Use the react-datepicker component with monthsShown={2} and selectsRange={true}.
Open the date picker and select a start date from the second calendar.
Observe that the second calendar shifts its displayed month, and the focus returns to the first calendar.

**Expected behavior**
The second calendar should remain in place, displaying the same month, and the focus should not switch to the first calendar. Both calendars should independently maintain their current view.

Screenshots

https://github.com/user-attachments/assets/27321661-e2ab-436d-97c1-a9da29ade159

**Desktop**:
OS: Macbook 11
Browser: Chrome

**Additional context**
This issue seems related to the focusSelectedMonth behavior, where selecting a range date causes the calendar to refocus and update its displayed month.
Using a workaround with focusSelectedMonth={false} and dynamically setting it to true using a timeout not resolve this problem

martijnrusschen:
When using monthsShown={2} with selectsRange={true}, selecting a date from the second calendar no longer causes the view to jump back to the first month.

The previous fix for #5939 unconditionally reset monthSelectedIn to 0 on every update when selectsRange was true. This fix changes it to only reset when the calendar opens (transitions from closed to open), which preserves the #5939 fix while allowing normal interaction with the second calendar.

Fixes #5275

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -357,7 +357,16 @@ export class DatePicker extends Component<DatePickerProps, DatePickerState> {
     ) {
       this.setState({ monthSelectedIn: 0 });
     }
-    if (this.props.selectsRange && this.state.monthSelectedIn !== 0) {
+    // Reset monthSelectedIn when calendar opens for range selection
+    // This ensures startDate is displayed as the first month when reopening
+    // (Fix for #5939), but we don't reset during active selection to avoid
+    // the view jumping when clicking dates in the second calendar (Fix for #5275)
+    if (
+      this.props.selectsRange &&
+      prevState.open === false &&
+      this.state.open === true &&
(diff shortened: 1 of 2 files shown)


balajis-qb:
Hi  @martijnrusschen, I Verified this fix.  It's working good

PR #6182, opened 2025-12-19
fix: resolve webpack warning for optional date-fns-tz dependency

### LINKED ISSUE #6181: v.9.1.0 webpack warning (p2)
**Describe the bug**
When there is no `date-fns-tz` webpack shows a warning, but `date-fns-tz` is optional.
A followup on  #6154 warning. The previous warning is gone, but different warning appeared:

**To Reproduce**
Install `react-datepicker` in a project with `webpack`, but without `date-fns-tz`
```
[client] Compiled with warnings.
[client]
[client] Critical dependency: the request of a dependency is an expression
[client]
[client] Search for the keywords to learn more about each warning.
[client] To ignore, add // eslint-disable-next-line to the line before.
[client]
[client] WARNING in ./node_modules/react-datepicker/dist/index.es.js 150:16-44
[client] Critical dependency: the request of a dependency is an expression
[client]
[client] webpack compiled with 1 warning
```

**Expected behavior**
There should be no warning message

### LINKED ISSUE #6204: Timezone support broken in Vite due to dynamic require of date-fns-tz
**Describe the bug**
When developing with Vite with the default React config, the new `timeZone` prop fails to work.

This seems to be due to the code dynamically using `require` to import `date-fns-tz`, which apparently isn't supported by Vite:
https://github.com/Hacker0x01/react-datepicker/blob/644f2f512d2efcc83885811b0faa3de7fc111350/src/date_utils.ts#L118-L120

Screenshot from debugger, inside the `require` call in the above code.
Image

**To Reproduce**

1. See the minimal Vite setup in https://codesandbox.io/p/devbox/fmvlts (use https://fmvlts-5173.csb.app/ to view outside the editor)
2. Observe times are rendered in local time
3. Observe the `timeZone prop requires "date-fns-tz" package` warning in console

**Expected behavior**
Dates rendered in configured `timeZone` (`Pacific/Auckland` in example to make it clear).

If this requires specific configuration on the Vite side, it would be a good idea to document in `timezone.md` (but I couldn't find anything to enable dynamic `require` in Vite).

**Screenshots**
N/A

- OS: Debian testing
- Browser: Firefox 147.0b9

martijnrusschen:
## Summary

- Uses `__non_webpack_require__` to load the optional `date-fns-tz` module
- This tells webpack to use the native Node.js require at runtime instead of trying to bundle/analyze the module
- Eliminates the "Critical dependency: the request of a dependency is an expression" warning
- **Added:** Handles ESM-only environments (like Vite) where `require` is not available

## Background

When `date-fns-tz` is not installed (it's an optional peer dependency), webpack emits a warning because it can't statically analyze the dynamic `require(variable)` pattern. By using webpack's special `__non_webpack_require__` global, we bypass webpack's bundling analysis while still falling back to regular `require` in non-webpack environments.

### Vite Support

Vite and other ESM-only bundlers don't provide a `require` function. This PR adds a guard to check if `require` is available before attempting to use it, preventing runtime errors. However, due to the limitations of ESM, Vite users who want timezone support will need to configure their bundler - see #6204 for details.

## Test plan

- [x] TypeScript type checking passes
- [x] All tests pass
- [x] Build succeeds and includes the fix in dist output

Fixes #6181
Fixes #6204

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/date_utils.ts
+++ b/src/date_utils.ts
@@ -99,6 +99,10 @@ export function __setDateFnsTzNull(): void {
   dateFnsTzLoadAttempted = true;
 }
 
+// Declare webpack's special require function that bypasses bundling
+// This is used to avoid webpack warnings for optional dependencies
+declare const __non_webpack_require__: typeof require | undefined;
+
 /**
  * Attempts to load date-fns-tz module.
  * Returns null if the module is not installed.
@@ -111,13 +115,15 @@ function getDateFnsTz(): DateFnsTz | null {
   dateFnsTzLoadAttempted = true;
 
   try {
-    // Dynamic require for date-fns-tz
-    // Use a variable to prevent webpack from statically analyzing the require
-    // and showing warnings when the optional dependency is not installed
-    // See: https://github.com/Hacker0x01/react-datepicker/issues/6154
-    const dateFnsTzModuleName = "date-fns-tz";
+    // Use __non_webpack_require__ to tell webpack to use native require
+    // and avoid bundling warnings for this optional dependency
+    // See: https://github.com/Hacker0x01/react-datepicker/issues/6181
     // eslint-disable-next-line @typescript-eslint/no-require-imports
-    dateFnsTz = require(dateFnsTzModuleName) as DateFnsTz;
+    const requireFn =
+      typeof __non_webpack_require__ !== "undefined"
+        ? __non_webpack_require__
+        : require;
+    dateFnsTz = requireFn("date-fns-tz") as DateFnsTz;
   } catch {
     /* istanbul ignore next - only executes when date-fns-tz is not installed */
     dateFnsTz = null;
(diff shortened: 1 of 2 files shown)


PurkkaKoodari:
When `require` is not set, the code in f492f03 never reaches a path that would set `dateFnsTz` to something other than `null`. I don't think that's resolved by any dependency pre-bundling change in Vite.

[Here's](https://stackoverflow.com/questions/79459434/optional-dynamic-import-that-both-webpack-and-vite-are-happy-with) a relatively recent StackOverflow thread with related example code. It only talks about `import`, and I think you may need to use `import` in ESM... which also brings its own can of worms with it being async. Other options might be a "bring your own `date-fns-tz`" prop or a separate component that unconditionally pulls in `date-fns-tz`.

PurkkaKoodari:
Hmm, in fact, presumably after #6155, I can't get `timeZone` to work even with Webpack and the hello-world application in the repo.

FWIW, I've made another PR #6227 which provides another way forward, but I think that might need at least a docs update if this is broken in Webpack as well.

PR #6206, opened 2026-01-05
fix: correct timezone handling for date selection and display

### LINKED ISSUE #6193: Strange behavior when picking the date in different timezones in v. 9.1.0
**Describe the bug**
When the timeZone prop value (e.g., 'Pacific/Kiritimati') differs from the current browser timezone (e.g, 'UTC'), the date displayed in the date-picker input (and the calendar itself) is not the same user actually picks in the calendar.  

**Use case**
In our web application, users can select a time zone associated with their profile. This selected time zone is used consistently to display — and to create or edit — dates throughout the user interface, regardless of the user’s actual physical location. For example, if a user travels to the Pacific region and accesses the application from there, dates in the application will still be displayed according to the time zone defined in their profile (e.g., UTC).

**To Reproduce**

1. Go to example: https://codesandbox.io/p/sandbox/6hvd2k
2. Open Chrome DevTools and adjust the current browser location (Sensors panel) to be 'UTC'
3. Pick the date Dec 26th 2025 in the calendar
4. The calendar has actually picked Dec 25th 2025 and displays it in the input field + calendar

**Expected behavior**
When I pick a date in the calendar, the selection of this date should appear in the input field, regardless of the timeZone I have set and the current browser timezone.

**Screenshots**
I use calendar UI to pick Dec 26th 2025
Image

What I actually get is Dec 25th 2025
Image

- OS: Mac OS
- Browser: Chrome
- Version 143.0.7499.147 

**Additional context**

https://github.com/user-attachments/assets/b344b06e-4728-4f7f-8a3b-411c88430354

Thank you!

martijnrusschen:
## Summary

- Fixed input display not using timezone-aware formatting when `timeZone` prop is provided
- Fixed calendar not receiving zoned dates for `selected`, `startDate`, `endDate`, and `selectedDates` props, which caused day selection highlighting to fail when timezone differs significantly from browser timezone
- Added `timeZone` parameter support to `safeDateFormat`, `safeDateRangeFormat`, and `safeMultipleDatesFormat` functions

## Problem

When the `timeZone` prop was set to a timezone significantly different from the browser's timezone (e.g., `Pacific/Kiritimati` UTC+14 with browser in UTC), clicking a date in the calendar would result in the wrong date being selected and displayed.

For example:
1. Set `timeZone="Pacific/Kiritimati"` and browser to UTC
2. Pick Dec 26th in the calendar
3. The calendar showed Dec 25th instead

## Solution

1. Updated `safeDateFormat()` in `date_utils.ts` to use `formatInTimeZone()` when a `timeZone` parameter is provided
2. Updated `getInputValue()` in `index.tsx` to pass the `timeZone` prop to date formatting functions
3. Updated `renderCalendar()` in `index.tsx` to convert `selected`, `startDate`, `endDate`, and `selectedDates` to zoned time before passing to Calendar

## Test plan

- [x] Added new tests that reproduce the issue with extreme timezone differences
- [x] All 1484 existing tests pass
- [x] Type check passes
- [x] Lint passes

Fixes #6193

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/date_utils.ts
+++ b/src/date_utils.ts
@@ -439,19 +439,38 @@ export function formatDate(
  * Safely formats a date.
  *
  * @param date - The date.
- * @param options - An object containing the dateFormat and locale.
+ * @param options - An object containing the dateFormat, locale, and optional timeZone.
  * @returns - The formatted date or an empty string.
  */
 export function safeDateFormat(
   date: Date | null | undefined,
-  { dateFormat, locale }: { dateFormat: string | string[]; locale?: Locale },
+  {
+    dateFormat,
+    locale,
+    timeZone,
+  }: { dateFormat: string | string[]; locale?: Locale; timeZone?: TimeZone },
 ): string {
   const formatStr = (
     Array.isArray(dateFormat) && dateFormat.length > 0
       ? dateFormat[0]
       : dateFormat
   ) as string; // Cast to string because it's impossible to get `string | string[] | undefined` here and typescript doesn't know that
-  return (date && formatDate(date, formatStr, locale)) || "";
+
+  if (!date) {
+    return "";
+  }
+
+  // Use timezone-aware formatting if timeZone is specified
+  if (timeZone) {
+    // Resolve locale string to locale object for formatInTimeZone
+    // Cast to DateFnsLocale since LocaleObj is a compatible subset
+    const localeObj = (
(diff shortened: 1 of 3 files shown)


igor-menovsky-ipf:
Hi @martijnrusschen. Thanks a lot for the fix! It is much appreciated and awaited in our project, as it blocks us a bit. Do you have any expectation, when this fix could be released? Thanks a lot! Igor

PR #6232, opened 2026-01-28
Fix #6223 - ♿ Accessibility improvements: Month and Year dropdowns

### LINKED ISSUE #6223: Month and year dropdowns do not have an accessible name
**Describe the bug**
Currently, the year and month dropdowns (there may be others but these are the ones i've been flagged with) within the popover element render `select` elements, but they do not have an accessible name and there are no props exposed to add the necessary attributes and/or elements to satisfy WCAG 2.1 (A). See more info here on how a select is expected to have an associated label: https://dequeuniversity.com/rules/axe/4.9/select-name?application=axeAPI

**To Reproduce**

1. Render the datepicker with the props `showMonthDropdown` and `showYearDropdown` set to true
2. Open the popover
3. Inspect the dropdown menu

**Expected behavior**
There are several ways to associate a label with a `select`. One such example is `<select aria-label="select a year">...</select>`

**Screenshots**

Image

- OS: MacOS Sequoia 15.5
- Browser: all (screenshot is from Chrome)
- Version 143.0.7499.169 (Official Build) (arm64)

**Additional context**

balajis-qb:
## Description
**Linked issue**: #6223

**Problem**
As mentioned in the attached ticket, the month select dropdown (`showMonthDropdown`) and the year select drop down (`showYearDropdown`) lacks ARIA attributes. 

**Changes**
- Added the corresponding aria attributes to the select and the button element to support the dropdown mode (`dropdownMode`) of scroll and select
- Added test cases to validate the change

## Contribution checklist

### DIFF
--- a/src/month_dropdown.tsx
+++ b/src/month_dropdown.tsx
@@ -36,7 +36,12 @@ export default class MonthDropdown extends Component<
   renderSelectOptions = (monthNames: string[]): React.ReactElement[] =>
     monthNames.map<React.ReactElement>(
       (m: string, i: number): React.ReactElement => (
-        <option key={m} value={i}>
+        <option
+          key={m}
+          value={i}
+          aria-label={`Select Month ${m}`}
+          aria-selected={i === this.props.month ? "true" : "false"}
+        >
           {m}
         </option>
       ),
@@ -47,6 +52,7 @@ export default class MonthDropdown extends Component<
       value={this.props.month}
       className="react-datepicker__month-select"
       onChange={(e) => this.onChange(parseInt(e.target.value))}
+      aria-label="Select Month"
     >
       {this.renderSelectOptions(monthNames)}
     </select>
@@ -62,8 +68,14 @@ export default class MonthDropdown extends Component<
       style={{ visibility: visible ? "visible" : "hidden" }}
       className="react-datepicker__month-read-view"
       onClick={this.toggleDropdown}
+      aria-label="Select Month"
+      aria-expanded={this.state.dropdownVisible}
+      aria-haspopup="listbox"
     >
-      <span className="react-datepicker__month-read-view--down-arrow" />
+      <span
+        className="react-datepicker__month-read-view--down-arrow"
+        aria-hidden="true"
+      />
       <span className="react-datepicker__month-read-view--selected-month">
         {monthNames[this.props.month]}
       </span>
--- a/src/month_dropdown_options.tsx
+++ b/src/month_dropdown_options.tsx
@@ -50,6 +50,7 @@ export default class MonthDropdownOptions extends Component<MonthDropdownOptions
             }
           }}
           role="button"
+          aria-label={`Select Month ${month}`}
           tabIndex={0}
           className={
             this.isSelectedMonth(i)
--- a/src/test/month_dropdown_test.test.tsx
+++ b/src/test/month_dropdown_test.test.tsx
@@ -44,6 +44,40 @@ describe("MonthDropdown", () => {
       monthDropdown = getMonthDropdown();
     });
 
+    it("sets proper ARIA on read view button and toggles aria-expanded", () => {
+      const monthReadView = safeQuerySelector(
+        monthDropdown,
+        ".react-datepicker__month-read-view",
+      );
+      expect(monthReadView.getAttribute("aria-label")).toBe("Select Month");
+      expect(monthReadView.getAttribute("aria-haspopup")).toBe("listbox");
+      expect(monthReadView.getAttribute("aria-expanded")).toBe("false");
+
+      fireEvent.click(monthReadView);
+
+      const monthReadViewAfterOpen = safeQuerySelector(
+        monthDropdown,
+        ".react-datepicker__month-read-view",
+      );
+      expect(monthReadViewAfterOpen.getAttribute("aria-expanded")).toBe("true");
+    });
+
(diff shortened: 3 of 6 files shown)


src/month_dropdown.tsx
```diff
@@ -36,7 +36,12 @@ export default class MonthDropdown extends Component<
   renderSelectOptions = (monthNames: string[]): React.ReactElement[] =>
     monthNames.map<React.ReactElement>(
       (m: string, i: number): React.ReactElement => (
-        <option key={m} value={i}>
+        <option
+          key={m}
+          value={i}
+          aria-label={`Select Month ${m}`}
+          aria-selected={i === this.props.month ? "true" : "false"}
```
halvorsanden, line 43: aria-label is not necessary here, the value is already inside this element and the context this appears in already communicates that this is an option that can be selected.
The same with aria-selected I believe, the option element is inside a select element, both have built-in roles and meaning stating what is selected.
Adding aria to elements like these that already have the semantics and meaning built-in can make things less accessible because it is announced differently than the user might expect when encountering the native HTML elements.

src/month_dropdown.tsx
```diff
@@ -47,6 +52,7 @@ export default class MonthDropdown extends Component<
       value={this.props.month}
       className="react-datepicker__month-select"
       onChange={(e) => this.onChange(parseInt(e.target.value))}
+      aria-label="Select Month"
```
halvorsanden, line 55: This is what the referred issue points to, and also what I’m after, but this should be a prop so it can use i18n. If a default is needed, I would recommend only "Month" since "select" is already given by the element type. The entire date dialog is also announced as "choose date", so there is no need to repeat "select/choose" too many places.

src/month_dropdown.tsx
```diff
@@ -62,8 +68,14 @@ export default class MonthDropdown extends Component<
       style={{ visibility: visible ? "visible" : "hidden" }}
       className="react-datepicker__month-read-view"
       onClick={this.toggleDropdown}
+      aria-label="Select Month"
```
halvorsanden, line 71: This would make the element always announce "Select month" even when a month has been selected.

src/month_dropdown.tsx
```diff
@@ -62,8 +68,14 @@ export default class MonthDropdown extends Component<
...
+      aria-expanded={this.state.dropdownVisible}
+      aria-haspopup="listbox"
```
halvorsanden, line 73: This also needs a reference to what element it opens and what element has keyboard focus when the element is open. An aria-controls that takes the id of the related element and an aria-activedescendant with the focused element.

src/month_dropdown_options.tsx
```diff
@@ -50,6 +50,7 @@ export default class MonthDropdownOptions extends Component<MonthDropdownOptions
             }
           }}
           role="button"
+          aria-label={`Select Month ${month}`}
```
halvorsanden, line 53: The role of this element should most likely be changed to "option" rather than adding an aria-label, if the parent element is a listbox. I would also change the parent to a ul element and this from a div to a li.

halvorsanden (commented):
I’m glad to see this being worked on. I think this would benefit from taking a step back and simplifying and considering the elements, roles and context too, so it doesn’t end up with more aria than necessary.

My comments are mainly from experience with making comboboxes, which aren’t that different from this. I find https://www.w3.org/WAI/ARIA/apg/patterns/ very useful, and I also recommend inspecting the solution in the accessibility tree and, if you’re not doing it already, test it with your OS screen reader.

I commented on the month section, but the same goes for the year.

armhaj:
Hi, are there any plans to move this PR forward? If it's no longer being worked on, could you please close it so we can create a new PR and implement the necessary fixes?

balajis-qb:
Thank you @halvorsanden for your suggestions.

@armhaj, I'll look into the suggested changes and work on that.

balajis-qb:
Closing this PR as I ran into issues with the existing branch. I’ve created a new PR with the suggested changes and improvements [here](https://github.com/Hacker0x01/react-datepicker/pull/6287)

Please feel free to continue the review on the new PR — thanks!

@halvorsanden  @armhaj

src/month_dropdown.tsx
balajis-qb: Agreed. Removed redundant aria-* from native <select>/<option> to rely on built-in semantics and avoid conflicting announcements.  

Addressed in the new PR.

src/month_dropdown.tsx
balajis-qb: Using a default aria-label="Month" for now to keep API minimal and consistent with existing patterns.

src/month_dropdown.tsx
balajis-qb: Updated to a neutral label (“Month”) to avoid misleading announcements once a value is selected.

src/month_dropdown.tsx
balajis-qb: Not adding these since the custom dropdown is not a true combobox pattern. It follows a simpler listbox model, so I feel aria-controls/aria-activedescendant would be unnecessary here.

src/month_dropdown_options.tsx
balajis-qb: Updated custom dropdown to use ul/li with role="option" for selectable items. Navigation controls remain buttons as they represent actions, not options.

